Issue #5, opened 2018-08-20
pathOr does not work with optional properties


epeli:
Just found out this lib and it seems sweet!

I tried to playaround with the `pathOr` function but it looks like it does not work if the object properties are optional

```ts
type PartialThing = {
    foo?: {
        bar?: {
            num?: number;
            char?: string;
        };
    };
};

const thing: PartialThing = {};

const val = R.pathOr(thing, ["foo", "bar", "num"], 3);
```

![kuva](https://user-images.githubusercontent.com/225712/44369093-d69ad880-a4dd-11e8-8de3-80d2ab2ab3ad.png)

But I found out that it is possible to workaround it by casting the data to be deeply required:

```ts
type DeepRequired<T> = {
    [P in keyof T]-?: T[P] extends Array<infer U>
        ? Array<DeepRequired<U>>
        : T[P] extends Array<infer U>
            ? Array<DeepRequired<U>>
            : DeepRequired<T[P]>
};


// Type checks!
const val = R.pathOr(
    thing as DeepRequired<typeof thing>,
    ["foo", "bar", "num"],
    3,
);
```

I wonder if it would be possible to bake this into the `pathOr` types? Thanks.

BetterCallSky:
Can you provide your `tsconfig.json` and typescript version?
Do you use `"strictNullChecks": true`?

I copied your code to unit tests, and it worked with `strictNullChecks: false`.

![image](https://user-images.githubusercontent.com/4478684/44399605-81e07780-a548-11e8-92fe-4bebb0f8f967.png)

I will try to make it compatible with strictNullChecks.
Thanks!

epeli:
> Do you use "strictNullChecks": true?

Yes. I actually never use TypeScript without it so I actually forgot it can be turned off  😄 

Here's full config I used


```json5
{
    "compilerOptions": {
        "module": "commonjs",
        "target": "es6",
        "lib": ["es6", "dom"],
        "sourceMap": true,
        "noEmit": true,
        "jsx": "react",
        "moduleResolution": "node",
        "forceConsistentCasingInFileNames": true,
        "strict": true, // enables strictNullChecks and other strict rules
        "esModuleInterop": true,
        "noImplicitReturns": true,
        "noImplicitThis": true,
        "noImplicitAny": true,
        "noUnusedLocals": true
    }
}
```

BetterCallSky:
fixed in 0.0.10

PR #13, opened 2019-03-06
Enable strictNullChecks

mulias:
Addresses #5 

These changes allow remeda to compile with `strictNullChecks: true`, and should ensure that functions such as `pick` and `pathOr` will have the same typing behavior no matter if a project has `strictNullChecks` enabled or disabled.

Changes:
- Add `.prettierrc` and exclude `/dist` from test runs. I had to make these changes to get my environment working correctly, but they can be dropped if they aren't helpful.
- Refactor `LazyResult` to work better with `strictNullChecks`.
- Explicitly allow the `object` argument of `pick` to be `undefined` or `null`, matching the `strictNullChecks: false` behavior.
- Allow `pathOr` to select keys that are possibly missing in the object, matching the `strictNullChecks: false` behavior.
- Allow `pathOr`'s `defaultValue` argument to be of any type. This broadens the behavior or `pathOr` a bit, but seems appropriate.
- Enable `strictNullCheck`, which no longer triggers compiler errors.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,2 @@
+singleQuote: true
+trailingComma: es5
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     },
     "testRegex": "/src/.*.test.(jsx?|tsx?)$",
     "testPathIgnorePatterns": [
-      "/node_modules/"
+      "/node_modules/", "/dist/"
     ],
     "collectCoverageFrom": [
       "src/**/*.tsx",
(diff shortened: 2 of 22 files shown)


mulias:
I'm planning on adding more tests for `pathOr`, since I suspect there's still a bug related to unions with non-indexable types. For now I'm opening this PR to gauge interest and find out what would be most helpful to this project.

BetterCallSky:
Great job @mulias!

I will review it in 1 week because I am on vacation right now.

BetterCallSky:
thanks again
changes are released in 0.0.10

PR #17, opened 2019-04-19
 be more elaborate in the lazy evaluation example

evdama:
be more elaborate in the lazy evaluation example, avoid confusion


BetterCallSky:
thanks!

PR #18, opened 2019-05-09
Added R.forEachObjIndexed

mablin7:
Added forEachObjIndexed, based on [ramda's implementation](https://ramdajs.com/docs/#forEachObjIndexed)

### DIFF
--- a/src/forEachObj.test.ts
+++ b/src/forEachObj.test.ts
@@ -1,0 +1,62 @@
+import { forEachObj } from './forEachObj';
+import { pipe } from './pipe';
+
+const obj = {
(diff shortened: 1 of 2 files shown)


BetterCallSky:
Can you implement something similar to existing `forEach`?

```ts
R.forEachObj(array, fn) // fn has only 1 arg 'value'
R.forEachObj.indexed(array, fn) // fs has 2 args 'value', 'key'
```

mablin7:
Yeah i guess that makes sense, will do!

BetterCallSky:
Thanks!
Sorry for late response

PR #47, opened 2020-03-20
Misc jsdoc fixes

sobstel:
* Fix function description for data-first drop()Tiny doc fix for data-first drop()
* Fix signature typo in forEachObj


lstkz:
thanks!

PR #49, opened 2020-03-21
Adds `includes` function

SirJayEarn:


mapping.md
```diff
@@ -56,4 +55,5 @@
 | `times`        | `times`        | `times`             |
 | `toPairs`      | `toPairs`      | `toPairs`           |
 | `type`         | `-`            | `type`              |
-| `uniq`         | `uniq`         | `uniq`              |
\ No newline at end of file
+| `uniq`         | `uniq`         | `uniq`              |
+| `includes`     | `includes`     | `includes`          |
```
sobstel: I think this is sorted alphabetically.

src/includes.test.ts
```diff
@@ -0,0 +1,126 @@
...
+      expect(includes([1, 2, 3, 4, 5], 6)).toBe(false);
+    });
+
+    it('works with string arrays', () => {
+      expect(includes(['a', 'b', 'c', 'd', 'e'], 'c')).toBe(true);
+      expect(includes(['a', 'b', 'c', 'd', 'e'], 'f')).toBe(false);
+    });
+
+    it('works with objects', () => {
+      expect(includes({ a: 1, b: 2, c: 3 }, 'a')).toBe(true);
```
sobstel: This is not how lodash `includes` works. It checks for value, not for a key.

src/includes.ts
```diff
@@ -0,0 +1,60 @@
+import { purry } from './purry';
+import { _reduceLazy } from './_reduceLazy';
+import { equals } from './equals';
+
+export type Includable<T> = T[] | Object | string;
```
sobstel: I'm not sure if it should support both object and string. As far as I saw all the other functions support either object OR string, so it's a bit inconsistent wit the rest of the lib.

lstkz:
Can we support only strings (without regex), and arrays?

src/includes.test.ts
SirJayEarn: Ok I see, so checking keys would rather be the job of `has`?

src/includes.ts
SirJayEarn: Idk either - so what should `includes` handle? Only objects and arrays? Will there be a separate function for checking strings?

SirJayEarn:
I'm fine with only supporting strings and arrays. But than it's not compatible with lodashs includes

src/includes.test.ts
SirJayEarn: Now checking object values

lstkz:
We don't have to replicate lodash in 100%.
`R.includes` should simply wrap the original `Array.includes`, and `String.includes` function.

> Why not using regex?

It can be confusing, the provided string must be a valid regex.
For example:
`R.includes('users (0)', '0)')`
throws an error

SirJayEarn:
> We don't have to replicate lodash in 100%.
> `R.includes` should simply wrap the original `Array.includes`, and `String.includes` function.
> 
> > Why not using regex?
> 
> It can be confusing, the provided string must be a valid regex.
> For example:
> `R.includes('users (0)', '0)')`
> throws an error

Ah ok, I see. I'll change that

src/includes.ts
```diff
@@ -0,0 +1,48 @@
...
+    return _arrayIncludes(item, source);
+  } else if(typeof item === 'string') {
+    return _stringIncludes(item, source);
+  } else {
+    return false;
+  }
+}
+
+function _arrayIncludes<T>(item: T, arr: T[]) {
+  return arr?.some(a => equals(a, item)) ?? false;
```
lstkz: - `arr?.` is not needed, it should be `arr.`
- `?? false` is not needed
- no partial matching, it should call `arr.includes`

src/includes.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+function _stringIncludes(item: string, source: string) {
+  return source?.includes(item) ?? false;
```
lstkz: - similar to above, remove `??`, and `source?.` -> `source.`

src/includes.ts
```diff
@@ -0,0 +1,48 @@
...
+  return purry(_includes, arguments);
+}
+
+function _includes<T>(source: Includable<T>, item: T): boolean {
+  if (Array.isArray(source)) {
+    return _arrayIncludes(item, source);
+  } else if(typeof item === 'string') {
+    return _stringIncludes(item, source);
+  } else {
+    return false;
```
lstkz: if the type is not supported, typescript will report an error
instead of `return false`, we can throw an error

lstkz:
Closing because of inactivity.

PR #51, opened 2020-03-23
Reverse function

sobstel:


sobstel:
Second though: it's still one-liner, I think it might be better to document it in mapping doc or so instead of making remeda unnecessarily  bloated.

src/reverse.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @category Array
+ */
+export function reverse<T>(): (array: readonly T[]) => Array<T>;
+
+export function reverse() {
+  return purry(_reverse, arguments);
+}
+
+function _reverse(array: any[]) {
+  return array.slice(0).reverse();
```
vladimiry: Using spread operator for copying the array provides better readability and doesn't require using `magic number / zero`.

lstkz:
The build is failing. Can you check it?  You can also run `yarn run prettier` to auto-format everything.

sobstel:
> The build is failing. Can you check it? You can also run `yarn run prettier` to auto-format everything.

Please take a look at my comment above (https://github.com/remeda/remeda/pull/51#issuecomment-603260186). I think we should just close it.

src/reverse.ts
sobstel: The thing is they're not equivalent. There's at least one edge-case I know about:

```
const arr = []
arr.length = 2;
[...arr] // => [undefined, undefined]
```

slice() will return empty array 

I would normally use spread operator for the project I have a control over, but not for the lib like this.

vladimiry: Good point. Still one improvement is possible, `slice` could be called without args since 0 is the default value.

lstkz:
thanks!

src/reverse.ts
vladimiry: @leilapearson By changing the `length` you didn't put the `undefined` values into the array but `empty slots`, so why putting `undefined` values in the cloned copy doing `[...arr]` when the original array has `empty slots` in those indexes. In general, another big difference is that empty slots are non-iterable, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/length.

leilapearson: Correcting myself after looking at it a bit more closely. When @sobstel said it would return an empty array I thought it was losing array sparseness in the process, but it appears not... and furthermore empty items are not the same thing as undefined items. 

```ts
console.log(arr); // [ 1 ]
console.log([...arr]); // [ 1 ]

arr.length = 3;

console.log(arr); // [ 1, <2 empty items> ]
console.log([...arr]); // [ 1, undefined, undefined ]
console.log([1, , ,]); // [1, <2 empty items> ]

console.log(arr.slice()); // [1, <2 empty items> ]
```
So you're right  @sobstel (naturally :-) that `arr.slice()` is the way to go.

leilapearson: Thanks, @vladimiry. Didn't see your comment before typing my second comment. Great to learn something new :-)

Issue #37, opened 2019-11-17
Roadmap / contribution


SirJayEarn:
#10 already raises a similar question, but is closed now. So I think this question is still open for potential contributors: **what's next?**
I like this project and for it to be a true alternative to lodash or ramda I think it should focus on ease of migration (from one of the named libraries) and provide alternatives to all the main utils.
This leads to another question: _what are the main functions?_
So just to get this conversation going,  here is what I think is missing in remeda:

`isNil`, `defaultTo`, `is`, `isEmpty`, `uniqBy`, `path`, `prop`, `propOr`, `propEq`, `nth`, `length`, `includes`, `has`, `assoc`, `assocPath`, `tail`, `init`, `debounce`, `throttle`

If this lists seems reasonable I would pick one of these from time to time and mark them as in progress (by creating an issue or something) and then create a pull request when done.

What I therefore also would find helpful, would be a short introduction about how to implement lazyness, as this is not yet completely clear to me.

BetterCallSky:
If you think there are core functions are missing, then I will be happy to accept PRs, but some functions from lodash/ramda are 1-liners.

- `isNil`, I think it's not needed because it can be replaced with `x => x == null`
-  `defaultTo` we can add it
- `is`, not sure how to implement it, and I don't like ramda implementation https://ramdajs.com/docs/#is. Currently, I always use `typeof` or `Array.isArray`. We can discuss it in a separate issue.
- `uniqBy` we can add it
- `path`, I think it's not needed because in the new typescript version we can use `a?.b?.c` syntax
- `prop`, not sure because can be replaced with `x => x.foo`
- `propOr`, not sure because can be replaced with `x => x.foo || defaultValue`
- `propEq `, not sure because can be replaced with `x => x.foo === value`
- `nth`, not sure because can be replaced with `x => x.foo[1]`
- `length`, not sure because can be replaced with `x => x.length`
- `includes`,  in ramda https://ramdajs.com/docs/#includes they support string, objects, and arrays, we can do something similar
- `has` we can add it
- `assoc`, we can use `merge` instead
- `assocPath`, it could be good to have it, but it can be non-trivial to implement or there can be many use cases. For example: updating arrays. Maybe it's better to use `immer` library?
- `tail`, `init`,  we can add it, but how often people use such functions?
- `debounce`, `throttle`, it would be good to have it. Something similar to lodash, but with a single `options` argument.

> would find helpful, would be a short introduction about how to implement lazyness,

[...]

Issue #29, opened 2019-10-26
Incomplete mapping.md


0x80:
For people migrating from Lodash or Ramda it would be useful if the [mapping.md](https://github.com/remeda/remeda/blob/master/mapping.md) provides a complete list. Or is that not the goal of this file?

For example, I'm not sure how to replace the Lodash `times` utility...

BetterCallSky:
Unfortunately, there is no replacement for `_.times`. Would you be interested in implementing it?

SirJayEarn:
I think I can have a look at implementing `times` in the next couple of days

leilapearson:
Have created related pr #53 to document some of the missing functions as Typescript one-liners.

Needs more discussion though...

lewisl9029:
Since there's already a toPairs function, I found fromPairs to be a curious omission, as they go hand in hand. 

Would you be willing to accept a PR for it? Thanks!

lstkz:
@lewisl9029 Sure, feel free to submit PR.

PR #53, opened 2020-03-25
Add one-liners to mapping.md

leilapearson:
Initial list of one-liners.

### DIFF
--- a/mapping.md
+++ b/mapping.md
@@ -70,5 +70,5 @@ are listed below.
 | `isNil`        | `isNil`        | `x => x == null`                           |
 | `nth`          | `nth`          | `n => a[n]`                                |
 | `-`            | `path`         | `a?.b?.c`                                  |
-| `-`            | `propEq`       | `x => x[a] === value`                      |
-| `-`            | `propOr`       | `x => x[a] ?? defaultValue`                |
+| `-`            | `propEq`       | `x => x.a === value`                      |
+| `-`            | `propOr`       | `x => x.a ?? defaultValue`                |
--- a/mapping.md
+++ b/mapping.md
@@ -19,7 +19,7 @@
 | `find`         | `find`         | `find`              |
 | `findIndex`    | `findIndex`    | `findIndex`         |
 | `first`        | `head`         | `head`              |
-| `flatMap`      | `flatMap`      | `-`                 |
+| `flatMap`      | `flatMap`      | `chain`             |
 | `flatten`      | `flatten`      | `flatten`           |
 | `flattenDeep`  | `flattenDeep`  | `flatten`           |
 | `forEach`      | `forEach`      | `forEach`           |
@@ -67,8 +67,18 @@ are listed below.
 | Lodash         | Ramda          | TypeScript                                 |
 | -------------- | -------------- | ------------------------------------------ |
 | `add`          | `add`          | `(x, y) => x + y`                          |
+| `-`            | `append`       | `(arr, val) => [...arr, val]`              |
+| `concat`       | `concat`       | `(arr1, arr2) => arr1.concat(arr2)`<sup id="a1">[1](#f1)</sup>  |
+| `constant`     | `always`       | `x => () => x`                             |
+| `filter`       | `filter`       | `(arr, pred) => arr.filter(pred)`<sup id="a2">[2](#f2)</sup> |
+| `identity`     | `identity`     | `x => x`                                   |
 | `isNil`        | `isNil`        | `x => x == null`                           |
 | `nth`          | `nth`          | `n => a[n]`                                |
 | `-`            | `path`         | `a?.b?.c`                                  |
-| `-`            | `propEq`       | `x => x.a === value`                      |
-| `-`            | `propOr`       | `x => x.a ?? defaultValue`                |
+| `-`            | `propEq`       | `x => x.a === value`                       |
+| `-`            | `propOr`       | `x => x.a ?? defaultValue`                 |
+
+#### Footnotes
+
+1. <small id="f1">Incomplete replacement for Lodash `concat`. Lodash `concat` accepts multiple values to concatenate, including values that are not within an array. </small> [↩](#a1)
+2. <small id="f2">To filter objects instead of arrays, use `(obj, pred) => Object.fromEntries(Object.entries(obj).filter(pred));` </small> [↩](#a2)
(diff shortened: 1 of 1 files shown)


leilapearson:
Looking at these lists as a possible source of one-liners to include - where not already part of Remeda:

- https://mostly-adequate.gitbooks.io/mostly-adequate-guide/appendix_a.html
- https://mostly-adequate.gitbooks.io/mostly-adequate-guide/appendix_c.html 

A question for you all... For functions that take more than one argument, should we be showing both the regular and manually curried version in the one-liners? Or just one or the other?

For example:

| Lodash         | Ramda          | TypeScript                                 |
| -------------- | -------------- | ------------------------------------------ |
| `add`          | `add`          | `(x, y) => x + y` or `x => y => x + y`     |

For a bit more complicated example, what about something like `prop`? The way it's shown in the PR assumes you know the property name at compile time - which is often true - but the actual `prop` function in Ramda looks more like this:

`(p, obj) => obj != null ? obj[p] : undefined` or `p => obj => obj != null ? obj[p] : undefined`

As you can see, the one-liner is getting more complicated...

Even `nth` is not so simple. Here's the Ramda implementation:
var nth = _curry2(function nth(offset, list) {
  var idx = offset < 0 ? list.length + offset : offset;
  return _isString(list) ? list.charAt(idx) : list[idx];
});

To be honest I'm starting to question the value of not including functions like these in Remeda?

(1) It would be easier for people to move to Remeda with drop-in replacement functions for the functions they already use.

(2) It's a utility library, right? And what are utilities but small functions that save you from re-writing the same code over and over? 

So philosophically, should we think that including these types of functions would make Remeda more bloated... or that it would make Remeda more useful? 

No answers from me. Just thinking out loud...

leilapearson:
Forgot to reference the related issues.

This PR partially resolves issue #37  and also relates to  issue #29

sobstel:
@leilapearson  Great work and insight!

I can see where your confusion comes from, but I don't think we should treat `remeda` as a `ramda` clone (neither `lodash` clone for that matter). The latter is a  functional library for JS, the former is a TS utility library.

Saying that, it's much simpler for remeda. `prop` is a very good example, because it's already implemented. See https://github.com/remeda/remeda/blob/master/src/prop.ts. It's just `obj[name];` and we should follow this convention here in mapping file too.

We should also be aware that mappings are not 1-1 as functions between libs sometimes does not match one to one and it's fine. It's just help from people migrating from other libs. 

So answering your original question, for `add` I believe it should be `(x, y) => x + y`.

mapping.md
```diff
@@ -57,3 +57,18 @@
...
+are listed below.
+
+| Lodash         | Ramda          | TypeScript                                 |
+| -------------- | -------------- | ------------------------------------------ |
+| `add`          | `add`          | `(x, y) => x + y`                          |
+| `isNil`        | `isNil`        | `x => x == null`                           |
+| `nth`          | `nth`          | `n => a[n]`                                |
+| `-`            | `path`         | `a?.b?.c`                                  |
+| `-`            | `propEq`       | `x => x[a] === value`                      |
+| `-`            | `propOr`       | `x => x[a] ?? defaultValue`                |
```
lstkz: Could you change `x => x[a]` to `x => x.a`?

lstkz:
@leilapearson 
Thanks, this is a good idea to add 1 liners that are missing.

We can also consider removing existing functions like `R.pathOr` or `R.prop`

sobstel:
> We can also consider removing existing functions like `R.pathOr` or `R.prop`

Removing is always very tricky (that's why you should be careful about adding new stuff). It would break all the clients using it. First, we would need to mark it deprecated, discourage usage in docs, explain it'll be removed in 1.0  and remove it only in 1.0.

mapping.md
leilapearson: Sure

leilapearson:
Added a couple of one-liners that include footnotes. Unfortunately github flavored markup doesn't support footnotes, so I had to use html. 

I think that footnotes are helpful but should be minimized since they will be a bit painful to maintain in terms of maintaining their numbering. Numbers need to be manually entered as link text. 

@lstkz  and @sobstel  can you let me know what you think of:
1. the idea of using footnotes in this way, and
2. if you have any other ideas for how to handle situations where the one-liner is not a drop-in replacement?

One thought would be just to have use a couple of generic footnotes. Something like

1. Lodash function differs. See the [Lodash documentation](https://lodash.com/docs/).
2. Ramda function differs. See the [Ramda documentation](https://ramdajs.com/docs/).

These footnotes would actually apply to some functions in the main mapping table too.

These don't help with situations like `filter` though - where - if you want a simple one-liner - you need a different one-liner for objects versus arrays. Another way to handle this would be two lines in
the one-liner table I guess... representing each overloaded signature... 

Oh, and I've used `pred` for predicate function, `obj` for object, and `arr` for array. Let me know if these don't work.

leilapearson:
Oh, and of course remeda actually does implement `filter` - but it looks like only for arrays. So this is an interesting case where we potentially have both a remeda implementation that covers part of the functionality in Lodash and Ramda AND a one-liner that covers the rest.

I'll tidy things up once the best way to handle this is decided. 

I'll tidy it up once we decide on the best representation.

sobstel:
Just a quick general comment: IMHO not sure if we should be THAT detailed with footnotes (it'll be very hard to maintain and keep them in sync). General note that these are not exact 1-1 mapping should be good enough.

lstkz:
I agree with @sobstel, it's better to not complicate it.
@leilapearson can we include only 1 liner functions that don't exist in Remeda? Functions like `filter`, `map` are re-implemented because we must support lazy evaluation.

leilapearson:
Thanks @sobstel and @lstkz. I've removed the offending one-liners and replaced the footnotes 
with general notes. Take a look and let me know what you think. 

If the general format is okay, and the content is okay too so far, then I'm thinking:

-  if there are a few more very commonly used one-liner functions that people think we should add, then we can add them now and publish the initial list
- the community can continue to update the list over time with more functions as we see the need

Thoughts?

Thanks again for the reviews and helpful comments.

leilapearson:
@sobstel and @lstkz  - One other thing... I wrote the `reverse` function to include a type on the argument so it will actually compile - but that isn't the case with `append` for example - which would become `append = <T>(arr: Array<T>, val: T) => [...arr, val]`. 

Which style do you prefer?

lstkz:
@leilapearson looks good, thanks!

PR #56, opened 2020-04-11
Fix missing uniqBy from exports

montzkie18:


src/index.ts
```diff
@@ -53,3 +53,4 @@ export * from './times';
 export * from './toPairs';
 export * from './type';
 export * from './uniq';
+export * from './uniqBy';
```
geekflyer, line 56: perhaps add en emtpy newline.

lstkz:
@montzkie18 thanks!

PR #58, opened 2020-04-13
chore: added uniqBy to default library exports

totev:
Just a minor thing -> saw that **uniqBy** was missing from the default exports and must be referenced directly from its inner path.


lstkz:
Thanks, but already fixed in https://github.com/remeda/remeda/pull/56

PR #60, opened 2020-04-23
Narrow typings for compact()

MikeyBurkman:
I noticed that calling `compact()` didn't actually remove falsy values according to the typings, which is one of the main reasons I use that function in lodash. IE: `_compact((string|undefined)[])` should result in `string[]`. This doesn't change any functionality, just the typings.

Also, I noticed that for `R.uniq`, two typings are provided for the function, `items => items` and `() => items => items`, while compact() just has this single definition. This lead to some confusion in R.pipe, as I had to call `R.uniq()`, but just use `R.compact` without calling it. I feel like we should standardize on one of these, maybe the `R.uniq()` style?

### DIFF
--- a/src/compact.ts
+++ b/src/compact.ts
@@ -1,4 +1,4 @@
-function notNil<T>(value: T | null | undefined | false | '' | 0): value is T {
+function isTruthy<T>(value: T | null | undefined | false | '' | 0): value is T {
   return !!value;
 }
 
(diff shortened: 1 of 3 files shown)


src/compact.ts
```diff
@@ -1,3 +1,7 @@
+function isTruthy<T>(value: T | null | undefined | false | '' | 0): value is T {
```
MikeyBurkman, line 1: This would probably just be a useful top-level function perhaps?

lstkz:
thanks!

PR #64, opened 2020-05-11
Added support for type guards in filter function

digimuza:
Currently, the filter function and other functions like reject do not support type guards.
I added some simple overloads to support this feature.

```typescript
// Guard format that I use
function isNumber<T>(data: T): data is Extract<T, number> {
  return typeof data === 'number'
}
```

Before
```typescript
const result = filter([1, 2, 3, 'abc', true] as const, isNumber); value is [1,2,3] incorect type 1 | 2 | 3 | 'abc' | true
`

After
```typescript
const result = filter([1, 2, 3, 'abc', true] as const, isNumber); value is [1,2,3] type 1 | 2 | 3
```

This pull request will add this feature only for filter and filter.indexed functions.

In next pull request, I will add this support for reject and other functions that can accept type guards

P.S Please test not only functionality but also types.
NOTE: broken type is equal to broken function :) 
It's really simple to write tests for typescript. Just create function assertType

```typescript
function assertType<T>(data: T) { return data } // Simple test function

// And test your type output like asserType<YOUR FINAL TYPE>(resultTrasformation) 
```
You can see it in my pull request.

### DIFF
--- a/src/filter.test.ts
+++ b/src/filter.test.ts
@@ -2,10 +2,26 @@ import { pipe } from './pipe';
 import { filter } from './filter';
 import { createCounter } from './_counter';
 
+function assertType<T>(data: T): T {
+  return data
+}
+
+function isNumber<T>(data: T): data is Extract<T, number> {
+  return typeof data === 'number'
+} // TODO Refactor to remeda function
+
 describe('data_first', () => {
+
   it('filter', () => {
     const result = filter([1, 2, 3] as const, x => x % 2 === 1);
+    assertType<(1 | 2 | 3)[]>(result) // Type test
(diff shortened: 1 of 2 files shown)


lstkz:
Thanks, it looks good.
Can you remove this comment https://github.com/remeda/remeda/pull/64/files#diff-df4810139fa860e50ada49471cc25088R76?

digimuza:
> Thanks, it looks good.
> Can you remove this comment https://github.com/remeda/remeda/pull/64/files#diff-df4810139fa860e50ada49471cc25088R76?

Comment is removed

lstkz:
thanks, good work!

PR #65, opened 2020-05-11
Added multi keys group support to groupBy function

digimuza:
Currently, the groupBy function lacks some additional functionality. Actually this feature is not supported by ramada and loadash but it is very useful.

There is a lot of use cases where you want to group data with an arbitrary list of values. 

Let's take an example. 
Objective 1: Extract active chatRoomsIds from the user list.
Objective 2: Users can join multiple rooms. And we want to know other users that joined the same room

```typescript
interface User { 
  name: string,
  joinedChatRooms: string[]
}

const users: User[] = [...]
R.groupBy(users, (user) => user.joinedChatRooms)
```

The provided example is solving a common problem where you need to extract statistics from the list of data. In most cases to solve this problem, you have to implement custom groupBy function
An additional example where such a feature is useful

Find out how much particular attribute have products
```typescript
interface Product {
  id: string,
  attr: Record<string, string>
}
const products: Product[] = [...]

R.pipe(
     products,
     R.groupBy((product) => Object.keys(product.attr)),
     R.toPairs(),
     R.map(([attr, productList])=> [attr, productList.lenght]),
     R.sortBy(([,count])=>count)
)
// An display it in sidebar
```

### DIFF
--- a/src/groupBy.test.ts
+++ b/src/groupBy.test.ts
@@ -16,13 +16,66 @@ const expected = {
   2: [{ a: 2, b: 1 }],
 };
 
+const employees = [
+  {
+    name: 'John',
+    skills: ["Printing", 'Painting', 'Writing']
+  },
+  {
+    name: 'Britney',
+    skills: ["Printing", 'Managing', 'Acting']
+  }
+]
+const expectedEmployees = {
+  "Printing": [
+    {
+      name: 'John',
+      skills: ["Printing", 'Painting', 'Writing']
+    },
+    {
+      name: 'Britney',
+      skills: ["Printing", 'Managing', 'Acting']
+    }
+  ],
+  "Painting": [
+    {
+      name: 'John',
+      skills: ["Printing", 'Painting', 'Writing'] 
+    }
+  ],
+  "Managing": [
+    {
+      name: 'Britney',
+      skills: ["Printing", 'Managing', 'Acting']
+    }
+  ],
+  "Acting": [
+    {
+      name: 'Britney',
+      skills: ["Printing", 'Managing', 'Acting']
+    }
+  ],
+  "Writing": [
+    {
+      name: 'John',
+      skills: ["Printing", 'Painting', 'Writing']
+    },
+  ]
+}
+
 describe('data first', () => {
   test('groupBy', () => {
     expect(groupBy(array, x => x.a)).toEqual(expected);
   });
+  test('groupBy multi keys', () => {
+    expect(groupBy(employees, (x) => x.skills)).toEqual(expectedEmployees)
+  })
   test('groupBy.indexed', () => {
     expect(groupBy.indexed(array, x => x.a)).toEqual(expected);
   });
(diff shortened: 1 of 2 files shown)


lstkz:
Thanks for the PR, but I don't think it's core functionality.
Lodash, Ramda, underscore don't support it. GroupBy in other languages (like c#) also doesn't work in this way.
Sometimes it's just better to create your own function.

digimuza:
> Thanks for the PR, but I don't think it's core functionality.
> Lodash, Ramda, underscore don't support it. GroupBy in other languages (like c#) also doesn't work in this way.
> Sometimes it's just better to create your own function.

It's kinda sad that remeda want's to be just loadash alternative. 
This function has clear and common use case it does not add additional complexity, modification kinda minimal and API is fully compatible with any other groupBy function.

It just adds pure value to this function

And let's compare this eavil function with other `core` functions
1. R.pipe({ foo: 'bar' }, R.prop('foo')) // Adds unnesasasry complexity + huge type problems
2. R.pathOr(['y'], 2) // Why ? Typescript have a native way to do this. This is not a typescript friendly function at all.
R.pick(['a', 'd']) Realy bad function for typescript does not work with unions and this function tend to lie. Useful for javascript but pain in the ass with typescript
3. R.objOf(10, 'a') // => { a: 10 } // What use of this function? Not typescript friendly and can lie really hard
```typescript
objOf(10, 'a' as 'a' | 'b' | 'c' ) // This will be really common bug
```
4. set({ a: 1 }, 'a', 7 ) // Why? This function does not have purpose. It's add more pain. { ...{a: 1}, a: 7 } it's explicit 
```typescript
interface SuperUser {
  a: number
}

interface RegularUser {
  b: number
}

type User = SuperUser | RegularUser
```

5. addProp('lastName', 'doe')({} as User)  // { lastName: string } 

What's my vision for remeda 
1. No nasty shortcuts - function must provide real value. 
2. Functions that solve real problems - not add them (addProp, set, objOf) function like this can really easily mess up types 
3. The functions should never lose types
4. Functions have to prevent errors (mistypes, runtime errors, ...)
5. Functions should adapt to changes. 

PS. 

1. Current groupBy function not typescript friendly 

interface Item {
  type: 'a' | 'b' | 'c'
}
const items: Item[] = [] 
```typescript
R.groupBy(items, (item) => item.type) // Expecting type Record<'a' | 'b' | 'c', Item[]> not Record<string, Item[]>
```

2. It has a hidden bug 

```typescript

interface User {
 a: number
}
const users: User[] = [...]

// Somewhere in code
FP.groupBy(users,(u)=>u.a) // Record<string, User>

// Small refactor of the user object
interface User {
 a: { group: number }
}
FP.groupBy(users,(u)=>u) 

// And we have a worst possible bug 
```

lstkz:
My point is that groupBy is a well-known function that is available in many languages, and it has the same implementation everywhere. Extending it can be confusing for potential people.

PR #66, opened 2020-05-11
Improved sort functionality. Added way to sort by multiple fields

digimuza:
Currently, the sort function is kinda limited. 

I added a couple of modifications to make it more useful 

1. Ability to sort by multiple fields
```typescript
interface User {
   firstName: string,
   lastName: string,
   active: boolean
}
const users: [...]
R.sortBy(users, (user) => ([ user.active, user.lastName, user.firstName  ]))
```

2. Ability to define order 

```typescript
R.sortBy(users, (user) => ([ { value: user.active, order: 'desc' } , { value: user.lastName, order: 'asc' } , { value: user.firstName, order: 'asc' }  ]))
```

3. Ability to use the custom comparison function

```typescript
R.sortBy(users, (user) => ([ { value: user.active, order: 'desc' } , { value: user.lastName, order: 'asc' } , { value: user.firstName, order: 'asc', compare: (a,b) => a.toString().localeCompare(b.toString()) }  ]))
```

### DIFF
--- a/src/sortBy.test.ts
+++ b/src/sortBy.test.ts
@@ -4,14 +4,58 @@ import { pipe } from './pipe';
 const items = [{ a: 1 }, { a: 3 }, { a: 7 }, { a: 2 }] as const;
 const sorted = [{ a: 1 }, { a: 2 }, { a: 3 }, { a: 7 }];
 
+
+const multiPropSortItems = [
+  { favorite: false, category: "A" },
+  { favorite: true, category: "C" },
+  { favorite: true, category: "B" },
+  { favorite: false, category: "B" }
+]
+
+// Expect favorites sorted alphabetically
+const multiPropSortItemsSorted = [
+  { favorite: true, category: "B" },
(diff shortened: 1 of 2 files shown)


lstkz:
Is it possible to have it as an extra function? For example `multiSortBy`? I don't like multiple functions overloads because type errors are hard to read.

digimuza:
Yes, I can rename it.
But I don't see a point to have two sortBy functions.
 
I did not add any overloads and type is not that complex :) 

```typescript
type = SortValue | ComplexValue | ReadonlyArray<SortValue | ComplexValue> 
```

lstkz:
If there are many overloads, and there is a type error, then the error message from typescript is not readable.
for example
```ts
    const arr = [{ a: 1, b: {c: 1} } ] as const;
    sortBy(arr, x => x.b)
```
image


if you split this function into two separate functions, then the error message is much more friendly.

lstkz:
Closing because of inactivity.

PR #67, opened 2020-05-12
Feature/guards

digimuza:
Added common utility typescript guards.

Tested with Array.filter method and simple condition

PR #69, opened 2020-05-14
Feature/mergeDeep

digimuza:
This feature depends on pull request #67 

Created deepMergeLeft and deepMergeRight

```typescript
/**
 * Merging object from left to right
 * 
 * @description
 * @param target value be preserved if possible.
 * Consider following
 * 
 * array + obj = array
 * obj + array = obj
 * obj + obj = obj (recursively merged)
 * array + array = array (removes duplicates using Set)
 * (truthy plain value) + ob = (truthy plain value)
 * (truthy plain value) + undefined = (truthy plain value)
 * A(truthy plain value) + B(truthy plain value) = A(truthy plain value)
 * 
 * Handles circular references
 * @returns {T}
 */
export function deepMergeLeft<T extends any>(target: T, ...sources: DeepPartialObject<T>[]): T
```

```typescript
/**
 * Merging object from right to left
 * 
 * @description
 * @param target value will be replaced if possible.
 * Consider following
 * 
 * array + obj = obj
 * obj + array = array
 * obj + obj = obj (recursively merged)
 * array + array = array (removes duplicates using Set)
 * (truthy plain value) + undefined = (truthy plain value)
 * A(truthy plain value) + B(truthy plain value) = B(truthy plain value)
 * Handles circular references
 * @returns {T}
 */
export function deepMergeRight<T extends any>(target: T, ...sources: DeepPartialObject<T>[]): T
```

The final output type will be the same as target.
There are a couple of reasons why I prefer this interface over the typescript & intersection

1. This always outputs clean type
2. While merging you have autocomplete. (this is really good for filling default values)
3. Type never lies

Issues with this that you can't do merging without knowing the output type.

```typescript
deepMergeLeft({ a: 5 }, { c: 5 }) // Typescript error

interface Item {
   a: number
   c?: number
}
const data1: Item = { a: 5 }
deepMergeLeft(data1, { c: 5 }) // All good
```

Use cases

1. Merging data in the reducer.

```typescript
// It's really convenient to reduce action like this
function reducer(s: State, payload: Action) {
     return deepMergeRight(s, {  a: { b: { c: payload.value }  } })
} 
```

2. Adding default values to config files

```typescript
// It's really convenient to reduce action like this
interface Config {
   ...big configuration object with a lot of optional values
}
const myBasicConfig = {} // My small config
const defaultConfig = {...}

const newConfig = deepMergeLeft(myBasicConfig, defaultConfig)
```

Renamed function to deepMerge... instead of mergeDeep...
    
This change is just for consistency. Later on, this package will have deepUniq, deepEquals, deepSort...


lstkz (changes requested):
CI is failing.

Bessonov:
`merge` (or `deepMerge*`) is a very needed helper. But, well, the semantic of `merge` is not so easy. Proposed MR is more natural way of merge:

```typescript
    const d1: Data = {
        a: [1, 2, 3, { a: 1 }],
    };
    const d2: Data = {
        a: [7, 2, 3, { a: 1 }, { a: 2 }],
    };
    expect(deepMergeLeft(d1, d2)).toEqual({
        a: [1, 2, 3, { a: 1 }, 7, { a: 1 }, { a: 2 }],
    });
```

But, for example, lodash has another semantic on merging. It treats Arrays like Objects with properties `0`, `1` etc.:
```typescript

    const d1: Data = {
        a: [1, 2, 3, { a: 1 }],
    };
    const d2: Data = {
        a: [7, 2, 3, { a: 1 }, { a: 2 }],
    };
    expect(_.merge(d1, d2)).toEqual({
        a: [7, 2, 3, { a: 1 }, { a: 2 }],
    });
```

Therefore, I propose following changes:
1. Remove the current `merge`. In my experience, there are a few reasons to have a shallow merge, if at all. Furthermore, it can be accomplished with `{...obj1, ...obj2}` or friends easily.
2. From my point of view, the proposed semantic is what I'm expecting from merge (I didn't look at edge cases like merging Object with Array) and should be introduced as `merge`.
3. I'm not sure, but I think it's possible to implement `merge` in exchangeable way by providing something like `merger` as argument. So we can have something like `merge(obj1, obj2, R.merger.lodash)`, `merge(obj1, obj2, R.merger.right)`or `merge(obj1, obj2, R.merger.shallow)` etc. But I'm fine with `merge.lodash(obj1, obj2)` too.

What do you think?

hgezim:
@TkDodo why was this closed?

TkDodo:
@hgezim because CI was failing, it was dependent on a PR that we closed and there was no movement for a year.

PR #75, opened 2020-06-19
add groupWith function from ramda

atombrenner:
This PR adds the groupWith function from ramda. I just copied implementation and tests and converted them to typescript and jest.


lstkz:
@atombrenner Thanks looks good. Few more things
- If you copy the code from somewhere, you should add a link to its license.
- New functions must be added to index.ts.
- I think the return type is wrong? It should be `T[][]` instead of `T[]`.

lstkz (changes requested):
- If you copy the code from somewhere, you should add a link to its license.
- New functions must be added to index.ts.
- I think the return type is wrong? It should be T[][] instead of T[].

TkDodo:
I'll close this given the staleness.

PR #85, opened 2020-08-24
improve types of mapKeys

TkDodo:
the type S was actually never used, and the result object always had `any` as value;

now, we can let the result type have the key be whatever we return from the mapping function and the value will stay the same as before

restricting S to extends string | number | symbol will also ensure that the function only returns something that can actually be used as a key in an object


lstkz:
@TkDodo thanks!

PR #86, opened 2020-08-24
add mapValues

### LINKED ISSUE #77: Feature Request: mapValues
As per `mapKeys` but iterating over the values of the object.

```typescript
R.mapValues({ a: 1, b: 2 }, (key, value) => key + value) // => { a: a1, b: b2 }
```

TkDodo:
closes #77 

similar to the mapKeys function, but the keys stay the same and the value of the object is mapped instead

I opted for inverting the parameters compared to `mapKeys`, because most of the time, you won't need the `key`, but only the `value`, so having it as a first parameter seems better. This is also how [mapValues from lodash](https://lodash.com/docs/4.17.15#mapValues) works.

### DIFF
--- a/src/mapValues.ts
+++ b/src/mapValues.ts
@@ -1,0 +1,43 @@
+/**
+ * Maps values of `object` and keeps the same keys.
+ * @param object the object to map
+ * @param fn the mapping function
+ * @signature
+ *    R.mapValues(object, fn)
+ * @example
+ *    R.mapValues({a: 1, b: 2}, (value, key) => value + key) // => { a: 1a, b: 2b }
+ * @data_first
+ * @category Object
+ */
+export function mapValues<T extends object, S>(
+  object: T,
+  fn: (value: T[keyof T], key: keyof T) => S
+): Record<keyof T, S>;
+
+/**
+ * Maps values of `object` and keeps the same keys.
+ * @param fn the mapping function
+ * @signature
+ *    R.mapKeys(fn)(object)
+ * @example
+ *    R.pipe({a: 1, b: 2}, R.mapValues((value, key) => value + key)) // => { a: 1a, b: 2b }
+ * @data_last
+ * @category Object
+ */
+export function mapValues<T extends object, S>(
+  fn: (key: keyof T, value: T[keyof T]) => S
+): (object: T) => Record<keyof T, S>;
+
+export function mapValues(arg1: any, arg2?: any): any {
+  if (arguments.length === 1) {
+    return (data: any) => _mapValues(data, arg1);
+  }
+  return _mapValues(arg1, arg2);
+}
+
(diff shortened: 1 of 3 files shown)


src/mapValues.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Maps values of `object` and keeps the same keys.
+ * @param object the object to map
+ * @param fn the mapping function
+ * @signature
+ *    R.mapValues(object, fn)
+ * @example
+ *    R.mapValues({a: 1, b: 2}, (value, key) => value + key) // => { a: 1a, b: 2b }
```
lstkz, line 8: quotes are missing in `{ a: 1a, b: 2b }`

src/mapValues.ts
```diff
@@ -0,0 +1,43 @@
...
+  fn: (value: T[keyof T], key: keyof T) => S
+): Record<keyof T, S>;
+
+/**
+ * Maps values of `object` and keeps the same keys.
+ * @param fn the mapping function
+ * @signature
+ *    R.mapKeys(fn)(object)
+ * @example
+ *    R.pipe({a: 1, b: 2}, R.mapValues((value, key) => value + key)) // => { a: 1a, b: 2b }
```
lstkz, line 23: quotes are missing in `{ a: 1a, b: 2b }`

lstkz (changes requested):

src/mapValues.ts
TkDodo: did that here: 7e309318

I also removed the spaces around the brackets, so that it is the same formatting as in other examples.

src/mapValues.ts
TkDodo: also here: 7e309318

lstkz:
@TkDodo thanks, that was a quick fix :)

TkDodo:
Thanks for merging 🎉

src/mapValues.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @param fn the mapping function
+ * @signature
+ *    R.mapKeys(fn)(object)
+ * @example
+ *    R.pipe({a: 1, b: 2}, R.mapValues((value, key) => value + key)) // => {a: '1a', b: '2b'}
+ * @data_last
+ * @category Object
+ */
+export function mapValues<T extends object, S>(
+  fn: (key: keyof T, value: T[keyof T]) => S
```
TkDodo, line 28: I made a terrible mistake here: The function should actually take the `value` first and the `key` as second parameter, just like [here](https://github.com/remeda/remeda/pull/86/files#diff-e107e2315272076f1de4296bd17beb11R14).

The tests pass because the "runtime" function does it right :/

I'll make a PR to fix this immediately.

TkDodo: here is the PR: https://github.com/remeda/remeda/pull/93 🙈

PR #94, opened 2020-09-12
Feature/map to object

TkDodo:
This PR adds two utility functions for transforming an Array to an Object, which is something that I think many utility libraries are missing, yet it is a use-case that I have found to come up often (at least for me). `lodash` has [keyBy](https://lodash.com/docs/4.17.15#keyBy), but it doesn't allow for value transformation at the same time. I don't think ramda has anything like that. People often use `reduce` for that or `map + fromEntries`.

I named those functions `mapToObj` and `flatMapToObj` - the iteratee functions expect you to return one (for map), or many (0-n, for flatMap) key-value pairs as tuples.


TkDodo:
personally, I'm more of a fan of no abbreviations, so I would actually want to name the methods:

- `mapToObject`
- `flatMapToObject`

But since there are already a couple of methods with the `obj` abbreviation for object (`objOf` and `forEachObj`), I decided to also use this naming pattern here. What do you think?

TkDodo:
any input on this @lstkz?

lstkz (changes requested):
Thanks, looks good! Can you add also exports to `index.ts?`.

TkDodo:
> Thanks, looks good! Can you add also exports to `index.ts?`.

right, I did that here: ae712d32

Maybe we should (in another PR) have the tests import from the `index.ts` barrel rather than from the source files directly? That would make sure that we cannot forget that in the future 😅 .

lstkz (approved):

lstkz:
thanks, good work!

Issue #68, opened 2020-05-14
groupBy hidden bug


digimuza:
Currently, the group by function has hidden BUG.

```typescript

interface User {
     prop1: string
}
const data: User[] = [{ prop1: "a" }, { prop1: "b" } ]
R.pipe(
    data,
    R.groupBy((c)=>c.prop1)
) //  { a: [{ prop1: "a" }], b: [ {prop1: 'b'} ] }

// Change in code 
interface User {
     prop1: string[]
}
R.pipe(
    data,
    R.groupBy((c)=>c.prop1)
) //  { "[object Object]": [{ prop1: "a" }, {prop1: 'b'}] }

// Types not complaining no runtime error.
```

Expecting: Typescript should complain that this is not a valid input

TkDodo:
>  to only allow a string for the return type.

I think it should be a bit more exact than that. Maybe `T extends string | number | symbol`, as those are the three things allowed as keys in an object. `mapKeys` had the same issue before I adapted the types, compare here:

https://github.com/remeda/remeda/pull/85

Would you be willing to make a PR?

MikeyBurkman:
@TkDodo As long as the return type is Record<string, T[]>, you won't be able to have the arg return anything other than a string. Though I don't think that would be a breaking change? 
```ts
export function groupBy<T, K extends string|number|boolean|symbol>(
  items: readonly T[],
  fn: (item: T) => K
): Record<K, T[]>;
```

Also, keys in JS objects are always strings (or symbols I suppose), and numbers are booleans are implicitly converted to strings, which could cause some confusing behavior maybe. `obj[true]` and `obj['true']` are the same thing. Lodash has this same issue that has bit me a few times. For instance, `Object.keys(obj)` will return a list of strings, regardless of whether I used numbers or whatever as the grouping key. At work I ended up just writing my own groupBy function that uses a Map as the backing data structure, so that any primitive could be used as the key, though I'm not sure if Remeda really is in the business of trying to define new data structures.

TkDodo:
@MikeyBurkman I think your example code is correct. I would still stick to `T extends string | number | symbol`, as to me, this seems what typescript thinks a PropertyKey of an object can be, according to the lib.es5.d.ts:

https://github.com/microsoft/TypeScript/blob/5305e4a44e4e08c911d181e207602bd4cf335ba1/lib/lib.es5.d.ts#L105

[...]

PR #95, opened 2020-09-13
Feature/reverse for tuples

TkDodo:
currently, calling `reverse` an on a tuple, e.g. `[string, number, boolean]` will return `[string | number | boolean][]` - which means any position can have any of those three types. But actually, because we have a fixed length, we can know which item will have which type.

so with this PR, `[string, number, boolean]` will be reversed to `[boolean, number, string]`, which I think is nice. The usages for arrays remain intact, so a `number[]` will still be reversed to `number[]`.

I also added compile time tests, which will make sure that the functions return the correct type depending on the input. The `AssertEqual` util will yield a compile time error, so I enforced that check with a `compile` script that is called during the test phase. Note that the actual `expect` is only there to make sure that we have a usage of `const result` 🤷 .

### DIFF
--- a/src/reverse.ts
+++ b/src/reverse.ts
@@ -1,11 +1,18 @@
 import { purry } from './purry';
 
 // prettier-ignore
-type ReverseTuple<T extends readonly unknown[], R extends readonly unknown[] = []> = ReturnType<
-  T extends readonly [infer F, ...infer L] ? () => ReverseTuple<L, [F, ...R]> : () => R
+type Reverse<
+  T extends readonly unknown[],
+  R extends readonly unknown[] = []
+> = ReturnType<
+  T extends IsNoTuple<T>
+    ? () => [...T, ...R]
+    : T extends readonly [infer F, ...infer L]
+    ? () => Reverse<L, [F, ...R]>
+    : () => R
 >;
 
-type IsTuple<T> = T extends readonly [unknown, ...unknown[]] ? T : never;
+type IsNoTuple<T> = T extends readonly [unknown, ...unknown[]] ? never : T;
 
 /**
  * Reverses array.
@@ -17,9 +24,7 @@ type IsTuple<T> = T extends readonly [unknown, ...unknown[]] ? T : never;
  * @data_first
  * @category Array
  */
-export function reverse<T extends readonly unknown[]>(
-  array: T
-): T extends IsTuple<T> ? ReverseTuple<T> : T;
+export function reverse<T extends readonly unknown[]>(array: T): Reverse<T>;
 
 /**
  * Reverses array.
@@ -33,7 +38,7 @@ export function reverse<T extends readonly unknown[]>(
  */
 export function reverse<T extends readonly unknown[]>(): (
   array: T
-) => T extends IsTuple<T> ? ReverseTuple<T> : T;
+) => Reverse<T>;
 
 export function reverse() {
   return purry(_reverse, arguments);
--- a/src/reverse.ts
+++ b/src/reverse.ts
@@ -1,24 +1,9 @@
 import { purry } from './purry';
 
-type Prepend<Tuple extends readonly unknown[], Added> = ((
-  _: Added,
-  ..._1: Tuple
-) => unknown) extends (..._: infer Result) => unknown
-  ? Result
-  : never;
-
-type Reverse<
-  Tuple extends readonly unknown[],
-  Prefix extends readonly unknown[] = []
-> = {
-  0: Prefix;
-  1: ((..._: Tuple) => unknown) extends (
-    _: infer First,
-    ..._1: infer Next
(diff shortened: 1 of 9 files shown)


lstkz:
@TkDodo wow, this is nice :) 
What TS version is required for it? Does it require variadic tuples from 4.0?

TkDodo:
> @TkDodo wow, this is nice :)
> What TS version is required for it? Does it require variadic tuples from 4.0?

I made this so it works with 3.8. It doesn't support variadic tuples because they don't exist in 3.8, the results with this for variadic tuples are actually quite bad. `[number, ...Array<number>].reverse()` will yield `[number]`, which is wrong. ideally, it would yield `Array<number>`, but I couldn't make that work :/

If we want to make this lib 4.0 compatible, I think we should add `typescript: ^4.0.0` to the `peerDependencies` so that people know that this version requires ts 4.0

There is also an easier approach to all of this that requires ts 4.0, please see [this stackoverflow answer](https://stackoverflow.com/a/63326883)

I haven't tried it out yet, but I'm happy to. I just wasn't sure how the policy is for this lib regarding upgrading the requirements for the minimal ts version...

lstkz:
I am fine with bumping to 4.0, and using the simpler version. Also, we can skip tests for type checking because other functions don't have it.
Thanks!

TkDodo:
> I am fine with bumping to 4.0, and using the simpler version.

right, I will give it at try, not sure if I find the time today though. Are you fine with adding the `peerDependency` as well? It might not be such a good idea after all for people using only `javascript`, they would get a warning when installing... could we also communicate this in another way?

> Also, we can skip tests for type checking because other functions don't have it.

I can, if you insist, but I don't see the harm in those tests, and I would rather add _more_ tests on type level for other functions as well. Issues like this one for example (https://github.com/remeda/remeda/issues/68) can only be tested with a  type level test.

TkDodo:
I updated the PR, please have another look at @lstkz. It is now much simpler and also works on variadic tuples.

For `[number, ...Array<string>]`, it will infer `(string | number)[]`. I don't think we can type that any better, because we don't know how many strings there are and that the last element will be a `number` is kind of lost, but I think that's better than the alternative.

The tests were actually super helpful in developing this, because I was breaking thins a lot on type level while trying to find the right solution :)

I am seeing TS errors now in `clone`, probably due to the typescript upgrade - let me check them

TkDodo:
I made the least invasive possible fix here: 0a421631

src/reverse.ts
```diff
@@ -1,5 +1,19 @@
 import { purry } from './purry';
 
+// prettier-ignore
```
TkDodo, line 3: FYI, prettier keeps adding brackets around `(...infer L)`, and typescript doesn't like that ... updating prettier to the latest version didn't fix this, so I didn't do that. I will try to find an open issue in `prettier` or create one if I don't find anything

TkDodo: So I'm pretty sure this is the issue: https://github.com/prettier/prettier/issues/9041

It has been solved already, and it should actually be fixed with 2.1.1: https://github.com/prettier/prettier/blob/master/CHANGELOG.md#fix-broken-format-for-infer-t-9044-by-fisker

I'll try uprading again

TkDodo: right, I actually just needed to restart my editor to make it pick up the new prettier 🙄 .

upgraded and re-formatted here: 48cc6c1e

lstkz:
@TkDodo thanks for the quick fixes. Good work :)

src/reverse.ts
```diff
@@ -1,11 +1,18 @@
 import { purry } from './purry';
 
 // prettier-ignore
-type ReverseTuple<T extends readonly unknown[], R extends readonly unknown[] = []> = ReturnType<
-  T extends readonly [infer F, ...infer L] ? () => ReverseTuple<L, [F, ...R]> : () => R
+type Reverse<
+  T extends readonly unknown[],
+  R extends readonly unknown[] = []
+> = ReturnType<
```
olsonpm, line 6: @TkDodo Hey I was looking at how Reverse was implemented in remeda and [was told to ping you](https://github.com/remeda/remeda/issues/894#issuecomment-2422846205) about the usage of ReturnType here.

Mind explaining ?  To be clear, I'd expect the type to be written as

```ts
type Reverse<
  T extends readonly unknown[],
  R extends readonly unknown[] = []
> =
  T extends IsNoTuple<T>
    ? [...T, ...R]
    : T extends readonly [infer F, ...infer L]
      ? Reverse<L, [F, ...R]>
      : R
```

I figure there's some typescript behavior being manipulated and was hoping you could explain what.

olsonpm: Haha thanks so much for taking the time to respond.  Understandable to not remember !

Issue #81, opened 2020-07-28
Pick returns undefined values for keys not in source object


cilquirm:
Hi Remeda Team,

This is more of a clarification than an actual bug, as i don't know what the expected response is supposed to be.

Right now, if you use `pick` to pull properties from an object:
```javascript
const R = require('remeda')

const obj = { a: 1 , b: 2 } 

R.pick(obj, ['a', 'c'])
> Object {a: 1, c: undefined} 
```

`pick`, in Ramda, effectively ignores that property and does not return it in the resulting object:
```javascript
const R = require('ramda')

const obj = { a: 1 , b: 2 } 

R.pick(obj, ['a', 'c'])
> Object {a: 1 } 
```

Here's a runkit link with the 2 side by side: https://runkit.com/cilquirm/5f205d2827d3e10013e2fb45

Should this be considered a bug?  I think it's a strange side effect to add properties that aren't present in the source object, and it's definitely tripped me up before ( not to mention it doesn't jive with the Ramda analog ).

Happy to open a PR to resolve with tests if you think this makes sense.

TkDodo:
So, trying out your example on type level, I am getting a type error because pick is strictly typed and you can only pick keys that are keyof your object:

![Screenshot 2020-09-23 at 21 25 37](https://user-images.githubusercontent.com/1021430/94059949-75900e00-fde3-11ea-858d-84af8725870f.png)

I am not sure what the stance is here regarding javascript users, or if you put in `any`.

We could see this as a case of `garbage in -> garbage out`, but it also doesn't hurt to provide runtime correctness here. There are also tests that test for `null` or `undefined` input.

Since @lstkz already added the bug label, I think it's faire to assume that a PR is welcome :)

PR #98, opened 2020-10-01
[remeda/remeda#81] Do not return nonexistent source properties from pick

steevsachs:
Test for existence of object property before including it in returned object. 

As part of this PR I upgraded `jest` and `@types/jest` in order to pick up the `toStrictEqual` matcher (since `toEqual` treats `undefined` the same as nonexistent).

I ran all tests and compiled to confirm no adverse impact of the library upgrades.


lstkz:
thanks!

Issue #72, opened 2020-06-08
Add "zipObj" function


dodas:
Are we interested in `zipObj(keys, values)` function? I use it quite frequently and would like to have it in remeda.

If so, I can send an PR.

jbpotonnier:
The `zip` and `zipWith` functions would be nice too.
Maybe add theses first.

Then add a function that goes the opposite direction than `toPairs`. 
I think it would be more consistent and composable.

steevsachs:
Created the ingredients I think we need to get to `zipObj`.  If these PRs land, `zipObj` (I think just a pipe through `zip` (or `zipWith` for `zipObjWith`?) and then `fromPairs` should hopefully be pretty straightforward).

eranhirsch:
Feels like this has been covered by other functions and the ability to build a pipe for it.

PR #99, opened 2020-10-01
Add `zip` utility

steevsachs:
Per remeda/remeda#72, adds the `zip` utility:

*Data first*
```
R.zip([1, 2], ['c', 'd']) // => [[1, 'c'], [2, 'd']
```
*Data last*
```
R.zip(['c', 'd'])([1, 2]) // => [[1, 'c'], [2, 'd']
```

### DIFF
--- a/src/zip.ts
+++ b/src/zip.ts
@@ -12,7 +12,7 @@ import { purry } from './purry'
  * @data_first
  * @category Array
  */
-export function zip<F extends Array<unknown>, S extends Array<unknown>>(first: F, second: S): Array<[F, S]>
+export function zip<F extends unknown, S extends unknown>(first: Array<F>, second: Array<S>): Array<[F, S]>
 
 /**
  * Creates a new list from two supplied lists by pairing up equally-positioned items.
@@ -25,7 +25,7 @@ export function zip<F extends Array<unknown>, S extends Array<unknown>>(first: F
  * @data_last
  * @category Array
  */
-export function zip<F extends Array<unknown>, S extends Array<unknown>>(second: S): (first: F) => Array<[F, S]>
+export function zip<F extends unknown, S extends unknown>(second: Array<S>): (first: Array<F>) => Array<[F, S]>
 
 export function zip() {
   return purry(_zip, arguments)
--- a/src/zip.ts
+++ b/src/zip.ts
@@ -8,7 +8,7 @@ import { purry } from './purry'
  * @signature
  *   R.zip(first, second)
  * @example
- *   R.zip([1, 2], ['c', 'd']) // => [[1, 'c'], [2, 'd']
+ *   R.zip([1, 2], ['a', 'b']) // => [1, 'a'], [2, 'b']
  * @data_first
  * @category Array
  */
@@ -21,11 +21,11 @@ export function zip<F extends Array<unknown>, S extends Array<unknown>>(first: F
  * @signature
  *   R.zip(second)(first)
  * @example
- *   R.zip(['c', 'd'])([1, 2]) // => [[1, 'c'], [2, 'd']
+ *   R.zip(['a', 'b'])([1, 2]) // => [[1, 'a'], [2, 'b']
  * @data_last
  * @category Array
  */
-export function zip<F extends Array<unknown>, S extends Array<unknown>>(first: F): (second: S) => Array<[F, S]>
+export function zip<F extends Array<unknown>, S extends Array<unknown>>(second: S): (first: F) => Array<[F, S]>
 
 export function zip() {
   return purry(_zip, arguments)
--- a/src/zip.ts
+++ b/src/zip.ts
@@ -1,4 +1,4 @@
-import { purry } from './purry'
+import { purry } from './purry';
 
 /**
  * Creates a new list from two supplied lists by pairing up equally-positioned items.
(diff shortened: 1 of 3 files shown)


src/zip.ts
```diff
@@ -0,0 +1,42 @@
...
+ * The length of the returned list will match the shortest of the two inputs.
+ * @param second the second input list
+ * @signature
+ *   R.zip(second)(first)
+ * @example
+ *   R.zip(['a', 'b'])([1, 2]) // => [[1, 'a'], [2, 'b']
+ * @data_last
+ * @category Array
+ */
+export function zip<F extends Array<unknown>, S extends Array<unknown>>(second: S): (first: F) => Array<[F, S]>
```
TkDodo, line 28: I don’t think the return type of `Array<[F, S]>` is correct.
This only works for inputs that have the same length. Otherwise, they are filled up with `undefined`.

It would also be super nice to make it work with Tuples as input, so if you do:

`zip(['foo', 'bar'] as const, [1, 2] as const)`

It would yield:

`['foo', 1], ['bar', 2]` _on type level_

TkDodo (changes requested):

src/zip.ts
TkDodo: Maybe we could even make zip expect an arbitrary amount of array as parameters, as shown here: https://stackoverflow.com/questions/51977209/can-a-typed-zip-function-be-implemented-now-with-typescript-3-0s-generic-rest-p

steevsachs: Wow, you're right, that type's not correct, I'll revise that. 

The logic here does handle `first` and `second` being different lengths--can you provide inputs for a test case that would produce `undefined` values?

TkDodo: Ah right, I mis-read the implementation. You are taking the _shorter_ length [here](https://github.com/remeda/remeda/pull/99/files#diff-3ba70535993f56e0474d4976a9e95c92R35)

So, when looking at the `lodash` implmentation, it seems that always the larger length is taken, and the missing values are filled with `undefined`:

Screenshot 2020-10-09 at 21 37 21

You can also see that lodash supports a variadic argument list of arrays.

steevsachs: Ok, so I was working from the ramda implementation https://ramdajs.com/docs/#zip. It appears to be quite different from lodash. 

Is there a preference here of which to match when they disagree?

IMO, the `zipObj`/`zipObject` implementation of ramda/lodash is more consistent with the ramda definition of `zip` than the `lodash` definition.

TkDodo: Right - i just looked up how `zip` works in other languages, e.g. [scala](https://www.scala-lang.org/api/current/scala/collection/AbstractIterable.html#zip[B](that:scala.collection.IterableOnce[B]):CC[(A@scala.annotation.unchecked.uncheckedVariance,B)]), or [haskell](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-List.html#g:17), and it seems that they all also correspond with the `ramda` logic:
- only allow two lists as input
- discard elements from lists that are longer

so I think your implementation is fine if we stick to that 👍

src/zip.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @param first the first input list
+ * @param second the second input list
+ * @signature
+ *   R.zip(first, second)
+ * @example
+ *   R.zip([1, 2], ['a', 'b']) // => [1, 'a'], [2, 'b']
+ * @data_first
+ * @category Array
+ */
+export function zip<F extends unknown, S extends unknown>(first: Array<F>, second: Array<S>): Array<[F, S]>
```
TkDodo, line 15: ah, we now need to make that `ReadonlyArray` please. Otherwise, we cannot input readonly array, e.g. derived from `const assertions`, see this [playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAMwK4DsDGMCWEVwF6ZgA8AYnKDMCgCYDOcqA1ihAO4oA0cAyuSJTXpMW7AHwAKBJii0YALjgBBKFACGAT1KiutYGhzUFytZu6iAlEZUaiAbRJduAXVFwA3gFgAUHF9wA9P5wskgARnCYALZgADZcrMBwODHqcGiqUImqoRBI8DAAFokw6mDAtN5+cJkwSFC4tk5wqvSqKOreAL7e3voosogQEHAAvHC2AOQIQxNcE6EZE00taTiyvWvwC3ij4wCMXABMy-R9614b-fCZtEgx8GMEYJJDXNvmQA)

The good thing about `readonly` is that you can always pass a mutable Array into it, but it doesn't work the other way around. The return value should still be an `Array`.

Maybe we could add some type-level tests like we [do them here](https://github.com/remeda/remeda/blob/17e9b0e2985d26bc6689a77a4e192baa80fcc29f/src/reverse.test.ts#L11-L15) ?

src/zip.ts
TkDodo: > It would also be super nice to make it work with Tuples as input, so if you do:
>
> zip(['foo', 'bar'] as const, [1, 2] as const)
> It would yield:
> ['foo', 1], ['bar', 2] on type level

I looked into this a bit, but for me it's _very hard_ to do that. The currently returned types are `["foo" | "bar", 2 | 1][]` and  I think that is good enough 👍

steevsachs: Yeah, I've been banging my head against this and I can't seem to get it to infer the exact types of returned tuples.  If you do have any ideas, suggestions definitely welcome!

src/zip.ts
steevsachs: Added those.  Thank you for your patience and the detailed reviews.  I have a few PRs open here--I'm planning to apply the takeaways here to those PRs since I think many of these items will apply.

TkDodo (approved):
lgtm

lstkz:
sorry for long reply, good work!

PR #102, opened 2020-10-05
add minBy and maxBy

lstkz:


tomaszzmudzinski:
@lstkz please fix conflict xDDDD

PR #106, opened 2020-11-03
Fix max by tests

santialbo:


lstkz:
thanks, good catch :wink:

PR #114, opened 2020-12-29
Add "tap" function

acro5piano:
Hi, thank you for the great library. I'm a fan of Remeda.

I would like to include `tap`, which is an utility function acting as a transducer.

Useful for:

- Add logging of current operation.

```ts
R.pipe(
  [1, 2, 3],
  R.filter(item => item > 1),
  R.tap(array => console.log(array)), // => logs "[2, 3]"
  R.map(item => item * 2),
  //...
)
```

- Run exception in operation.

```ts
function validateInvalidValue(array: any[]) {
  if (array.includes('invalid')) {
    throw new Error('Invalid value contained')
  }
}

R.pipe(
  [1, 2, 3, "invalid"],
  R.tap(validateInvalidValue), // throws "Invalid value contained"
  //...
)
```

Inspired by: https://ramdajs.com/docs/#tap

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -57,6 +57,7 @@ export * from './splitAt';
 export * from './splitWhen';
 export * from './take';
 export * from './takeWhile';
+export * from './tap';
 export * from './times';
 export * from './toPairs';
 export * from './type';
--- a/src/tap.test.ts
+++ b/src/tap.test.ts
@@ -1,9 +1,11 @@
 import { tap } from './tap';
 import { pipe } from './pipe';
+import { take } from './take';
+
+const mockFn = jest.fn(() => void 0);
 
 describe('data_first', () => {
   it('tap', () => {
-    const mockFn = jest.fn(() => void 0);
     expect(tap(1, mockFn)).toBe(1);
     expect(mockFn).toBeCalledWith(1);
   });
@@ -11,8 +13,15 @@ describe('data_first', () => {
 
 describe('data_last', () => {
   it('tap', () => {
-    const mockFn = jest.fn(() => void 0);
     expect(pipe(1, tap(mockFn))).toBe(1);
     expect(mockFn).toBeCalledWith(1);
+  });
+
+  it('tap lazily', () => {
+    expect(pipe([1, 2, 3, 4, 5], tap(mockFn), take(3))).toEqual([1, 2, 3]);
+    expect(mockFn).toBeCalledWith(1);
+    expect(mockFn).toBeCalledWith(2);
+    expect(mockFn).toBeCalledWith(3);
+    expect(mockFn).toHaveBeenCalledTimes(3);
   });
 });
--- a/src/tap.test.ts
+++ b/src/tap.test.ts
@@ -1,0 +1,18 @@
+import { tap } from './tap';
+import { pipe } from './pipe';
+
+describe('data_first', () => {
+  it('tap', () => {
+    const mockFn = jest.fn(() => void 0);
+    expect(tap(1, mockFn)).toBe(1);
+    expect(mockFn).toBeCalledWith(1);
+  });
+});
+
+describe('data_last', () => {
+  it('tap', () => {
+    const mockFn = jest.fn(() => void 0);
+    expect(pipe(1, tap(mockFn))).toBe(1);
+    expect(mockFn).toBeCalledWith(1);
+  });
+});
--- a/src/tap.ts
+++ b/src/tap.ts
@@ -1,4 +1,5 @@
 import { purry } from './purry';
+import { _reduceLazy, LazyResult } from './_reduceLazy';
 
 /**
  * Runs the given function with the supplied object, then returns the object.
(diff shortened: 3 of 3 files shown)


lstkz (changes requested):
@acro5piano thanks for the PR, I will accept it.
The current version won't work when using `.pipe`. Can you create something similar to https://github.com/remeda/remeda/blob/master/src/uniq.ts?

acro5piano:
@lstkz Thank you for your review. I'll fix it!

acro5piano:
@lstkz 
Sorry, I look into the `uniq` and couldn't understand the point...

- Do you want me to implement `lazy` feature?
- As far as tested on my PC, `tap` works with `R.pipe`. Can you tell me in which case it won't work?

Thank you.

lstkz:
@acro5piano 
Consider the following example
```ts
R.pipe(
  [1, 2, 3, 4, 5],
  R.tap(console.log),
  R.take(3)
);
```
Only `1, 2, 3` should be logged. Without a lazy version, it would log all items of the array.
It can be implemented similar to `uniq`, but `set` and first `if` statement are not needed.

acro5piano:
@lstkz Oh okay. Thank you for your explanation. I thought it should work not lazily (useful for debug intermediate values) but I'll follow you and implement lazy!

acro5piano:
@lstkz Hi, I've tried to implement lazy feature, but gave up. I'm really sorry, it will take several hours to implement it. Feel free to close this PR.

lstkz:
@acro5piano 
Don't give up, it should be easy :)

Something like this:

```ts
export namespace tap {
  export function lazy(fn: (value: any) => any) {
    return (value: any): LazyResult<any> => {
      fn(value);
      return {
        done: false,
        hasNext: true,
        next: value,
      };
    };
  }
}

```

acro5piano:
@lstkz Thank you for your kind instruction. I'm really sorry, but stacking on how to pass arguments to lazy function.

I wrote this test case:

```ts
  it.only('tap lazily', () => {
    expect(pipe([1, 2, 3, 4, 5], tap(mockFn), take(3))).toEqual([1, 2, 3]);
    expect(mockFn).toBeCalledWith([1, 2, 3]);
  });
```

And following your code, current my implementation is:

```ts
export function tap() {
  return purry(_tap, arguments, tap.lazy);
}

export function _tap<T>(x: T, fn: (x: T) => void) {
  return tap.lazy(fn)(x);
}

export namespace tap {
  export function lazy(fn: (value: any) => any) {
    return (value: any): LazyResult<any> => {
      fn(value);
      return {
        done: false,
        hasNext: true,
        next: value,
      };
    };
  }
}
```

But it does not work. Test result is like this:

image

I tried the `_reduceLazy` referring to `uniq` and `uniqBy`, but it the result is the same.

```ts
export function _tap<T>(x: T[], fn: (x: T) => void) {
  return _reduceLazy(x, tap.lazy(fn));
}
```

Do you have any idea of it?
So sorry to bother you, and thank you for your help.

acro5piano:
If I change the fn in `tap` to log values to this,

```ts
  it.only('tap lazily', () => {
    expect(
      pipe(
        [1, 2, 3, 4, 5],
        tap(x => console.log(x)),
        take(3)
      )
    ).toEqual([1, 2, 3]);
    expect(mockFn).toBeCalledWith([1, 2, 3]);
  });
```

the output is:

image

I think this is because `pipe` pass arguments as spread. I'm really sorry but couldn't find how to pass arguments as array...

https://github.com/acro5piano/remeda/blob/fe2df0d136a450cbd43a315473cb4e62043ef2e3/src/pipe.ts#L78-L81

lstkz:
@acro5piano 
this is wrong `expect(mockFn).toBeCalledWith([1, 2, 3]);`, it expect a single call with 3 arguments. You can change it to something like this
```ts
expect(mockFn).toBeCalledWith(1);
expect(mockFn).toBeCalledWith(2);
expect(mockFn).toBeCalledWith(3);
```

or 
```ts
expect(mockFn).toHaveBeenCalledTimes(3).
```

acro5piano:
@lstkz Thanks! It works.

However, non-lazy version not working this time...

```ts
export function tap() {
  return purry(_tap, arguments, tap.lazy);
}

export function _tap<T>(fn: (x: T) => void, x: any) {
  return tap.lazy(fn)(x);
}

export namespace tap {
  export function lazy(fn: (value: any) => any) {
    return (value: any): LazyResult<any> => {
      fn(value);
      return {
        done: false,
        hasNext: true,
        next: value,
      };
    };
  }
}
```

```ts
import { tap } from './tap';
import { pipe } from './pipe';
import { take } from './take';

const mockFn = jest.fn(() => void 0);

describe('data_first', () => {
  it('tap', () => {
    expect(tap(1, mockFn)).toBe(1);
    expect(mockFn).toBeCalledWith(1);
  });
});

describe('data_last', () => {
  it('tap', () => {
    expect(pipe(1, tap(mockFn))).toBe(1);
    expect(mockFn).toBeCalledWith(1);
  });

  it('tap lazily', () => {
    expect(pipe([1, 2, 3, 4, 5], tap(mockFn), take(3))).toEqual([1, 2, 3]);
    expect(mockFn).toBeCalledWith(1);
    expect(mockFn).toBeCalledWith(2);
    expect(mockFn).toBeCalledWith(3);
    expect(mockFn).toHaveBeenCalledTimes(3);
  });
});
```

image

lstkz:
You can't call the lazy function because it will return a different result.
Just do
```ts
export function _tap<T>(fn: (x: T) => void, x: any) {
  fn(x);
  return x
}
```

acro5piano:
@lstkz Thanks, I've tried but couldn't pass the test. I pushed current code to make things clear.

TkDodo:
How are we looking here please? Are tests failing?

acro5piano:
@TkDodo 
I gave up to implement it. Feel free to close this PR.

PR #129, opened 2021-06-10
Adds support for multiple sort by projections as well as directions

bennidhamma:
Thank your for this wonderful library!

I often need to be able to specify multiple fields to project a sort on, in addition to parameters. I would like to add this behavior to remeda.

Can be used like:

```ts
const objects = [
  { color: 'red', weight: 2 },
  { color: 'blue', weight: 3 },
  { color: 'green', weight: 1 },
  { color: 'purple', weight: 1 },
];

R.sortBy(objects, x => x.weight, x => x.color);
R.sortBy(objects, [x => x.weight, 'desc'], [x => x.color, 'desc']);
R.sortBy(objects, x => x.weight, [x => x.color, 'desc']);
```

### DIFF
--- a/src/sortBy.ts
+++ b/src/sortBy.ts
@@ -1,29 +1,57 @@
 import { purry } from './purry';
 
+type Direction = 'asc' | 'desc';
+type SortProjection<T> = (x: T) => Sortable;
+type SortablePrimitive = number | string;
+type Sortable = SortablePrimitive | { valueOf(): SortablePrimitive };
+
 /**
- * Sorts the list according to the supplied function in ascending order.
+ * Sorts the list according to the supplied functions and directions.
  * Sorting is based on a native `sort` function. It's not guaranteed to be stable.
+ *
+ * Directions are applied to functions in order and default to ascending if not specified.
  * @param array the array to sort
- * @param fn the mapping function
+ * @param fnDirs a list of mapping functions and directions
  * @signature
- *    R.sortBy(array, fn)
+ *    R.sortBy(array, ...fnDirs)
  * @example
  *    R.sortBy(
  *      [{ a: 1 }, { a: 3 }, { a: 7 }, { a: 2 }],
  *      x => x.a
  *    )
  *    // => [{ a: 1 }, { a: 2 }, { a: 3 }, { a: 7 }]
+ *
+ *    R.sortBy(
+ *     const objects = [
+ *       {color: 'red', weight: 2},
+ *       {color: 'blue', weight: 3},
+ *       {color: 'green', weight: 1},
+ *       {color: 'purple', weight: 1},
+ *     ];
+ *      x => x.weight, 'asc', x.color, 'desc'
+ *    )
+ *    // =>
+ *    //   {color: 'purple', weight: 1},
+ *    //   {color: 'green', weight: 1},
+ *    //   {color: 'red', weight: 2},
+ *    //   {color: 'blue', weight: 3},
  * @data_first
  * @category Array
  */
-export function sortBy<T>(array: readonly T[], fn: (item: T) => any): T[];
+export function sortBy<T>(
+  array: readonly T[],
+  ...fnDirs: (SortProjection<T> | Direction)[]
+): T[];
 
 /**
- * Sorts the list according to the supplied function in ascending order.
+ * Sorts the list according to the supplied functions and directions.
  * Sorting is based on a native `sort` function. It's not guaranteed to be stable.
- * @param fn the mapping function
+ *
+ * Directions are applied to functions in order and default to ascending if not specified.
+ * @param fnDir first mapping function or direction
+ * @param fnDirs additional mapping function
  * @signature
- *    R.sortBy(fn)(array)
+ *    R.sortBy(...fnDirs)(array)
  * @example
  *    R.pipe(
  *      [{ a: 1 }, { a: 3 }, { a: 7 }, { a: 2 }],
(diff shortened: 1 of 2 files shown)


TkDodo:
Hmm, is that not what `pipe` is for?

```
R.pipe(
  objects,
  R.sortBy(x => x.weight),
  R.sortBy(x => x.color)
)
```

I am also not sure how the `asc` and `desc` go into the mix here. The type says:

```
...fnDirs: (SortProjection<T> | Direction)[]
```

but that means I can do:

```
R.sortBy(objects, 'asc', 'asc', 'desc', 'asc')
```

bennidhamma:
> Hmm, is that not what `pipe` is for?
> 
> ```
> R.pipe(
>   objects,
>   R.sortBy(x => x.weight),
>   R.sortBy(x => x.color)
> )
> ```

That just resorts by color and invalidates the earlier sort by weight call.

```ts
var arr = [
  {color: 'red', weight: 2},
  {color: 'blue', weight: 3},
  {color: 'green', weight: 1},
  {color: 'purple', weight: 1},
];

arr = R.pipe(
  arr,
  R.sortBy(a => a.weight),
  R.sortBy(a => a.color),
);

// =>  
  [{
    "color": "blue",
    "weight": 3
  },
  {
    "color": "green",
    "weight": 1
  },
  {
    "color": "purple",
    "weight": 1
  },
  {
    "color": "red",
    "weight": 2
  }
]
```
vs:

```ts
var arr = [
  {color: 'red', weight: 2},
  {color: 'blue', weight: 3},
  {color: 'green', weight: 1},
  {color: 'purple', weight: 1},
];

arr = R.pipe(
  arr,
  R.sortBy(a => a.weight, a => a.color)
);

// =>  
  [{
    "color": "green",
    "weight": 1
  },
  {
    "color": "purple",
    "weight": 1
  },
  {
    "color": "red",
    "weight": 2
  },
  {
    "color": "blue",
    "weight": 3
  }
]
```

I agree the direction syntax is bit wonky. It won't do anything particularly pathological in the case you describe. My initial version was like this:

```ts
R.sortby(items, [x => x.weight, x => x.color], ['desc', 'asc']);
```

But I personally prefer the DX of writing

```ts
R.sortby(items, x => x.weight, 'desc', x => x.color, 'asc');
```

bennidhamma:
hrmm... yeah, the backwards compatibility is nice. and, I do think it's a bit tedious to have to put in the brackets. but, I am totally happy to do either approach.

bennidhamma:
we could maybe maintain backwards compatibility with function overloads, if we go with tuples?

bennidhamma:
sounds good. i'll have a go at that :)

src/sortBy.ts
```diff
@@ -32,17 +60,51 @@ export function sortBy<T>(array: readonly T[], fn: (item: T) => any): T[];
...
  */
-export function sortBy<T>(fn: (item: T) => any): (array: readonly T[]) => T[];
+export function sortBy<T>(
+  fnDir: SortProjection<T> | Direction,
+  ...fnDirs: (SortProjection<T> | Direction)[]
+): (array: readonly T[]) => T[];
 
 export function sortBy() {
-  return purry(_sortBy, arguments);
+  if (Array.isArray(arguments[0])) {
```
bennidhamma, line 69: @TkDodo if I switch the signature to use tuples, this test will fail. I confess I don't fully understand purry. Do you have any suggestions about how I can make this test work if the rest args are also array types?

TkDodo: which tests will fail? I suggest you commit it anyways, and then I can take a look at the failing tests.

bennidhamma: I think I have something reasonable now :)

bennidhamma:
This feels pretty close to me. What do you think?

src/sortBy.ts
```diff
@@ -1,29 +1,59 @@
 import { purry } from './purry';
 
+type Direction = 'asc' | 'desc';
+type SortProjection<T> = (x: T) => Sortable;
+type SortablePrimitive = number | string;
+type Sortable = SortablePrimitive | { valueOf(): SortablePrimitive };
```
TkDodo, line 6: two questions:

- what's the  object with the `valueOf` doing?
- previously, the function was allowed to return _any_, now we only allow to sort by numbers or strings I guess. I like that we are being more strict, but what about `booleans` for example? I think they have an order as well ...

src/sortBy.ts
```diff
@@ -1,29 +1,59 @@
...
  *    // => [{ a: 1 }, { a: 2 }, { a: 3 }, { a: 7 }]
+ *
+ *    R.sortBy(
+ *     [
+ *       {color: 'red', weight: 2},
+ *       {color: 'blue', weight: 3},
+ *       {color: 'green', weight: 1},
+ *       {color: 'purple', weight: 1},
+ *     ],
+ *      [x => x.weight, 'asc'], x => x.color
```
TkDodo, line 33: I wasn't aware that we would _also_ allow for mixing the two paradigms - tuples and / or functions, but if it works it's nice. I don't see that we have a test for this mixture though, can you maybe add one?

TkDodo (changes requested):
I think it looks really promising, thanks 🚀

src/sortBy.ts
bennidhamma: valueOf just says that any type that has a valueOf method is a valid type to use as a comparer.

I didn't realize boolean operands be used with non-equal comparison operators. I've added boolean to the list. Should be decently flexible though.

src/sortBy.ts
```diff
@@ -32,17 +62,62 @@ export function sortBy<T>(array: readonly T[], fn: (item: T) => any): T[];
...
+    const array: readonly T[] = arrayOrSort as unknown as readonly T[];
+    return purry(_sortBy, [array, sorts]) as T[];
+  }
+  const sort = arrayOrSort as SortRule<T>;
+  return purry(_sortBy, [[sort, ...sorts]]) as (arr: readonly T[]) => T[];
+}
 
-export function sortBy() {
-  return purry(_sortBy, arguments);
+function isSortRule<T> (x: readonly T[] | SortRule<T>): boolean {
```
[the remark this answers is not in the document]
bennidhamma: got it. very cool.

bennidhamma:
cool. PTAL :)

TkDodo (approved):

TkDodo:
@lstkz I'm happy with the result but would like your final okay on this. It's backwards compatible and adds a nice feature to `sortBy` 🚀

Maybe `sort` should then also be able to accept multiple comparator functions to have the same signature, but imo that could be a follow-up. what do you think?

lstkz:
@TkDodo 
Looks good for me.

> Maybe sort should then also be able to accept multiple comparator functions to have the same signature, but imo that could be a follow-up. what do you think?

Not sure about custom comparators because we can use a native `sort` in a similar way.

bennidhamma:
Thanks y'all! Any idea how / when this might make it in to a published package? Should I submit a PR to bump the version number in package.json?

TkDodo:
I’m still not familiar with the release process. @lstkz can you release it? I can create the release notes manually if we don’t have automation yet…

lstkz:
@TkDodo @bennidhamma 
0.0.28 published

steps are
```
./scripts/build.sh 
npm version patch  
npm publish
```

btw I also fixed the git history because there were many merge commits. I fixed it by making everything linear. Also, I blocked this option in the merge button.

TkDodo:
sorry for screwing this up 🙈

TkDodo:
Hmm why do the tags only go until 0.0.11?

https://github.com/remeda/remeda/tags

lstkz:
@TkDodo 
Fixed. It seems like I forgot to push them 😄

PR #135, opened 2021-07-15
Feature/improve group by typings

### LINKED ISSUE #68: groupBy hidden bug

TkDodo:
closes #68 

by making the iterator fn of groupBy return a PropertyKey, because we need to be able index the resulting object with it

also: make groupBy return a NonEmptyArray so that access to the first element is safe; all created groups by groupBy must have at least on element, otherwise, there wouldn't be a group at all


lstkz:
@TkDodo Why do we need `NonEmptyArray`? I think it's valid only for tuples.

TkDodo:
> @TkDodo Why do we need NonEmptyArray? I think it's valid only for tuples.

when you call groupBy, the resulting arrays are guaranteed to have at least one element. Otherwise, there would be no group at all. So accessing the first element is considered safe. However, when you run typescript with the new (since 4.1) `noUncheckedIndexedAccess` flag, the access will still be considered unsafe. Example:

```
const grouped = R.groupBy(['one', 'two', 'three'], x => x.length) // => {3: ['one', 'two'], 5: ['three']}

const result = grouped.map(group => group[0])
```

I would expect result to be `Array<string>`, but if `groupBy` returns an `Array`, it's actually `Array<string | undefined>`.

`NonEmptyArray` solves this. It's just a tuple with one element and infinite other elements, so accessing `[0]` will yield `string` in the example, but `[1]` is still `string | undefined`. Hope that makes sense. I can also add a types test for the mentioned example if you want?

lstkz:
@TkDodo sounds good. Thanks for explaining!

lstkz (approved):

Issue #136, opened 2021-07-18
Am I doing something wrong when using Object.values?


rarenatoe:
Currently there is no `values` remeda function, so instead I use `Object.values` when I need to get the values of an object without its keys, however this causes the resulting values from `Object.values` to be of type `any[]`.
If I pipe this, the entire chain of excellent inferred typing that remeda does is thrown out of the window.
Here is an example of what I am taking about:
<img width="895" alt="Screen Shot 2021-07-18 at 12 30 04" src="https://user-images.githubusercontent.com/44821537/126076876-46a797c4-272d-4ed5-9476-ee3848c979f0.png">

I have tried using `(o): (A | B | C)[] => Object.values(o)` but it ends up being just a confusing mess going down the chain.

Am I doing something wrong or should I start a feature request for a `values` function?

Useful information:
- I use target: "es5" in tsconfig and cannot change it since my target are web browsers.

TkDodo:
it seems like the "wrong" overload of `Object.values` is chosen, from `lib.es2017.object.d.ts`:

```
    values<T>(o: { [s: string]: T } | ArrayLike<T>): T[];
    values(o: {}): any[];
```

not sure if we can do something about this, I tried the most minimal example:

```
const test = pipe(
    ['one', 'two', 'three'],
    Object.values
)
```

and `test` is `Array<any>` here, too. The types should flow through pipe, so I'm sure adding an `Object.values` as well as an `Object.keys` equivalent to remeda should work. Do you want to give it a try?

rarenatoe:
I would love to.
How should I go about it? Is there a sort of guide? I've never contributed to something like this before.

PR #137, opened 2021-07-21
Created values and keys functions

rarenatoe:
In order to cope with particular issues such as #136, but also to provide a more seamless piping, I implemented the mentioned solutions.

The following is the edits log:

ADDED:
- data-first values() with tests.
- data-first keys() with tests.


src/values.ts
```diff
@@ -0,0 +1,17 @@
...
+  return Object.values(object) as T[];
+}
+
+// data-first
+export function values<T>(object: { [key: string]: T } | ArrayLike<T>): T[];
+
+// data-last
+export function values<T>(): (
+  object: { [key: string]: T } | ArrayLike<T>
+) => T[];
```
TkDodo, line 13: I don't really think we need a data last version of this function. There are also no tests for it. I tried it out with `pipe({ 1: 'one', 2: 'two', 3: 'three' }, values)` and the inferred type is not correct. It looks good when removing the data last overload, and since we also don't have tests for it, I'd suggest to remove it.

src/keys.test.ts
```diff
@@ -0,0 +1,11 @@
...
+
+describe('Test for values', () => {
+  it('should return values of array', () => {
+    expect(keys(['x', 'y', 'z'])).toEqual(['0', '1', '2']);
+  });
+
+  it('should return values of object', () => {
+    expect(keys({ a: 'x', b: 'y', c: 'z' })).toEqual(['a', 'b', 'c']);
+  });
+});
```
TkDodo, line 11: this should be `test for keys` I guess, and `should return keys...`

looks like a copy-paste error from the values tests :)

src/keys.ts
```diff
@@ -0,0 +1,17 @@
+import { purry } from './purry';
+
+function _keys<T>(object: { [key: string]: T } | ArrayLike<T>) {
+  return Object.keys(object) as string[];
+}
+
+// data-first
+export function keys<T>(object: { [key: string]: T } | ArrayLike<T>): string[];
```
TkDodo, line 8: please add js-doc documentation to the functions, because the remeda docs will be auto-created from it. Please have a look at other functions for inspiration.

TkDodo (changes requested):
you would please have to:

- add js-doc
- add the new functions to `src/index.ts` because they won't be exported otherwise
- there are conflicts now, I think it would be better to revert the unrelated formatting changes in this PR. after it is merged, I would create a new PR that just reformats the whole project, and adds a prettier CI check so that wrongly formatted code can't make it to the main branch again.

src/values.ts
rarenatoe: My initial intention of use was for data last, since there is some piping that needs it.

I can further work more on the testing and getting everything worked with data last if it's ok.

TkDodo: With only one argument list, it’s implicitly data last. See the previous example, which works:

`pipe({ 1: 'one', 2: 'two', 3: 'three' }, values)`

with an explicit data last version, I think we'd need to do:

`pipe({ 1: 'one', 2: 'two', 3: 'three' }, values())`

And I don’t see the advantage in that

rarenatoe: Oh I see. I didn't quite understand that. Neat. I'll finish it up then 😄

rarenatoe:
@TkDodo It's done. Let md know what you think.

src/keys.ts
```diff
@@ -0,0 +1,28 @@
...
+ *    R.pipe(
+ *      { a: 'x', b: 'y', c: 'z' },
+ *      R.values,
+ *      R.first
+ *    ) // => 'a'
+ * @pipeable
+ * @category Object
+ */
+// data-first
+export function keys<T>(object: { [key: string]: T } | ArrayLike<T>): string[];
```
TkDodo, line 24: I think a better syntax instead of `{ [key: string]: T }` would be:

`Record<PropertyKey, T>`

`Record` is an alternative way of writing mapped objects, which is sufficient for this case, and keys can be any PropertyKey, which I think expands to `string | number | Symbol`.

src/keys.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function keys() {
+  return purry(_keys, arguments);
```
TkDodo, line 27: I don't think we need `purry` at all if we have no data-last version

src/values.test.ts
```diff
@@ -0,0 +1,11 @@
...
+
+describe('Test for values as data first', () => {
+  it('should return values of array', () => {
+    expect(values(['x', 'y', 'z'])).toEqual(['x', 'y', 'z']);
+  });
+
+  it('should return values of object', () => {
+    expect(values({ a: 'x', b: 'y', c: 'z' })).toEqual(['x', 'y', 'z']);
+  });
+});
```
TkDodo, line 11: can you add a newline here pls

src/values.ts
```diff
@@ -0,0 +1,28 @@
...
+ *    ) // => 'x'
+ * @pipeable
+ * @category Object
+ */
+// data-first
+export function values<T>(object: { [key: string]: T } | ArrayLike<T>): T[];
+
+export function values() {
+  return purry(_values, arguments);
+}
```
TkDodo, line 28: same here as with `keys`: `Record<PropertyType, T>` and `purry` is not needed

src/keys.ts
```diff
@@ -0,0 +1,28 @@
+import { purry } from './purry';
+
+function _keys<T>(object: { [key: string]: T } | ArrayLike<T>) {
+  return Object.keys(object) as string[];
+}
+
+/**
+ * Returns a new array containing the keys of the array or object.
+ * @param object
```
TkDodo, line 9: parameters here say it has to be an `object`, but it can also be an Array as far as I can see. This is how it looks in the docs so it's a bit confusing imo:

Screenshot 2021-07-25 at 20 50 17

see the deploy preview here: https://deploy-preview-137--trusting-lumiere-9c7fad.netlify.app/docs#values

I would say it would be good to rename `object` to maybe `source` and then describe here that it can be an object or an array?

TkDodo (changes requested):

rarenatoe:
Thanks for the comments. One learns something new everyday :)

Let me know what you think.

src/keys.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Returns a new array containing the keys of the array or object.
+ * @param source Either an array or an object
+ * @signature
+ *    R.keys(source)
+ * @example
+ *    R.keys(['x', 'y', 'z']) // => ['1', '2', '3']
+ *    R.keys({ a: 'x', b: 'y', c: 'z' }) // => ['a', 'b', 'c']
+ *    R.pipe(
+ *      { a: 'x', b: 'y', c: 'z' },
+ *      R.values,
```
TkDodo, line 11: this should probably be `R.keys` 😅

TkDodo (changes requested):
looks really good, just one minor thing in the docs and then we're good to go, thanks 🚀

src/keys.ts
rarenatoe: oof, yeah.

TkDodo:
🚀

TkDodo (approved):

PR #140, opened 2021-07-30
fix(doc): drop examples

nghiepit:


docs/src/data.json
```diff
@@ -421,7 +421,7 @@
         "category": "Array",
         "indexed": false,
         "pipeable": true,
-        "example": "R.drop([1, 2, 3, 4, 5], 2) // => [1, 2, 3]\n",
+        "example": "R.drop([1, 2, 3, 4, 5], 2) // => [3, 4, 5]\n",
```
TkDodo, line 424: I think this file is generated from the jsdoc comments of the source code file, can you please fix it there instead, see:

https://github.com/remeda/remeda/blob/2bfecfd1855a880b6c1359754874d78a3e79808a/src/drop.ts#L11

TkDodo (changes requested):

docs/src/data.json
nghiepit: Tks. I've updated

TkDodo (approved):

PR #143, opened 2021-08-13
Feature/improve group by key types

TkDodo:
by inferring the type of the key from the provided function; it must still extend a PropertyKey because that's what an object needs as index


Bessonov:
@TkDodo Wow, great! I had the same idea after wondering about `groupBy` return types. From my point of view, you need to include `undefined`:
```typescript
): Record<PropertyKey, NonEmptyArray<T> | undefined>;
```

But I'm not sure about this change: https://github.com/microsoft/TypeScript/issues/13195 . It could lead to different semantic.

TkDodo:
@Bessonov why would we need to include `undefined`? After we group, the resulting object will only have keys where also values exist for it. That's also the reason why we have a `NonEmptyArray` and not just an `Array` - because accessing the first element is safe, as it will always be there...

Bessonov:
@TkDodo correct me, if I'm wrong, but without `undefined` you get here no error:
```typescript
interface MyType {
	type: 'right' | 'wrong'
}

const myTypes: MyType[] = [
	{ type: 'right' }
]

const grouped = groupBy(myTypes, item => item.type)

grouped.wrong.length <-- fine
```

Bessonov:
Or, I think, using `Partial` would preserve the semantic:
```typescrpt
Partial<Record<PropertyKey, NonEmptyArray<T>>>
```

TkDodo:
@Bessonov you are most certainly right 👍 

However, adding `undefined` to the mix will make it harder to work with when iterating over the object with `Object.entries`. consider:

```
interface MyType {
	type: 'right' | 'wrong'
}

const myTypes: MyType[] = [
	{ type: 'right' }
]

const grouped = groupBy(myTypes, item => item.type)

Object.entries(grouped).forEach(([, value]) => {
    console.log(value[0])
})
```

Here, we should be able to access `value[0]`, but this won't work if we add `undefined` to the mix.

Maybe what I'm trying to do in this PR is not the right thing to do? After all, I think things like this are the reason why TypeScript only returns `string` for `Object.keys` ...

Bessonov:
@TkDodo I've created a [playground](https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true&ts=4.4.4#code/C4TwDgpgBAcg9gOwKIFsygIICcsEMQA8AKgHxQC8UA2kQDRQB0TRVAuqwFAcCWCwEWAGa4AxtACyIIuGgBvDgEhQkAFxQA5Fm4BzABbB1UAD4aA7lkTb1HAL5cRiAM7AoKKTMdrJ0yGwrVFWShlCDVNHX1DO04OAHpYqEcICBRHYLgoUzgsAGsoXDSIAA9IEX4AEw5yiBEAG1wsaEEAVwQy7kQobQtmsAAhEAAFBuBuXFqCAEl+FHoAaQgQKGL+BHK0wYtILFAFkBIACg4oKG4ZzyhG3HLEWqXplLZ6QQQ1A7OUtQeUAEoKMj2tA4PzUwx2YwmACUatlygRAbBEKh0CBsHhCN8SFj7E4XN04L0IOUwaNxv58b0BiSIQc3D4II56B8UP9TjMGCEflwKZBiSMIQxzJYqAAGVhQeJQAgAWmlmWyOUcHB5ROp40FFgQ2gA-AxReLJTK5VlckqVXzweqAEa1ZpW-UShJG+Wm5U9Xlq2oMG123UOw2yl2KrgAeStACsasAGBA+FoGQdzZ6fgxBNkkKJdAcDlR6AA3cbNCCsP7kMjyE4OBCOOC1CAMWpwbQHAu2iD6jkyP4B40KpWVpy1+uN5utotUACMrE7kG7TsDJuDNi5cQSQq1UE6YcjZRjce4DKqNXqjSgLTao06PIGAFU1hBBLwiVMZvNFssiqt1lBNnBtrtFkOY42RSC4rhuBA7igb4njPV4oHeGYvhmUsAUWIEQSgaEHCwOEEXgZA0EwHB8BfFIyBMVpqkfBAiRIHFqzxd0iTvain3Kcl3Vve8aKJWl3EgRkQJZMthJnCAV3NViH3YjVhTFR0pQXPs3QJXlpN48o5K1P0FJ7IMzWY8oNNkn17T0+de1dKSeNM20rV0g1LIM0MIyjPdgHjRxEyMkzaPKFM0ywDMRCzHN80LYtUKgCsoCrGs6wbJsW0ijtOUU50bgZBB1BcRdgPiocktHVKp3EuclKspcV0lYA4AyeoiiPOoGiaVp2ivLiQBDLRtF4cYyJQICTmZMCIGuW57hmWCXjeZlkJSaLf3-EBAWBNRsNhAhloEACQHoAjkWI9FBuxDh4qYtSiR6nR+tqTi1IGG6+oQcZ+PpITmVZZlyu5Iznru7TtH9ZzF0Mq7ygB16vXXHU9QsyqXPNKHrXskHEbB1TCUh3rAbMxyMsDLLHByvKFVcndo1jTyD285HcehwL00zbNcygMcotZWLCsSkcUrbNKu0Jqr+ziwdeeSjnJ2ndL9Mx5cgA) with all three definitions above. It looks like the "Partial"-Definition
```typescript
declare function groupByPartial<Item, Key extends PropertyKey>(
  items: readonly Item[], fn: (item: Item) => Key,
): Partial<Record<Key, NonEmptyArray<Item>>>
```
works as expected. What do you think about that?

> However, adding undefined to the mix will make it harder to work with when iterating over the object with Object.entries. consider:

If I must choose between `groupByOriginal` and `groupByUndefined`, then I would choose `groupByUndefined`, because in this case I think that it's a bug of `Object.entries` and in my practice `groupByUndefined` gives more benefit. But I hope that `groupByPartial` works as expected.

TkDodo:
@Bessonov thanks for digging into this. You are right that the `Partial` solution works - with TS 4.3 and onwards. I was just double checking why it doesn't work when I wrote a test for it, and it seems to be that we are on TS 4.0 still in the repo 😅 .

I will update the PR to use `Partial` and also update the TS version to the latest.

TkDodo:
@Bessonov have another look please

src/groupBy.ts
```diff
@@ -54,13 +54,13 @@ const _groupBy = (indexed: boolean) => <T>(
...
-    array: readonly T[],
-    fn: PredIndexed<T, PropertyKey>
-  ): Record<string, NonEmptyArray<T>>;
-  export function indexed<T, K>(
-    fn: PredIndexed<T, PropertyKey>
-  ): (array: readonly T[]) => Record<string, NonEmptyArray<T>>;
+  export function indexed<Item, Key extends PropertyKey>(
+    array: readonly Item[],
+    fn: PredIndexed<Item, Key>
+  ): Record<Key, NonEmptyArray<Item>>;
```
Bessonov, line 60: I'm not using the indexed version, but shouldn't it have same return type?

src/groupBy.test.ts
```diff
@@ -43,3 +44,38 @@ describe('data last', () => {
...
+    interface MyType {
+      type: 'right' | 'wrong';
+    }
+
+    const myTypes: MyType[] = [{ type: 'right' }];
+
+    const grouped = groupBy(myTypes, item => item.type);
+
+    // @ts-expect-error we can't access keys unconditionally
+    expect(() => grouped.wrong.length).toThrow();
```
Bessonov, line 77: maybe add
```typescript
    // @ts-expect-error other values are not allowed by typings
   expect(grouped.notexists).toBeUndefined();
```

src/groupBy.ts
```diff
@@ -13,14 +13,14 @@ import { NonEmptyArray, PredIndexedOptional, PredIndexed } from './_types';
  * @indexed
  * @category Array
  */
-export function groupBy<T>(
-  items: readonly T[],
-  fn: (item: T) => PropertyKey
-): Record<PropertyKey, NonEmptyArray<T>>;
+export function groupBy<Item, Key extends PropertyKey>(
```
Bessonov, line 16: Not related to this PR, but I think we can replace this definition one day:
```
/** types that may be returned by `keyof` */
export type Key = string | number | symbol;
```
with `PropertyKey`.

Bessonov (approved):
Great, thanks!

src/groupBy.test.ts
TkDodo: 70438f6d

src/groupBy.ts
TkDodo: done: 83e04aac

src/groupBy.ts
TkDodo: thx for seeing that, you are right 🙏 

b66c4167

Bessonov:
Great, I'm fine with it, thank you!

PR #145, opened 2021-08-17
Feature/144 fix guards when input is unknown

### LINKED ISSUE #144: Guards narrow to `never` when starting from `unknown`
all guards narrow to `never` when you have an input that is of type `unknown`.

Given:

`x: unknown`

calling:

`isBoolean(x)`

will result in `x` having the the type `never`.

I would expect `x` to be of type `boolean` then. Note that it works fine if you start from e.g. `boolean | string`

@digimuza maybe you can help us out here?

TkDodo:
closes #144


src/guards.ts
```diff
@@ -107,10 +148,20 @@ export function isArray<T>(
...
   data: T | object
-): data is T extends Record<string, unknown> ? T : never {
-  return !!data && !Array.isArray(data) && typeof data === 'object';
+): data is DefinitelyObject<T> {
+  return (
+    !!data &&
+    !Array.isArray(data) &&
+    !isPromise(data) &&
+    !isDate(data) &&
+    !isError(data) &&
```
TkDodo, line 157: I changed the implementation here because I don't think that we'd want `isObject` to return `true` for a promise, a date or an error, but the `typeof` check does that at runtime. They are now also excluded on type level [here](https://github.com/remeda/remeda/pull/145/files#diff-d340d57cd15a4762ee2c5cdf7d5f3650908cd5c9ff483bd12e8130b22b99bac4R136).

This is debatable of course, but you'd mostly want to check for a "plain object", and you can check all other things with the other guards. If you're against this, we could keep `isObject` as it was, and maybe add another `isPlainObject`(like lodash also has it) that does this check. Please let me know what you think.

src/guards.ts
```diff
@@ -72,10 +97,17 @@ export function isBoolean<T>(data: T): data is Extract<T, boolean> {
  *    R.isPromise('somethingElse') //=> false
  * @category Guard
  */
-export function isPromise<T>(data: T): data is Extract<T, Promise<any>> {
+// @ts-expect-error
+export function isPromise<T>(data: T): data is DefinitelyPromise<T> {
```
TkDodo, line 101: for some reason, the trick with making the parameter a union type (see e.g. `T | string` on _isString_) didn't work here for promises, I can't really pinpoint why, so this is the only ts-expect-error that is left.

TkDodo:
@lstkz could you take a look at the open discussions, especially how the `isObject` guard should work at runtime in your opinion :)

src/guards.ts
lstkz: I think returning`true` for promises, dates, and errors is ok, and most people would expect that.  
Adding `isPlainObject` is ok.

src/guards.ts
lstkz: ok for me

src/guards.ts
TkDodo: okay, I will change the implementation, add the additional `isPlainObject` guard and update the docs accordingly :)

lstkz: great, thanks :)

TkDodo: done. let me know what you think

lstkz: Looks good, but I am not sure if we should add more checks? For example, add `!isRegex(data)`? Does it return false for custom classes?

TkDodo: I think you are right. Given that I'd like to get this merged, I'll remove the `isPlainObject` guard for now. The `isObject` guard looks fine the way it is, and we can always make another PR that adds an `isPlainObject` guard, maybe with a similar implementation like [lodash](https://github.com/lodash/lodash/blob/2f79053d7bc7c9c9561a30dda202b3dcd2b72b90/isPlainObject.js) is it (but with better types)

TkDodo: f793f762

lstkz (approved):

lstkz:
@TkDodo netlify deployment should be fixed now

TkDodo:
thanks. I think there are still some syntax errors in the docs generation. I'll take a look later

PR #146, opened 2021-08-27
add pickBy function

heleg:
Added lodash and ramda pickBy analog.

### DIFF
--- a/src/pickBy.ts
+++ b/src/pickBy.ts
@@ -1,0 +1,47 @@
+import { purry } from './purry';
+
+/**
+ * Returns a partial copy of an object containing only the keys that satisfy the supplied predicate.
+ * @param object the target object
+ * @param pred predicate
+ * @signature R.pickBy(object, (value, key) => true )
+ * @example
+ *    R.pickBy({ a: 1, b: 2, c: 3, d: 4 }, (value) => value === 1 || value === 4) // => { a: 1, d: 4 }
+ * @data_first
+ * @category Object
+ */
+export function pickBy<T extends object>(
+  object: T,
+  pred: (value: T[keyof T], key: keyof T) => boolean
+): Partial<T>;
+
+/**
+ * Returns a partial copy of an object containing only the keys that satisfy the supplied predicate.
+ * @param pred predicate
+ * @signature R.pickBy((value, key) => true )(object)
+ * @example
+ *    R.pipe({ a: 1, b: 2, c: 3, d: 4 }, R.pickBy((value) => value === 1 || value === 4)) // => { a: 1, d: 4 }
+ * @data_last
+ * @category Object
+ */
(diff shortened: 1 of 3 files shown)


TkDodo:
> By the way, I tried to generate docs with `docs:generate-json` and it generated much more than just my function. Is there some kind of autogenerated docs?

I'm not even sure if the docs should be checked in. As far as I'm aware, they will be generated from the jsdocs when the project is built, and there is no need to have that checked in. I've never run this locally on my PRs tbh 😅

src/pickBy.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param pred predicate
+ * @signature R.pickBy(object, (value, key) => true )
+ * @example
+ *    R.pickBy({ a: 1, b: 2, c: 3, d: 4 }, (value) => value === 1 || value === 4) // => { a: 1, d: 4 }
+ * @data_first
+ * @category Object
+ */
+export function pickBy<T extends object>(
+  object: T,
+  pred: (value: any, key: string) => boolean
```
TkDodo, line 15: those types are definitely too wide. Please have a look at how `mapValues` is typed to get inspiration.

src/pickBy.ts
```diff
@@ -0,0 +1,47 @@
...
+): Partial<T>;
```
TkDodo, line 16: Partial is definitely a very wide type. It would be good if the return type could be strictly inferred from what the predicate function returns. So if it returns `true` for a specific key, it will be included in the result.

src/pickBy.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Returns a partial copy of an object containing only the keys that satisfy the supplied predicate.
+ * @param object the target object
+ * @param pred predicate
+ * @signature R.pickBy(object, (value, key) => true )
+ * @example
+ *    R.pickBy({ a: 1, b: 2, c: 3, d: 4 }, (value) => value === 1 || value === 4) // => { a: 1, d: 4 }
+ * @data_first
+ * @category Object
+ */
+export function pickBy<T extends object>(
```
TkDodo, line 13: I think instead of `object`, this could be `T extends Record<PropertyKey, any>`

src/pickBy.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function _pickBy(object: any, pred: (value: any, key: string) => boolean) {
+  if (object == null) {
+    return {};
+  }
+
+  const result = {} as any;
+  for (const key in object) {
+    if (pred(object[key], key)) {
```
TkDodo, line 42: I think we should split the function into a normal version where the predicate function only receives the value and an `indexed` version where both value and key are passed. Have a look at how we do this in `forEachObj`.

TkDodo (changes requested):

PR #149, opened 2021-10-29
feat: add tap function

adrgautier:
Hello, I really like this project and I think a `tap` function would be a useful addition.

It seems the idea to create add the `tap` function has already been dicussed but the PR didn't make it to the end :
https://github.com/remeda/remeda/pull/114

The issue with this PR is that the function suggested by @acro5piano was not pipeable/lazy.

However, I believe the point of this function is to call the callback function with any kind of input value.
If we make the `tap` function pipeable, that means the function will only take an array in input, which is not the point.

One way to make the tap function lazy/pipeable is to use it in combination with the map function :

```ts
pipe([1, 2, 3, 4, 5], map(tap(console.log)), take(3));

// logs 1, 2 and 3 separetely
```

We can also concider to add a `tapArray` function (in addition to the "standard" `tap` function) which would be pipeable.

What do you think ?


TkDodo:
can you add a test please that shows that the scenario described here:

- https://github.com/remeda/remeda/pull/114#issuecomment-754524515

works as expected?

```
R.pipe(
  [1, 2, 3, 4, 5],
  R.tap(console.log),
  R.take(3)
);
```

this should only run the tap function with 1, 2 and 3. Happy to merge if that works 👍

adrgautier:
Hello !

The behavior you are describing can only work if the function `tap` is pipeable. 
My suggestion is that the `tap` function should not be pipeable because it should be able to take **any** value in input.

We need to implement a `tapArray` function (which would be pipeable) in order to fullfil the behavior you are describing.

adrgautier:
Hello @TkDodo ,

So I tried to implement a `tapArray` function without success. My understanding of remeda is that lazy functions are not aware when all operations in the pipe are done. That means that we can't add the `tapArray` function anywhere in the pipe and expect it to call the callback function with the partial array when all operations are done.

```ts
pipe(
    [1, 2, 3, 4, 5], 
    tapArray(callback),
    map(x => x * 2)
    take(3)
);
// unable to call callback with [1, 2, 3]
```

However, ordering the pipe differently and using the non-lazy `tap` function, 
we can perform the expected behavior without doing unnecessary operations :

```ts
pipe(
    [1, 2, 3, 4, 5], 
    take(3),
    tap(callback),
    map(x => x * 2)
);
// callback should be called with [1, 2, 3]
```

What I achieved however is calling the callback function for each item. But in this case the `map` function combined with `tap` will work just fine:

```ts
pipe([1, 2, 3, 4, 5], map(tap(callback)), take(3));
// callback should be called 3 times
```

**TL;DR**

The pipeable `tapArray` function cannot exist with the current `pipe` implementation.   
`tap` should be enough for 99% use cases.

vlad-yakovlev:
Strange result of gh desktop. I'll reopen PR on my fork

PR #178, opened 2022-07-10
Add pickBy

vlad-yakovlev:
Add pickBy, like in [Ramda](https://ramdajs.com/docs/#pickBy) or [Lodash](https://lodash.com/docs/4.17.15#pickBy). Due to TS limitations I've added types for Record-like objects only

### DIFF
--- a/src/pickBy.test.ts
+++ b/src/pickBy.test.ts
@@ -1,0 +1,20 @@
+import { pickBy } from './pickBy';
+import { pipe } from './pipe';
+
+describe('data first', () => {
+  test('it should pick props', () => {
+    const result = pickBy({a: 1, b: 2, A: 3, B: 4}, (val, key) => key.toUpperCase() === key);
+    expect(result).toStrictEqual({ A: 3, B: 4 });
+  });
+  test('allow undefined or null', () => {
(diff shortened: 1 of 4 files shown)


src/pickBy.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+describe('data first', () => {
+  test('it should pick props', () => {
+    const result = pickBy({a: 1, b: 2, A: 3, B: 4}, (val, key) => key.toUpperCase() === key);
+    expect(result).toStrictEqual({ A: 3, B: 4 });
+  });
+  test('allow undefined or null', () => {
+    expect(pickBy(undefined as any, (val, key) => key === 'foo')).toEqual({});
+    expect(pickBy(null as any, (val, key) => key === 'foo')).toEqual({});
+  });
```
TkDodo, line 12: why would we want to allow null or undefined as input? None of the other object methods do this...

vlad-yakovlev: I made same as https://github.com/remeda/remeda/blob/master/src/pick.test.ts#L10. Null/undefined disabled by types but covered with tests.

TkDodo: oh I didn't know that `pick` also allows it 👍

TkDodo (approved):

PR #180, opened 2022-07-16
Add differenceWith

vlad-yakovlev:
Add the same function as presented in [Lodash](https://lodash.com/docs/4.17.15#differenceWith) and [Ramda](https://ramdajs.com/docs/#differenceWith)

Diff so big because of changes in `mapping.md`. Code and tests mostly the same as in the `difference` method


src/differenceWith.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Excludes the values from `other` array.
+ * @param array the source array
+ * @param other the values to exclude
+ * @param isEquals the comparator
+ * @signature
+ *    R.differenceWith(array, other, isEquals)
+ * @example
+ *    R.differenceWith([1, 2, 3, 4], [2, 5, 3], R.equals) // => [1, 4]
```
TkDodo, line 14: same here, lets please use better examples & tests

vlad-yakovlev: Fixed

TkDodo (approved):

TkDodo:
ah, we missed adding the export to `index.ts` 🤦 . Can you add that please?

vlad-yakovlev:
Yep, 2 min and I'll open new PR :)

PR #181, opened 2022-07-16
Add uniqWith

vlad-yakovlev:
Add the same function as presented in [Lodash](https://lodash.com/docs/4.17.15#uniqWith) and [Ramda](https://ramdajs.com/docs/#uniqWith)

Code and tests based on `uniq` method


src/uniqWith.ts
```diff
@@ -0,0 +1,63 @@
...
+type IsEquals<T> = (a: T, b: T) => boolean;
+
+/**
+ * Returns a new array containing only one copy of each element in the original list.
+ * @param array
+ * @param isEquals the comparator
+ * @signature
+ *    R.uniqWith(array, isEquals)
+ * @example
+ *    R.uniqWith([1, 2, 2, 5, 1, 6, 7], R.equals) // => [1, 2, 5, 6, 7]
```
TkDodo, line 14: This example is not really useful, because you could just use `R.unique` as well. How about trying to find unique objects that wouldn't be found with `R.unique` because it does a triple equals comparison?

also, the tests should likely also test something like that.

vlad-yakovlev: Thank you for suggestion. I'll fix that here and in #180

vlad-yakovlev: Fixed

vlad-yakovlev:
I've resolved conflicts

TkDodo:
thx. also here, `index.ts` re-export is missing :)

vlad-yakovlev:
Fixed :)

PR #187, opened 2022-07-16
chore: Release automation

vlad-yakovlev:
**Before merge we need to add GH_TOKEN and NPM_TOKEN to Secrets**

**v0.1.0 will be automatically published after merge**

Added release automation (#185)

I've `dry-run`-ed parts of this action for test:

<details>
  <summary>semantic-release dry-run</summary>
  
```
[3:28:05 AM] [semantic-release] › ℹ  Running semantic-release version 19.0.3
[3:28:05 AM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/npm"
[3:28:05 AM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/github"
[3:28:05 AM] [semantic-release] › ✔  Loaded plugin "analyzeCommits" from "@semantic-release/commit-analyzer"
[3:28:05 AM] [semantic-release] › ✔  Loaded plugin "generateNotes" from "@semantic-release/release-notes-generator"
[3:28:05 AM] [semantic-release] › ✔  Loaded plugin "prepare" from "@semantic-release/npm"
[3:28:05 AM] [semantic-release] › ✔  Loaded plugin "publish" from "@semantic-release/npm"
[3:28:05 AM] [semantic-release] › ✔  Loaded plugin "publish" from "@semantic-release/github"
[3:28:05 AM] [semantic-release] › ✔  Loaded plugin "addChannel" from "@semantic-release/npm"
[3:28:05 AM] [semantic-release] › ✔  Loaded plugin "addChannel" from "@semantic-release/github"
[3:28:05 AM] [semantic-release] › ✔  Loaded plugin "success" from "@semantic-release/github"
[3:28:05 AM] [semantic-release] › ✔  Loaded plugin "fail" from "@semantic-release/github"
[3:28:12 AM] [semantic-release] › ⚠  Run automated release from branch master on repository git@github.com:vlad-yakovlev/remeda.git in dry-run mode
[3:28:14 AM] [semantic-release] › ✔  Allowed to push to the Git repository
[3:28:14 AM] [semantic-release] › ℹ  Start step "verifyConditions" of plugin "@semantic-release/npm"
[3:28:14 AM] [semantic-release] [@semantic-release/npm] › ℹ  Verify authentication for registry https://registry.npmjs.org/
[...]

PR #188, opened 2022-07-16
Upgrade node for travis

vlad-yakovlev:
We need to upgrade node to 16 in ci to properly install semantic-release in #187. I can't find a way to do that on node v10

@TkDodo I can't be sure that this PR wouldn't break travis check completely. But if you have admin right in this repo we would be able to revert changes.

Make sure that you:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "10"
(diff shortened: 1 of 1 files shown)


vlad-yakovlev:
Oh, just found that installation is not required if ran through npx

PR #192, opened 2022-07-17
chore: Unify GH workflows

vlad-yakovlev:
I think it will be better to run check also on main branch before publishing. So I've renamed PR Check to CI. Publish job will be added to this workflow

---

Make sure that you:


package.json
```diff
@@ -18,6 +18,7 @@
   },
   "scripts": {
     "test": "jest --coverage && yarn run compile",
+    "test:ci": "jest",
```
TkDodo, line 21: should we keep `--coverage` here so that the codecov bot we want to add can pick up the coverage?

vlad-yakovlev: I'll add --cooverage for codecov separately. Think we souldn't calculate in on every test run. Configuration example: https://github.com/vlad-yakovlev/raxee-bot/tree/main/.github/workflows

PR #195, opened 2022-07-18
feat: add strict version to keys where the return type is a union of the input

TkDodo:
---

Make sure that you:

### DIFF
--- a/docs/src/FnDoc.tsx
+++ b/docs/src/FnDoc.tsx
@@ -12,6 +12,7 @@ export interface MethodDoc {
   returns: JsTagProps;
   indexed: boolean;
   pipeable: boolean;
+  strict: boolean;
 }
 
 export interface FnDocProps {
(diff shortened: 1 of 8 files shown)


docs/src/FnDoc.tsx
```diff
@@ -38,6 +39,11 @@ export function FnDoc(props: FnDocProps) {
               <div className=" badge pipeable-color">pipeable</div>
             </small>
           )}
```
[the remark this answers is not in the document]
TkDodo: done

src/keys.test.ts
```diff
@@ -8,4 +9,13 @@ describe('Test for keys', () => {
   it('should return keys of object', () => {
     expect(keys({ a: 'x', b: 'y', c: 'z' })).toEqual(['a', 'b', 'c']);
   });
+
+  describe('strict', () => {
+    const actual = keys.strict({ a: 'x', b: 'y', c: 'z' } as const);
+    expect(actual).toEqual(['a', 'b', 'c']);
+
+    const result: AssertEqual<typeof actual, Array<'a' | 'b' | 'c'>> = true;
```
vlad-yakovlev, line 17: Have never seen this type comparison before. TS would broke here on Jest compile?

TkDodo: it's a type level "test". the jest assertion itself doesn't really do anything (compares true with true), but the typescript compiler will complain if the two types are not the same.

vlad-yakovlev: Look like possible replacement but we'll be required configure eslint for that. Think we can got back to this later, I'll create issue

PR #196, opened 2022-07-18
chore: Add pages workflow

vlad-yakovlev:
Add GH pages build and publishing. Would work only on merge to main branch so we need to use something like Netify for PRs. Example: https://remeda.vlad-yakovlev.dev/

Could be safely merged because current website uses travis & netify

---

Make sure that you:


vlad-yakovlev:
GitHub configuration:
image

Domain configucation:
image

Issue #147, opened 2021-09-03
Feature Request: make map, mapValues, filter, ... like functions become null-safe like lodash?


ericfong:
I really like "data-first", "data-last" and "fixed number of arguments" approach.

It would be better and easier to use if map, mapValues, filter, ... like functions become null-safe.

So, we can

```js
R.map(arrOrNull, ...)
```

instead of 

```js
R.map(arrOrNull || [], ...)
```

TkDodo:
I think that's an okay change that would also be backwards compatible, provided that type inference will still work in data-last versions and pipes. If you want, you can play around with it and create a PR, maybe for one of the functions as proof of concept and we can take it from there?

Bessonov:
I'm not sure that I like the idea. I have following reasons:
- The semantic isn't clear. Is `null` converted to and handled as an empty array?
- What's about `undefined`? Same handling as for `null`?
- It contradicts with "normal" array semantic. `null.map` is invalid and `null?.map` doesn't execute map.
- How this is handled in other libraries?

Overall, I think it will introduce silent failing errors, which are hard to debug.

TkDodo:
All valid points. I know lodash works with falsy input. There are also some methods where the semantics are not so obvious. Like, what should `R.clone(null)` do - fallback to empty object and then clone that?

Bessonov:
> I know lodash works with falsy input.

... and mutates passed objects. Two of the reasons why lodash shouldn't be used :)

> Like, what should R.clone(null) do - fallback to empty object and then clone that?

In this concrete example, if null is allowed at all, I would expect to get `null`. For me it's the same as `{prop: null}`. Or do I miss something :mag: ?

TkDodo:
> Two of the reasons why lodash shouldn't be used :)

👍 

> Or do I miss something 🔍 ?

It was just my attempt at signaling why accepting falsy input for all functions might be intransparent in some cases :)

ericfong:
Hi,

BTW, I think lodash treat `null` and `undefined` as empty array in map/filter.

I think one of the Remeda Design Goals is 'The usage must be programmer friendly, and that's more important than following XYZ paradigm strictly.' 
And my personal reason for using lodash or remeda like util instead of Array.map() just because they wrap native js func up and handle this and that (auto null handling) for me. 

I would like to know 'accepting falsy input' will it break "Manual annotation should never be required, and proper typings should infer everything"?

[...]

Issue #4, opened 2018-08-06
pick doesn't support undefined


rifler:
In lodash I can do this
```
pick(undefined, ['key']);
// {}
```

What do you think about this kind of feature?

BetterCallSky:
Could you provide an example?

rifler:
yes, sorry, update desc

BetterCallSky:
Agree, it should be supported. I will fix it soon.

PR #204, opened 2022-07-28
feat: Add omitBy

### LINKED ISSUE #202: Support omitBy
This library already supports `omit`, please add equivalent of [lodash/omitBy](https://lodash.com/docs/4.17.15#omitBy)

vlad-yakovlev:
Closes #202. Made from `pickBy`

---

Make sure that you:

### DIFF
--- a/src/omitBy.test.ts
+++ b/src/omitBy.test.ts
@@ -1,0 +1,26 @@
+import { omitBy } from './omitBy';
+import { pipe } from './pipe';
+
+describe('data first', () => {
+  test('it should pick props', () => {
+    const result = omitBy(
+      { a: 1, b: 2, A: 3, B: 4 },
+      (val, key) => key.toUpperCase() === key
+    );
+    expect(result).toStrictEqual({ a: 1, b: 2 });
+  });
+  test('allow undefined or null', () => {
+    expect(omitBy(undefined as any, (val, key) => key === 'foo')).toEqual({});
+    expect(omitBy(null as any, (val, key) => key === 'foo')).toEqual({});
+  });
+});
+
+describe('data last', () => {
+  test('it should pick props', () => {
+    const result = pipe(
+      { a: 1, b: 2, A: 3, B: 4 },
+      omitBy((val, key) => key.toUpperCase() === key)
+    );
+    expect(result).toEqual({ a: 1, b: 2 });
+  });
+});
--- a/src/omitBy.test.ts
+++ b/src/omitBy.test.ts
@@ -2,21 +2,17 @@ import { omitBy } from './omitBy';
 import { pipe } from './pipe';
 
 describe('data first', () => {
-  test('it should pick props', () => {
+  test('it should omit props', () => {
     const result = omitBy(
       { a: 1, b: 2, A: 3, B: 4 },
       (val, key) => key.toUpperCase() === key
     );
     expect(result).toStrictEqual({ a: 1, b: 2 });
   });
-  test('allow undefined or null', () => {
-    expect(omitBy(undefined as any, (val, key) => key === 'foo')).toEqual({});
-    expect(omitBy(null as any, (val, key) => key === 'foo')).toEqual({});
-  });
 });
 
 describe('data last', () => {
-  test('it should pick props', () => {
+  test('it should omit props', () => {
     const result = pipe(
       { a: 1, b: 2, A: 3, B: 4 },
       omitBy((val, key) => key.toUpperCase() === key)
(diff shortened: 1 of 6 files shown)


src/omitBy.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    const result = omitBy(
+      { a: 1, b: 2, A: 3, B: 4 },
+      (val, key) => key.toUpperCase() === key
+    );
+    expect(result).toStrictEqual({ a: 1, b: 2 });
+  });
+  test('allow undefined or null', () => {
+    expect(omitBy(undefined as any, (val, key) => key === 'foo')).toEqual({});
+    expect(omitBy(null as any, (val, key) => key === 'foo')).toEqual({});
+  });
```
sushantdhiman, line 15:
```suggestion
  });
  test('allow partial type', () => {
    expect(omitBy({} as Partial<{ a: string, b: number }>, (val, key) => key === 'a')).toEqual({});
  });
```

vlad-yakovlev: That wouldn't work. OmitBy like a PickBy works only with Record

src/omitBy.ts
```diff
@@ -0,0 +1,46 @@
...
+): (object: Record<K, V>) => Record<K, V>;
+
+export function omitBy() {
+  return purry(_omitBy, arguments);
+}
+
+function _omitBy(object: any, fn: (value: any, key: any) => boolean) {
+  if (object == null) {
+    return {};
+  }
```
TkDodo, line 39: I think it's pretty weird that `omitBy` and `pickBy` both accept `undefined` or `null` as input, when the types clearly state that you cannot pass that. It shows in the tests because we need `as any`.

Other methods like `map` or `filter` do _not_ allow passing nullish values, and I think this is on purpose, given the discussion here:
- https://github.com/remeda/remeda/issues/147

It seems that `pick` also does a null check, but `omit` does not 🤔 

---

so my thoughts here would be to:
- align `omitBy` with `omit` rather than with `pickBy` and remove this null check
- it was added to `pick` because of:
  - https://github.com/remeda/remeda/issues/4

but given that it is the _only_ function that does that, we should consider removing this check in `pick` and `pickBy` as well, either in the next major version (too bad we see this right after v1, lol), or, just in a minor because one goal is to align the implementation with the types... what do you think?

vlad-yakovlev: Hard question :)

I have some options:
- We can add nullckecks to every function, think this feature could be useful. But we should update types.
- We can remove all nullcheck, make v2, coz this is breaking change. I think that can help us to slightly improve performance (so slightly that we won’t mention) but make lib harder to use.
- Remove this check from omitBy for now, and discuss this problem later :)

vlad-yakovlev: First option not ultimate solution too. We should discuss some points. For example, which type should be returned - null/undefined (as on input) or empty array/object.

vlad-yakovlev: Second option improves transparency: if input was incorrect we'll throw error instead of doing other's job

vlad-yakovlev: Third option solves nothing but rising tech debt

src/omitBy.test.ts
sushantdhiman: I think both OmitBy/PickBy should also work with arbitrary object kind, if possible. It is sometimes impossible to control input argument.

One hackish way to use a typed object with rameda will looks like this :--

```ts
const data = {} as Partial<{ a: string }>;
const result = R.omitBy(
  (data as unknown) as Record<keyof typeof data, unknown>,
  (v, k) => k === "a"
);
```

src/omitBy.ts
vlad-yakovlev: Removed this check here for now

TkDodo: yeah let's just keep this in mind for the future: no more unnecessary null checks, and maybe we can remove the existing ones some day :)

src/omitBy.test.ts
vlad-yakovlev: I understood your problem. Try to implement workaround & cover with tests. I'll let you know when finish

vlad-yakovlev: @sushantdhiman Done :)

sushantdhiman: @vlad-yakovlev Thanks 👍🏻

vlad-yakovlev:
@TkDodo think everything done here, could you make codereview please? :)

TkDodo (approved):

PR #205, opened 2022-07-29
feat: Add tap function

vlad-yakovlev:

See #149, #114
---

Make sure that you:

### DIFF
--- a/mapping.md
+++ b/mapping.md
@@ -61,6 +61,7 @@ documentation when migrating._
 | `splitAt`        | `-`              | `splitAt`           |
 | `splitWhen`      | `-`              | `splitWhen`         |
 | `take`           | `take`           | `take`              |
+| `tap`            | `tap`            | `tap`               |
 | `takeWhile`      | `takeWhile`      | `takeWhile`         |
 | `times`          | `times`          | `times`             |
 | `toPairs`        | `toPairs`        | `toPairs`           |
--- a/src/index.ts
+++ b/src/index.ts
@@ -62,6 +62,7 @@ export * from './splitAt';
 export * from './splitWhen';
 export * from './take';
 export * from './takeWhile';
+export * from './tap';
 export * from './times';
 export * from './toPairs';
 export * from './type';
--- a/src/tap.test.ts
+++ b/src/tap.test.ts
@@ -1,0 +1,18 @@
+import { tap } from './tap';
+import { pipe } from './pipe';
+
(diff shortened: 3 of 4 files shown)


TkDodo:
any chance we can add the `lazy` version as described here?

- https://github.com/remeda/remeda/pull/114#issuecomment-754524515

vlad-yakovlev:
@TkDodo not changed yet. See if there any way to implement this on current version of pipe

As I read discussion, @lstkz suggested [here](https://github.com/remeda/remeda/pull/114#issuecomment-755370215) to run `fn` callback on every lazy call. That already could be done with `R.map(R.tap(fn))`, no changes required. It that way is ok, let's left as-is.

For me more interesting to make tap that will run tap in pipe with array but only with those element that will go through the pipe. That's what I'm trying to implement now

vlad-yakovlev:
@TkDodo I totally forgot how should `R.pipe(['a', 'b', 'c'], R.tap(console.log))` print result?
1. 3 calls with `'a', 'b', 'c'`
2. 1 call with `'a,b,c'` or something like that?

paour:
I hope you can continue with this PR, `tap` is a very useful quick debug tool.

cjquines:
hey, i do want to see this merged, i think the current implementation works fine as-is?

`R.pipe(['a', 'b', 'c'], R.tap(console.log))` has `console.log` called once

`R.pipe(['a', 'b', 'c'], R.map(R.tap(console.log)))` has `console.log` called thrice

in particular `R.tap` should not be lazy, as it does not take an array as input, it takes in any `T` and returns `T`

TkDodo:
> in particular i dont think R.tap should be lazy

why not? As shown in this example, we'd only expect 3 logs, not 5

```
R.pipe(
  [1, 2, 3, 4, 5],
  R.tap(console.log),
  R.take(3)
);
```

I don't mind if we don't have that in the first version, but since it works for other functions, it should work for tap, too, don't you think?

cjquines:
i did say that i think this example would have one log:

```
R.pipe(
  [1, 2, 3, 4, 5],
  R.tap(console.log),
  R.take(3)
);
```
should log `[1, 2, 3, 4, 5]`, while

```
R.pipe(
  [1, 2, 3, 4, 5],
  R.map(R.tap(console.log)),
  R.take(3)
);
```

should log `1`, then `2`, then `3`. it follows from `map` being lazy, not `tap`.

PR #218, opened 2022-10-18
Add sumBy

thelawclaw11:
Added a function called sumBy. My team uses this package and came across a need for this function recently. I used it a lot in Lodash and it seems like a reasonable inclusion.


TkDodo (approved):
thank you ❤️

PR #219, opened 2022-10-18
feat: Add randomInteger

hsligting:
Added a function that generates a random integer between an upper and lower bound called randomInteger.

### DIFF
--- a/mapping.md
+++ b/mapping.md
@@ -53,6 +53,7 @@ documentation when migrating._
 | `pipe`           | `flow`           | `pipe`              |
 | `prop`           | `get`            | `prop`              |
 | `purry`          | `-`              | `-`                 |
+| `randomInteger`  | `-`              | `-`                 |
 | `randomString`   | `-`              | `-`                 |
 | `range`          | `range`          | `range`             |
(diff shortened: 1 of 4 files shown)


TkDodo:
I'm not sure if we should add this to remeda. I just saw that we have `randomString` as well, but I'm not a fan of using `Math.random` as it's not a "real" random thing. Not sure what I would use this for over, say, `nanoid`.

@vlad-yakovlev thoughts?

thelawclaw11:
If random string is in Remeda I kinda think this should be in as well but that's just my 2 cents. I feel like it would be nice to have pseudo random ints for stuff that doesn't need to be truly random.

vlad-yakovlev:
@TkDodo hard question :) Yeah, we already have `randomString`, think we can safely add `randomInteger` and think about these functions in future major versions

About implementation: I think we should not include `upper`. `range` does not include `end`. Also, suggest rename `lower` to `begin` and `upper` to `end`

eranhirsch:
It looks like this PR has been stale for very long. I'm closing it for housekeeping, but @hsligting if you still want to push forward with this i'd be happy to help so just re-open it.

PR #220, opened 2022-10-31
add meanBy

thelawclaw11:
I added a function called meanBy which behaves the same as the Lodash version. A teammate and I needed this function this week and we found out Remeda doesn't have it so I wrote it up. I was unsure what the behavior should be if the input array is empty so I just had the function return NaN which is the same as Lodash and Ramda. Personally I might have simply made the function return 0 instead of NaN but for someone used to Lodash or Ramda that behavior would be unexpected.


mapping.md
```diff
@@ -39,6 +39,7 @@ documentation when migrating._
 | `last`           | `last`           | `last`              |
 | `map`            | `map`            | `map`               |
 | `mapKeys`        | `mapKeys`        | `-`                 |
+| `meanBy`         | `meanBy`         | `-`                 |
```
TkDodo, line 42: in your PR description, you said that both `lodash` and `ramda` support `meanBy`, but you haven't added `ramda` to the mapping table 🤔

src/meanBy.test.ts
```diff
@@ -0,0 +1,33 @@
+import { meanBy } from './meanBy';
+import { pipe } from './pipe';
+
+const array = [{ a: 1 }, { a: 2 }, { a: 4 }, { a: 5 }, { a: 3 }] as const;
+
+describe('data first', () => {
```
TkDodo, line 6: can you add a test for the mentioned empty array case please?

TkDodo (commented):
please run your changes through prettier

thelawclaw11:
I ran prettier on my changes, added tests for empty arrays. As for a Ramda mapping, I misspoke earlier about Ramda supporting meanBy. They support mean but not meanBy.

PR #222, opened 2022-11-06
feat: add setPath helper

### LINKED ISSUE #78: Feature Request: setPath
It would be awesome to have an option to set values deeply nested with a path value. Ideally, something that supports lodash's format would be ideal, although I realise that might be complex to implement.

```javascript
var object = { 'a': [{ 'b': { 'c': 3 } }] };
 
var object2 = R.setPath(object, 'a[0].b.c', 4);

console.log(object2.a[0].b.c);
// => 4
```

If it's easier, or to keep in line with `R.pathOr` it might make sense to use an array for path instead.

``` javascript
var object3 = R.set(object, ['x', '0', 'y', 'z'], 5);

console.log(object3.x[0].y.z);
// => 5
```

airjp73:
Adds a new `setPath` helper that supports both array and string paths. (Resolves #78)

Examples:
```
setPath(obj, 'a.b.c[0].d', 'someValue');
setPath(obj, ['a', 'b', 'c', 0, 'd'], 'someValue');
```

Notes:

The types themselves are working correctly, but I'm not totally happy with where the auto-complete is for this. I wouldn't be surprised if there's some more advance TS magic I could do to make both work great.

---

Make sure that you:


package.json
```diff
@@ -10,9 +10,10 @@
     "@types/jest": "^26.0.14",
     "jest": "^26.4.2",
     "marked": "^0.4.0",
-    "prettier": "^2.1.2",
+    "prettier": "^2.7.1",
```
airjp73, line 13: Prettier wasn't formatting the `_paths.ts` file correctly until I updated. I suspect it was the `infer Tail extends Key[]` syntax.

package.json
```diff
@@ -10,9 +10,10 @@
...
     "ts-jest": "^26.4.1",
     "ts-node": "^8.4.1",
+    "ts-toolbelt": "^9.6.0",
```
airjp73, line 16: Could also just copy over the `Narrow` type if you'd rather not add a dep.

tsconfig.json
```diff
@@ -42,7 +42,7 @@
     "noFallthroughCasesInSwitch": true /* Report errors for fallthrough cases in switch statement. */,
 
     /* Module Resolution Options */
-    // "moduleResolution": "node",            /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
+    "moduleResolution": "node",            /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
```
airjp73, line 45: This was necessary to get Typescript to import `ts-toolbelt`. Could undo this if we didn't pull in that dep.

src/_paths.ts
```diff
@@ -0,0 +1,78 @@
+import { Key } from './_types';
```
airjp73, line 1: I put all this in a separate file because I imagine it could be used to update `pathOr` to support an arbitrary path length (instead of just 3 path segments) and path strings.

package.json
TkDodo: yeah let's copy the type please. It would need to be a runtime dependency otherwise, right?

src/setPath.test.ts
```diff
@@ -0,0 +1,355 @@
...
+        a: {
+          ...obj.a,
+          b: { c: 2 },
+        },
+      });
+    });
+
+    test('should correctly type partial paths', () => {
+      // @ts-expect-error - this path should yield a type of { c: number }
+      expect<SampleType>(setPath(obj, 'a.b', 123)).toEqual({
```
TkDodo, line 176: I'm not sure if we should support the dot notation. I don't see much advantage of it. What if we have an object that has a dot in the name, like?

```
const obj = {
 a: {
  b: 5
 },
 'a.b': 'five'
}
```

package.json
TkDodo: i just merged a PR that enforces prettier checks in the pipeline.
if we want to update prettier, I think it would be better to create a separate PR that just updates prettier and runs prettier formatting on the whole codebase. It would avoid cluttering this PR with unrelated formatting changes. Could you do that?

src/setPath.test.ts
TkDodo: I'm also hoping that the types might get less complex if we remove this, because honestly, I feel a bit intimidated by them 😅

package.json
airjp73: I could do that. But I'm also wondering if I should be using this syntax at all. It would mean consumers of `rememda` would need to be on TS 4.7 or greater which would be a breaking change.

package.json
airjp73: I can do that. It's not really a _runtime_ dependency in the sense that no code from `ts-toolbelt` will be downloaded in the browser (its only type helpers, no actual code). But the library's type definitions will include the TS toolbelt type.

src/setPath.test.ts
airjp73: For me, the string notation is the primary reason I need a helper like this (migrating [this](https://www.remix-validated-form.io/arrays-and-nested) away from lodash is my use-case).

A lot of the type complexity also comes from supporting an arbitrary path length rather than only handling 3 like `pathOr` does. If we remove the string notation, it will simplify the types a little, but it will still be more complex than the types for `pathOr`.

What if we split this into two helpers, `setPath` and `stringToPath`? Then the string path use-case could be served this way:

```ts
setPath(obj, stringToPath('a.b.c'))
```

and then I can split up the types a bit more & add more comments to hopefully make it easier to grok.

package.json
airjp73: I might spend some time testing this -- it might be possible to document "this helper requires TS 4.7+" rather than requiring the whole lib to do that.

src/setPath.test.ts
airjp73: Also re: the object with `'a.b'` as a key:

My first instinct is to not handle that case, but I looked around at some of the prior art for this. The solution that makes the most sense to me is adding support for strings in the path like this: `a['b.c']`.

Makes things quite a bit more complex though. Almost makes me wonder if this kind of thing is worth having a dedicated library instead of adding it to a helper collection.

package.json
TkDodo: yeah, what we could do is could ship differential types. We used to do that for [react-query v3](https://github.com/TanStack/query/blob/5e37b145a8836b00dd20860e8eb1f94c608fec81/package.json#L83-L85) when we introduced something that used variadic tuples but we also wanted to support versions < 4.1

Question is if we could come up with a type for this helper that would work well enough with TS < 4.7, or maybe, it could just accept `Array<string>` and we document that it only works "well" with TS 4.7 and above.

package.json
TkDodo: I think there is one or two real runtime functions in ts-toolbelt, which is why it has [129 bytes](https://bundlephobia.com/package/ts-toolbelt@9.6.0). It can probably be tree-shaken away if we don't use it, but having "zero dependencies" is also something that sells nicely 🤷

src/setPath.test.ts
TkDodo: > What if we split this into two helpers, setPath and stringToPath? Then the string path use-case could be served this way:

I personally like that suggestion a lot. We could also document the limitation that objects where the key itself contains a dot will be tricky to work with `stringToPath` and decide to not support that. People can always decide to _not_ use that helper and just use `setPath` with the array of path notation.

I think this means this helper would automatically be usable for `pathOr`:

```
pathOr(obj, stringToPath('a.b.c')
```

and we could simply / streamline the types of `pathOr` as well.

Gonna pull in @vlad-yakovlev to see what he thinks about the suggestion :)

package.json
airjp73: I'm pretty sure `infer extends` is just a shorthand so we could ship the same type essentially.

```ts
Foo extends Array<infer U extends string>
  ? 'foo'
  : never

// Is equivalent to

Foo extends Array<infer U>
  ? U extends string
    ? 'foo'
    : never
  : never
```

The types get less readable but with the same functionality.

TkDodo: that would also be fine 👍

src/setPath.test.ts
vlad-yakovlev: When I think about usability, I would like to split this functions in two. Like this: `setPath(obj, stringToPath('a.b.c'))`. But types looks impossible to me in this case.

So, if we need this feature and one day we need to use `a.b` key, we can use array argument for that

vlad-yakovlev: Look like we need to pass `obj` to `stringToPath` to determine which keys we can use. If it's possible, then we definitely need to split this function in two

TkDodo: Happy new year 🎉
no worries - I've been absent a bit as well :)

src/stringToPath.test.ts
```diff
@@ -0,0 +1,11 @@
+import { stringToPath } from './stringToPath';
+
+test('should convert a string to a deeply nested path', () => {
+  const res = stringToPath('a.b[0].c');
+  expect<['a', 'b', '0', 'c']>(res).toEqual(['a', 'b', '0', 'c']);
+});
+
+test('should handle nested dot paths', () => {
+  const res = stringToPath('a.b[a.b].c');
```
airjp73, line 9: Supporting this kind of path actually didn't require any code changes, so that's neat.

airjp73 (commented):
Updated to separate this into two helpers `setPath` and `stringToPath`. I was also able to delete some types as a result of this. The autocomplete isn't great for `stringToPath`, but I guess the primary use-case for this helper doesn't need perfect autocomplete -- that's what the array syntax is for.

airjp73:
Sorry if "re-request review" spammed you notifications. The UI was acting weird so I clicked the button a couple times

TkDodo:
can you please look at the failing lint tasks?

I'm also still pondering if the string helper is something that is really valuable 🤔 . @eranhirsch what do you think?

eranhirsch:
I couldn't follow the logic here unfortunately, reading it through github with all the lint warnings just made it too confusing. I'd also say that I'm a little concerned with using code copied from ts-toolbelt which has been dead for 2 years now, as typescript has evolved in those 2 years and I don't trust the types to be as accurate as they should be.

@airjp73 - can you rebase, fix lint and typescript issues, and see if type-fest have similar types to what you need? I don't mind importing type-fest directly into remeda to use specific types so that we can keep the project more accurate with just the specializations we need from types.

TkDodo (approved):

TkDodo:
sorry for letting this lie for so long. We've updated some lint rules and TS in the meantime, so please take a look at the CI failures. Thank you 🙏

vlad-yakovlev:
Hey 👋 I will review the PR by the end of tomorrow

vlad-yakovlev (approved):
Hey 👋 I'm sorry for delay. Looks great!

PR #226, opened 2022-11-27
fix: types for keys.strict

TkDodo:
at runtime, keys will always be strings, so it is better to convert numbers to strings on type level

---

Make sure that you:


TkDodo:
the real changes are in the first commit (510b0c2). The second commit updates prettier because I was getting a prettier error with the template literal syntax I'm using.

vlad-yakovlev (approved):
🚀

PR #227, opened 2022-12-02
feat: allow different data types on differenceWith

santialbo:
This PR tweaks types on `differenceWith` to allow different data types on the two arrays.
This allows things like the following:
```ts
  differenceWith([1, 2, 3, 4], ['2', '3'], (a, b) => a.toString() === b)
```


TkDodo (commented):
Thanks, this is a good improvement.
One thing I'd like to change is the naming of the generics `T` and `K`. I think `T` is fine as long as we only have on generic, but once we have two, it should be more descriptive. Moreover, `K` usually stands for `Key`, but is used differently here.

I would suggest to change it towards:

```
export function differenceWith<TFirst, TSecond>(
  array: readonly TFirst[],
  other: readonly TSecond[],
  isEquals: IsEquals<TFirst, TSecond>
): TFirst[];
```

and follow this throughout the file.

santialbo:
I absolutely agree with you. Normally I try to change as little as possible when doing PRs to other libraries, but this is something I was thinking about when making these changes. I updated my branch.

vlad-yakovlev (approved):
👍

TkDodo (approved):

PR #231, opened 2022-12-07
docs(readme): Make data-last support more obvious to new readers

damassi:
Just discovered your library today and really nice work! 

Wanted to update the docs very slightly to make it clearer to new readers that Remeda is in fact a data-last library like one would expect. When I first landed on the README i scanned it looking for code, and when I saw that the Remeda / Ramda examples didn't match I immediately assumed that this wasn't actually data-last support and figured that lodash was sufficient. And then I read it again! And dove into the docs and saw that it does everything I'd like. This update makes that clearer for the new arrival, and particularly the type of new arrival who doesn't read words too closely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ R.pick(['firstName', 'lastName'], obj);
 _.pick(obj, ['firstName', 'lastName']);
 ```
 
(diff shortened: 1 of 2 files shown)


pvinis (approved):
yep. nice

PR #236, opened 2022-12-19
Add invert function

LandonSchropp:
This adds the handy `invert` method, which swaps the keys and values in an object. This method is present in both Lodash and Ramda, and I needed it for my current project, so I thought I'd add it here as well.

---

Make sure that you:

### DIFF
--- a/src/invert.ts
+++ b/src/invert.ts
@@ -31,5 +31,5 @@ function _invert(object: Invertable): Inverted {
   return Object.entries(object).reduce((accumulator, [key, value]) => {
     accumulator[value] = key;
     return accumulator;
-  }, {} as Inverted)
+  }, {} as Inverted);
 }
--- a/src/invert.ts
+++ b/src/invert.ts
@@ -1,0 +1,35 @@
+import { purry } from './purry';
+
+type Invertable = { [index: string]: string } | { [index: number]: string };
+type Inverted = { [index: string]: string };
+
+/**
+ * Returns an object whose keys are values are swapped. If the object contains duplicate values,
+ * subsequent values will overwrite previous values.
+ * @param object the object
+ * @signature
+ *    R.invert(object)
+ * @example
+ *    R.invert({ a: "d", b: "e", c: "f" }) // => { d: "a", e: "b", f: "c" }
+ *    R.invert({}) // => {}
+ *    R.pipe(
+ *      { a: "d", b: "e", c: "f" }
+ *      R.invert()
+ *    ); // => { d: "a", e: "b", f: "c" }
+ *
+ * @category object
+ * @pipeable
+ */
+export function invert<T>(object: Invertable): Inverted;
+export function invert<T>(): (object: Invertable) => Inverted;
+
+export function invert() {
+  return purry(_invert, arguments);
+}
+
+function _invert(object: Invertable): Inverted {
+  return Object.entries(object).reduce((accumulator, [key, value]) => {
+    accumulator[value] = key;
+    return accumulator;
+  }, {} as Inverted)
+}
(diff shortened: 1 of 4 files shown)


TkDodo:
entries in `index` and `mapping.md` are missing please

LandonSchropp:
I forgot to force-push the commit after I added them. 🙂

texastoland:
I think I accidentally wrote a more exact type for this that works with `as const` in sindresorhus/type-fest#543.

src/invert.ts
```diff
@@ -0,0 +1,35 @@
...
+ *    R.pipe(
+ *      { a: "d", b: "e", c: "f" }
+ *      R.invert()
+ *    ); // => { d: "a", e: "b", f: "c" }
+ *
+ * @category object
+ * @pipeable
+ */
+export function invert<T>(object: Invertable): Inverted;
+export function invert<T>(): (object: Invertable) => Inverted;
```
TkDodo, line 24: please also add jsdocs for the `data-last` version and add the `@category`

src/invert.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @param object the object
+ * @signature
+ *    R.invert(object)
+ * @example
+ *    R.invert({ a: "d", b: "e", c: "f" }) // => { d: "a", e: "b", f: "c" }
+ *    R.invert({}) // => {}
+ *    R.pipe(
+ *      { a: "d", b: "e", c: "f" }
+ *      R.invert()
+ *    ); // => { d: "a", e: "b", f: "c" }
```
TkDodo, line 18: this is actually the `data-last` example, that would need to go on the other method please. Also, the code needs to work, and there is a missing comma here. This is why the build fails:

```
12:26:59 AM: SyntaxError: ',' expected. (5:6)
12:26:59 AM:   3 |    R.pipe(
12:26:59 AM:   4 |      { a: "d", b: "e", c: "f" }
12:26:59 AM: > 5 |      R.invert()
12:26:59 AM:     |      ^
12:26:59 AM:   6 |    ); // => { d: "a", e: "b", f: "c" }
```

src/invert.ts
```diff
@@ -0,0 +1,35 @@
+import { purry } from './purry';
+
+type Invertable = { [index: string]: string } | { [index: number]: string };
+type Inverted = { [index: string]: string };
```
TkDodo, line 4: as @texastoland has pointed out, I think the types could be improved here to have a relation with each other. The generic `T` you've added to `invert` is actually unused ...

TkDodo (changes requested):
thanks for this PR, please have a look at my comments

TkDodo:
closing due to staleness

LandonSchropp:
@TkDodo Sorry for the staleness on this! I had a baby at the end of last year, and this completely slipped off my radar. I'll make the changes you suggested and push up a new branch today. 🙂

texastoland:
Happy to review too and congrats 👶🏼

LandonSchropp:
@TkDodo and @texastoland: I think I got the types refined a bit. I'll be honest, I wasn't quite following everything in the type that @texastoland posted. Here's what I wrote:

``` typescript
type Invertable =  Record<PropertyKey, PropertyKey> | PropertyKey[];
type Inverted<T extends Invertable> = Record<Extract<T[keyof T], PropertyKey>, T[keyof T]>

function _invert<T extends Invertable>(object: T): Inverted<T> {
  return Object.entries(object).reduce((accumulator, [key, value]) => {
    accumulator[value] = key;
    return accumulator;
  }, {} as Inverted<T>)
}
```

If that doesn't work, could you please walk me through where any issues might be? Thanks in advance!

PR #238, opened 2023-01-05
improve the type of curried omit, pick, and prop

yanick:
With the current type, the curried version of pick is... meh. `pick(['a'])` doesn't work, expect 'a' to be never, and the working alternatives, `pick(['a'] as any)` or `pick<any,any>(['a'])`, are verbose.


yanick:
The tests are failing because of my addition of `ts-expect`. I'll leave it now just so that the reviewers can see the tests, but I can remove them prior to merging.

TkDodo:
looks good - what about `omit` ? Doesn't it have a similar problem?

> The tests are failing because of my addition of ts-expect. I'll leave it now just so that the reviewers can see the tests, but I can remove them prior to merging.

I'm happy with moving to `ts-expect`, but maybe we should do this for all of our tests and remove the custom helpers that we have, provided that `ts-expect` can replace them?

yanick:
> looks good - what about `omit` ? Doesn't it have a similar problem?

Yup! But I wanted to do a trial balloon before patching ALLTHETHINGS! But I can sure extend the PR to `omit`.

> > The tests are failing because of my addition of ts-expect. I'll leave it now just so that the reviewers can see the tests, but I can remove them prior to merging.
> 
> I'm happy with moving to `ts-expect`, but maybe we should do this for all of our tests and remove the custom helpers that we have, provided that `ts-expect` can replace them?

I'll have a look at those helpers!

TkDodo:
fyi, we've updated prettier in an other PR. sorry for the conflicts 😅

yanick:
> fyi, we've updated prettier in an other PR. sorry for the conflicts sweat_smile

That's easily fixed. Done!

package.json
```diff
@@ -8,9 +8,11 @@
   "sideEffects": false,
   "devDependencies": {
     "@types/jest": "^26.0.14",
+    "assert": "^2.0.0",
     "jest": "^26.4.2",
     "marked": "^0.4.0",
     "prettier": "^2.8.0",
+    "ts-expect": "^1.3.0",
```
TkDodo, line 15: can you remove those changes from this PR please as the other PR also adds them?

also: why is `assert` needed? I don't see this used anywhere

tsconfig.json
```diff
@@ -42,7 +42,7 @@
     "noFallthroughCasesInSwitch": true /* Report errors for fallthrough cases in switch statement. */,
 
     /* Module Resolution Options */
-    // "moduleResolution": "node",            /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
+    "moduleResolution": "node",            /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
```
TkDodo, line 45: why is this change needed?

package.json
yanick: The `assert` is for `src/clone.test.ts`. 

As for removing the dependency on `ts-expect`, if I do the PR will fail its checks.

tsconfig.json
yanick: Without it I get:

```
src/omit.test.ts:1:28 - error TS2792: Cannot find module 'ts-expect'. Did you mean to set the 'moduleResolution' option to 'node', or to add aliases to the 'paths' option?

1 import { expectType } from 'ts-expect';
```

This being said, `expectType` is such a small function, we could define it ourselves and skip that dependency altogether, if we want.

TkDodo (approved):

eranhirsch:
The changes to prop broke my code, i'm getting weird typing from inference inside a pipe.

```ts
const x = R.pipe([{ a: 1 }, { a: 2 }, { a: 3 }], R.sortBy(R.prop("a")))
```

The type I get back is Record<any, any>... I don't even understand why...

PR #244, opened 2023-01-17
move to vitest's assertType

yanick:
See https://github.com/remeda/remeda/pull/238#issuecomment-1383131398

Basically, use the `expectType` function of `ts-expect` to check for types instead of the bespoke checking that was there. Technically it does the same thing, the syntax is just a wee bit shorter.

### DIFF
--- a/src/zip.test.ts
+++ b/src/zip.test.ts
@@ -36,10 +36,6 @@ describe('data first typings', () => {
     const actual = zip(first as [1, 2, 3], second as ['a', 'b', 'c']);
     assertType<Array<[1 | 2 | 3, 'a' | 'b' | 'c']>>(actual);
   });
-  test('tuples', () => {
-    const actual = zip(first as [1, 2, 3], second as ['a', 'b', 'c']);
-    assertType<Array<[1 | 2 | 3, 'a' | 'b' | 'c']>>(actual);
-  });
   test('variadic tuples', () => {
     const firstVariadic: [number, ...Array<string>] = [1, 'b', 'c'];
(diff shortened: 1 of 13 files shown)


yanick:
> @yanick vitest now has built-in typechecking functionality. https://vitest.dev/guide/testing-types.html

Oooooh... :heart_eyes: 

Patch incoming Real Soon:tm: !

yanick:
Moved to vitest's assertType

src/zip.test.ts
```diff
@@ -31,26 +30,21 @@ describe('data first', () => {
...
-    const result: AssertEqual<
-      typeof actual,
-      Array<[1 | 2 | 3, 'a' | 'b' | 'c']>
-    > = true;
-    expect(result).toBe(true);
+    assertType<Array<[1 | 2 | 3, 'a' | 'b' | 'c']>>(actual);
+  });
+  test('tuples', () => {
+    const actual = zip(first as [1, 2, 3], second as ['a', 'b', 'c']);
+    assertType<Array<[1 | 2 | 3, 'a' | 'b' | 'c']>>(actual);
```
eranhirsch, line 41: This looks accidental
```suggestion
```

eranhirsch (approved):
Looks great to me, can you also remove the actual type from the "_types" file?

src/zip.test.ts
yanick: Ooops. Merging collateral. Fixed!

yanick:
Aaaand removed _types.ts. Good call. :-)

TkDodo (approved):
❤️

PR #245, opened 2023-01-18
add a sample function

yanick:
Added a `sample` function that allows to pick a random sample of an array. Allows for repeating picks, and any sample size.

---

Make sure that you:

- [x ] Typedoc added for new methods and updated for changed
- [ x] Tests added for new methods and updated for changed
- [ x] New methods added to `src/index.ts`
- [ x] New methods added to `mapping.md`


eranhirsch:
I don't think that sample should return a shuffled response, I think doing it this way conflates 2 different concepts. I think it would be more useful and expected that the sample function returns a stable ordering based on the input ordering. If you also need to shuffle the results there is a specific utility to do just that.

At the end of the day a sampling method to get P elements on array of length N is just a filter with a probabilistic filtering method, e.g. 
```ts
filter(() => Math.random() < P/N)
```

but for small N, the result might contain ~P samples so we need a sample method that would make sure we return the correct number of values. This approach can also be built lazily, saving us actually building the output array if we don't need to.

As for duplicates version, I think that the name `sample` isn't correct for it, as it effectively over-samples values. Regardless, it too can have a stable output, randomizing the indices in the sample first, and then mapping over the values:
```ts
const overSampled = pipe(
  P,
  times(() => Math.floor(Math.random() * N)),
  sort((a, b) => a - b),
  map((index) => array[index]),
);
```

yanick:
For what it's worth: my undergrad studies were in Statistics, and my master was in Operation Research, specialization: Important Sampling. Can't mean I can't be wrong, but it might widen the confidence internal a wee bit. ;-)

 
> I don't think that sample should return a shuffled response, I think doing it this way conflates 2 different concepts. I think it would be more useful and expected that the sample function returns a stable ordering based on the input ordering. If you also need to shuffle the results there is a specific utility to do just that.

There are, of course, different ways to sample, but the cases for wanting an ordered sample are much, much less frequent than unordered samples. Like, for thinning down an already sorted array, sure, but for anything stochastic, I can't possibly imagine a case for it.

> At the end of the day a sampling method to get P elements on array of length N is just a filter with a probabilistic filtering method, e.g.
> 
> ```ts
> filter(array, () => Math.random() < P/array.length)
> ```
> 
> but for small N, the result might contain ~P samples 

Actually, the probability of getting exactly P entries using that filter is (if my probabilities aren't too rusty):

    N! / P! / (N-P)! * (P/N) ^ P * ( (N-P)/N)  ^ (N-P)

I.e., they follow a binomial distribution which, for a big enough N, can be approximated by a Gaussian distribution. Anyway, the short version of it is that the expectancy of the # of results will be P, but the probability of hitting exactly P will become itsy bitsy as N gets larger.

eranhirsch:
> For what it's worth: my undergrad studies were in Statistics, and my master was in Operation Research, specialization: Important Sampling.
:) worth a lot.

What I meant is that if a shuffled response is needed, users can always shuffle after the sampling in a pipe with the shuffle function, but if they don't, they'll need to reimplement the sample method themselves. I feel that the loss of flexibility is the point here, even if most users would then go ahead and shuffle (at least it would be a conscious choice they are making).

I think that randomizing P indices, instead of P elements, and then using a filter on the array's index is equivalent in terms of efficiency, but keeps the ordering, so there's no perf related concerns here. FWIW, this is how I defined a sample function in another project of mine, before I had remeda:

```ts
function sample<T>(arr: readonly T[], sampleSize: number): T[] {
  invariant(
    sampleSize >= 0,
    `Trying to extract a negative sample size ${sampleSize}!`
  );

  if (sampleSize === 0) {
    return [];
  }
  
  if (sampleSize >= arr.length) {
    // Trivial solution
    return arr;
  }

  // To optimize the selection we can toggle between an include and exclude
  // logic for the sample; when sampleSize is small we will pick a set of
  // random indices and use them to pick elements from the input array, and
  // when sampleSize is big we will pick which indices to skip when rebuilding
  // the array.

  // We use a set so that we can ignore duplicates
  const selectedIndices: Set<number> = new Set();
  while (selectedIndices.size < Math.min(sampleSize, arr.length - sampleSize)) {
    selectedIndices.add(Math.floor(Math.random() * arr.length));
  }

  return sampleSize <= arr.length - sampleSize
    ? map(sort([...selectedIndices], (a, b) => a - b), (index) => arr[index])
    : filter(arr, (_, index) => !selectedIndices.has(index));
}
```

src/sample.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+function _sample<T>(items: Array<T>, options: number | SampleType) {
+  let nbrSamples = typeof options === 'number' ? options : options.size;
+  const repeating = typeof options === 'object' && options?.repeating;
+
+  let size = items.length;
+  items = [...items];
+
+  if (nbrSamples > size && !repeating) nbrSamples = size;
```
eranhirsch, line 15: The second yellow flag is that we will go through sampling the whole array for P > N although the result is effectively a shuffle (in your impl) or a trivial identity (in my impl).

eranhirsch (commented):
The current implementation is always O(n), whereas a non-shuffling sampling impl can work at O(k).

yanick:
> What I meant is that if a shuffled response is needed, users can always shuffle after the sampling in a pipe with the shuffle function, but if they don't, they'll need to reimplement the sample method themselves. I feel that the loss of flexibility is the point here, even if most users would then go ahead and shuffle (at least it would be a deliberate choice they are making).

The thing is, I can't think of any common use-case for wanting a non-shuffled version. At best, you're going to curse most of your use-base to have to throw a pipe and a shuffle in the mix, at worst some people will not read the docs carefully and be bitten by the sorted return.

> I think that randomizing P indices, instead of P elements, and then using a filter on the array's index is equivalent in terms of efficiency, but keeps the ordering, so there's no perf related concerns here. FWIW, this is how I defined a sample function in another project of mine, before I had remeda:
> 
> ```ts
>   // We use a set so that we can ignore duplicates
>   const selectedIndices: Set<number> = new Set();
>   while (selectedIndices.size < Math.min(sampleSize, arr.length - sampleSize)) {
>     selectedIndices.add(Math.floor(Math.random() * arr.length));
>   }


That's a cromulent way of doing it, but there are a few trade-offs in there. The use of `Set` means more memory usage when P approaches N, and the loop trades the memory consumption for time -- which is stochastic instead of deterministic here.

src/sample.ts
```diff
@@ -0,0 +1,80 @@
...
+  size: number;
+  repeating: boolean;
+}
+
+function _sample<T>(items: Array<T>, options: number | SampleType) {
+  let nbrSamples = typeof options === 'number' ? options : options.size;
+  const repeating = typeof options === 'object' && options?.repeating;
+
+  let size = items.length;
+  items = [...items];
```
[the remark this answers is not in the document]
yanick: Yeah, you're right, for the repeating mode we don't change the original array, so that can be skipped.

src/sample.ts
yanick: Yes, that is correct. And as far as I know the Fisher-Yates' shuffle is still considered to be way to go for that: https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm

eranhirsch: That's actually not my concern, your impl is fisher-yates enough imho, the concern was that we now have 2 methods that effectively do the same thing (for a specific input). So shuffle === sample(HUGE_SAMPLE_SIZE), and I don't like having this duplication in case someone comes up with an optimization or whatever and doesn't port it to the other case.

src/sample.ts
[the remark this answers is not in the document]
eranhirsch: That's not the only inefficient case (although it's the only one I surfaced). If you are looking for a sample size of 1 and your array is huge, you'd also be copying the whole thing.

eranhirsch:
> The thing is, I can't think of any common use-case for wanting a non-shuffled version. 

The problem is I don't know anything about how our users would use this call, and what they expect out of it, tbh, i have no idea how similar functions are implemented in other frameworks and what people feel of those either.

I can only give an example of my own personal usage, which was to create a simple randomizer for board games where I had a predefined const array of faction names, and I wanted a random sampling of N of those factions. I didn't need those to be shuffled because the order wasn't important *semantically*, but because they have a specific ordering on the board I wanted the result to be *visually* in a specific order which was hard-coded and wasn't easy to sort by (without complicating the const to also contain an `order` prop. It might not be the most common use-case, but it's *a* use-case which shuffling breaks.

> At best, you're going to curse most of your use-base to have to throw a pipe and a shuffle in the mix, at worst some people will not read the docs carefully and be bitten by the sorted return.

This is were I'm stumped. I can't see how an "in-order" result would bite anyone, considering that it's a valid output of a shuffled response. It's one of K! possibilities for that specific sample, but there's no reason that their code, expecting a shuffled response, would break because it isn't shuffled. Because the sample is still random, if the fact that the output is also shuffled (in comparison to the input) is important, I think they are doing a service to their readers to have it explicitly there.

We can always add more flags to the options to control shuffling too but I don't like that this function already takes an option object at all. It's starting to shift from a utility function to a statistics framework; at the end of the day Remeda just isn't that. I think there's an implicit expectation that each function call in Remeda is extremely simple, to the point that anyone can more or less guess how it's implemented, and it's only there to save engineers time rewriting the same old logic in every project they need them in. 

> That's a cromulent way of doing it, but there are a few trade-offs in there. The use of Set means more memory usage when P approaches N, and the loop trades the memory consumption for time -- which is stochastic instead of deterministic here.

Because we use the optimization were we flip the logic past N/2, the Set memory footprint would be O(N/2) max. This means that the internal loop would only ever have to draw N/2 random indices at the worst case, and that shouldn't require much more than `N/2 + some small delta` iterations on average (where the small delta is the sum of geometric means(??) from 0...K-1 with P=(N-K)/N 🤷🏼 ).
In short, this impl will never need more than N/2 memory, and most often less, and your suggestion requires N at all inputs.

PR #246, opened 2023-01-18
switch jest for vitest

yanick:
This is less of a merge request, and more of a "just want to share with the group for giggles".

I was wondering how hard it'd be to switch from jest to vitest. The answer was "pleasantly trivial". But, at least on my machine, vitest is actually slower than jest (22ish seconds versus 13ish seconds).

So there we go. If we ever want to switch to vitest, it'd be easy, but purely for performance, it doesn't seem to be a winner.


jansedlon:
Hey @yanick. This caught my eye so I sent you a PR to your fork. I improved the test run time down from 5s to 800ms. I also added vitest globals so there's no need to import `describe`, `it`, etc.. 

What improved the run time is basically just turning off the environment isolation. I'm not sure if it's need in this project.

yanick:
> Hey @yanick. This caught my eye so I sent you a PR to your fork. I improved the test run time down from 5s to 700ms. I also added vitest globals so there's no need to import `describe`, `it`, etc..
> 
> What improved the run time is basically just turning off the environment isolation. I'm not sure if it's need in this project.

Niiiiiice. That's a hefty performance boost!

src/pipe.ts
```diff
@@ -76,7 +76,7 @@ export function pipe(
   const lazyOps = operations.map(op => {
     const { lazy, lazyArgs } = op as LazyOp;
     if (lazy) {
-      const fn: any = lazy(...lazyArgs);
+      const fn: any = lazy(...(lazyArgs ?? []));
```
TkDodo, line 79: why was this change made ?

TkDodo:
also getting an issue when deploying to netlify:

> error vitest@0.27.2: The engine "node" is incompatible with this module. Expected version ">=v14.16.0". Got "10.24.1"

src/pipe.ts
yanick: > why was this change made ?

Butt protection. If lazyArgs is not provided, the destructuration of `undefined` would go :boom:.

yanick:
> also getting an issue when deploying to netlify:
> 
> > error vitest@0.27.2: The engine "node" is incompatible with this module. Expected version ">=v14.16.0". Got "10.24.1"

Is there something in the repo that provides netlify with a config (minimum node version and such)? I'm `git grep`ping netlify but can't see anything.

TkDodo:
I don't have access to the netlify build. @lstkz could you take a look please or grant me access?

src/pipe.ts
TkDodo: types make sure that you only pass functions into a pipe, not something that can be undefined / null. since this PR is about changing the test framework, I'd revert that change please.

lstkz:
@TkDodo 
I can't add you beacuse it requires additional payment :/
The error is:
```
4:48:32 PM: error vitest@0.27.2: The engine "node" is incompatible with this module. Expected version ">=v14.16.0". Got "10.24.1"
```
node version needs to be updated in package.json

lstkz:
@TkDodo 
Can you set the node version inside the repo?
https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript

package.json
```diff
@@ -7,19 +7,19 @@
   "jsnext:main": "dist/es/index.js",
   "sideEffects": false,
   "devDependencies": {
-    "@types/jest": "^26.0.14",
-    "jest": "^26.4.2",
+    "@vitest/coverage-c8": "^0.27.2",
     "marked": "^0.4.0",
     "prettier": "^2.8.0",
     "ts-jest": "^26.4.1",
```
TkDodo, line 13: do we still need `ts-jest` ?

TkDodo: we now have a peerDependency warning:

> warning " > ts-jest@26.5.6" has unmet peer dependency "jest@>=26 <27".

TkDodo:
looking good:

> 2:42:26 PM: Attempting Node.js version '16.16.0' from .nvmrc
2:42:27 PM: Downloading and installing node v16.16.0...
2:42:27 PM: Downloading https://nodejs.org/dist/v16.16.0/node-v16.16.0-linux-x64.tar.xz...
2:42:27 PM: Computing checksum with sha256sum
2:42:27 PM: Checksums matched!
2:42:29 PM: Now using node v16.16.0 (npm v8.11.0)

src/pipe.ts
yanick: > types make sure that you only pass functions into a pipe, not something that can be undefined / null.

Assuming the end-user is using TypeScript, and didn't throw an any somewhere. I was thinking along the lines of "be strict in what you emit, lax in what you accept". But you are right, it's not related to the change of testing framework. I'll remove the line presently!

yanick: Okay, this is funky. I removed that protection, and the code complains that lazyArgs is undefined. Okay, fair, that's probably why I put the code there in the first place. But it works on master. So I went back there and did what expert programmers do: added print statements. The result: https://github.com/remeda/remeda/compare/master...yanick:undefined-yet-spreadable?expand=1  

And... the log says that lazyArgs for `first` is undefined, and yet the next line that spread it out doesn't go boom. Even better, if I change the log to be `console.log('!', lazyArgs, [ ...lazyArgs ] )`, that one goes boom.

I suspect it's something I changed in the tsconfig? I'll bisect the changes and find out.

yanick: > I suspect it's something I changed in the tsconfig? I'll bisect the changes and find out.

I think? I have an idea what's going on. `tsc` turns `lazy(...lazyArgs)` into `lazy.apply(void 0, lazyArgs)`, which is totally okay with `lazyArgs` being undefined. But `vitest` doesn't use `tsc` and mostly just strip away the type annotations, so the spread remains a spread.

So now I have to see why `first` in that test doesn't have a lazyArgs...

yanick: I think I got it. Basically, calling `first` instead of `first()` in the pipe only work if we explicitly define the lazyArgs to be [].

TkDodo: okay I mean we can keep the fallback, even it's only to make vite happy 🤷 . But please use `||` as nullish colesce will likely be transpiled to something larger

yanick: Will do in a few seconds! And you got me curious: the `??` would be transpiled into 

```
// lazyArgs ?? [] turns into --v
(lazyArgs !== null && lazyArgs !== void 0 ? lazyArgs : [])
```

PR #242, opened 2023-01-16
feat: Add zipObj

itelo:
Resolves (https://github.com/remeda/remeda/issues/72)

---

Make sure that you:

- [x] Typedoc added for new methods and updated for changed
- [x] Tests added for new methods and updated for changed
- [x] New methods added to `src/index.ts`
- [x] New methods added to `mapping.md`

PR #248, opened 2023-01-20
add zipObj

yanick:
For #72 

There are two TBDs in there:

1. If there are more values than keys passed in, I throw an error. Do we want to be more forgiving?
2. I've added the data last, curried version of the function, but in this instance it doesn't seem super-useful Should it be there for uniformity's sake, or should it be removed?

---

Make sure that you:

- [x ] Typedoc added for new methods and updated for changed
- [ x] Tests added for new methods and updated for changed
- [ x] New methods added to `src/index.ts`
- [ x] New methods added to `mapping.md`


TkDodo:
we already have an open PR for this:

- #242

can we please consolidate the efforts there?

PR #259, opened 2023-02-10
chore(sortBy): better type inference for data last

### LINKED ISSUE #249: sortBy type inference
# Context

When using `sortBy` with `SortPair<T>`, TS fail to infer the correct declaration.

# To reproduce the error

```typescript
const array = [
  { color: 'red', weight: 2 },
  { color: 'blue', weight: 3 },
  { color: 'green', weight: 1 },
  { color: 'purple', weight: 1 },
];

R.sortBy(array, [(x) => x.weight, 'asc']); // works correctly

R.pipe(
  array,
  R.sortBy([(x) => x.weight, 'asc']) // raise an error
);
```

# Expected result

The array `[(x) => x.weight, 'asc']` should be inferred as `SortPair<T>` and thus the `sortBy` declaration picked by TS is the "data-last" form.

# What happen

Due to declaration order, the declaration of `sortBy` picked by TS is the first one :
```ts
export declare function sortBy<T>(array: readonly T[], ...sorts: SortRule<T>[]): T[];
```
and thus the `[(x) => x.weight, 'asc']` leads to an inferred `type T = (string | ((x: any) => any))[]`

If "data-last" declaration is moved before the "data-first" one, TS infer the correct form by default.

balzdur:
fix #249

### DIFF
--- a/src/sortBy.test.ts
+++ b/src/sortBy.test.ts
@@ -1,5 +1,6 @@
 import { sortBy } from './sortBy';
 import { pipe } from './pipe';
+import { AssertEqual } from './_types';
 
 const items = [{ a: 1 }, { a: 3 }, { a: 7 }, { a: 2 }] as const;
 const sorted = [{ a: 1 }, { a: 2 }, { a: 3 }, { a: 7 }];
@@ -62,6 +63,21 @@ describe('data first', () => {
       [3, 2, 1, 1]
     );
   });
+
+  describe('sortBy typings', () => {
+    test('SortProjection', () => {
+      const actual = sortBy(items, x => x.a);
+      type T = typeof items[number];
+      const result: AssertEqual<typeof actual, T[]> = true;
+      expect(result).toEqual(true);
+    });
+    test('SortPair', () => {
+      const actual = sortBy(objects, [x => x.active, 'desc']);
+      type T = typeof objects[number];
+      const result: AssertEqual<typeof actual, T[]> = true;
+      expect(result).toEqual(true);
+    });
+  });
 });
 
 describe('data last', () => {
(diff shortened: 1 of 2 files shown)


balzdur:
When multiple overloads can apply to a situation, TS seems to pick the first one when inferring the types (TBC, I'm not an expert on TS)

From the linked issue, you can see that :
```ts
const array = [
  { color: 'red', weight: 2 },
  { color: 'blue', weight: 3 },
  { color: 'green', weight: 1 },
  { color: 'purple', weight: 1 },
];

R.pipe(
  array,
  R.sortBy([(x) => x.weight, 'asc']) // raise an error
);
```

The array `[(x) => x.weight, 'asc']` should be inferred as SortPair<T> but the declaration used by TS in inference process is the other one :
```ts
export declare function sortBy<T>(array: readonly T[], ...sorts: SortRule<T>[]): T[];
```
and thus the [(x) => x.weight, 'asc'] leads to an inferred type T = string | ((x: any) => any)

A workaround : enforce TS to pick the right declaration by doing (credit [here](https://github.com/remeda/remeda/issues/249#issuecomment-1424279887)):
```ts
R.pipe(
  array,
  R.sortBy<typeof array[number]>([x => x.weight, 'asc']) // T = string
);
```

TkDodo:
ooh, it's because they are both arrays (well, one is a tuple, but that doesn't matter). Yes, TS  picks overloads by order, so yes, this does help. Can we make sure that it doesn't introduce the inverse problem, like picking the wrong overload now when we want the `data-first` version? It seems like we don't really have any type-level tests yet. Have a look at the `reverse.test.ts` and the `AssertEqual` util.

TkDodo (approved):

PR #260, opened 2023-02-13
feat(groupBy): add support for generic key types

eranhirsch:
`groupBy` currently doesn't support refined types for it's grouper function, so you lose typing information in cases where those are available.

This diff only touches typing, and doesn't change anything in the runtime behaviour of the function.

It maintains backwards compatibility by still returning `Record<string, T>` if the grouper function returns a string. And for literal unions it will return a Partial record.

I tried adding tests, and they do provide some protection, but I'm not sure I'm using AssertEqual correctly, as it doesn't fail on the wrong types (sometimes). I validated manually via the IDE that the types returned are correct. 

Fixes: #262

---

Make sure that you:

- [?] Typedoc added for new methods and updated for changed
- [+] Tests added for new methods and updated for changed
- [-] New methods added to `src/index.ts`
- [-] New methods added to `mapping.md`

### DIFF
--- a/src/groupBy.test.ts
+++ b/src/groupBy.test.ts
@@ -84,4 +84,13 @@ describe('Result key types', () => {
     > = true;
     expect(result).toEqual(true);
   });
+  test('string | number', () => {
+    const data = groupBy(array2, (x):  string | number => x.b);
+    const result: AssertEqual<
+      typeof data,
+      Record<string | number, NonEmptyArray<Array2Item>>
+    > = true;
+    expect(result).toEqual(true);
+  });
 });
+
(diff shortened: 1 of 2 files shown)


TkDodo:
👍 can you please fix the formatting?

eranhirsch:
@TkDodo - should the better typing actually be implemented as a `strict` version of groupBy. My thoughts on this are that it can go either way. It would keep the existing logic backwards compatible for those already relying on it, but the downside is that it's redundant because it's achievable by widening the type of the grouper function to `string` or `number`.

TkDodo:
@eranhirsch yeah I like the idea of doing a `strict` version of that 👍

eranhirsch:
@TkDodo - I've moved the new typing to a strict variant

TkDodo (approved):

PR #265, opened 2023-02-25
feat(toPairs): Add 'strict' flavour

eranhirsch:
Similar to `keys`, the existing `toPairs` function strips the typing of the object keys. Taking inspiration and largley copying the work in ts-extra's objectEntries, this version of toPairs will return the strictest type for the entries tuple.

I've added it as a strict option instead of replacing the `toPairs` impl for backwards compatibility and to align with `keys`.
---

Make sure that you:

### DIFF
--- a/src/toPairs.ts
+++ b/src/toPairs.ts
@@ -3,10 +3,37 @@
  * @param object
  * @signature
  *    R.toPairs(object)
+ *    R.toPairs.strict(object)
  * @example
  *    R.toPairs({ a: 1, b: 2, c: 3 }) // => [['a', 1], ['b', 2], ['c', 3]]
+ *    R.toPairs.strict({ a: 1 } as { a?: number }) // => [['a', 1]] typed Array<['a', number]>
+ * @strict
  * @category Object
  */
 export function toPairs<T>(object: { [s: string]: T }): Array<[string, T]> {
   return Object.entries(object);
+}
+
+// Inspired and largely copied from `sindresorhus/ts-extras`:
+// @see https://github.com/sindresorhus/ts-extras/blob/44f57392c5f027268330771996c4fdf9260b22d6/source/object-keys.ts
+// @see https://github.com/sindresorhus/ts-extras/blob/44f57392c5f027268330771996c4fdf9260b22d6/source/object-entries.ts
+type ObjectKeys<T extends object> = `${Exclude<keyof T, symbol>}`;
+type ObjectValues<T extends Record<PropertyKey, unknown>> =
+  Required<T>[ObjectKeys<T>];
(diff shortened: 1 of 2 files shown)


TkDodo (approved):

src/toPairs.ts
```diff
@@ -3,10 +3,37 @@
  * @param object
  * @signature
  *    R.toPairs(object)
+ *    R.toPairs.strict(object)
  * @example
  *    R.toPairs({ a: 1, b: 2, c: 3 }) // => [['a', 1], ['b', 2], ['c', 3]]
+ *    R.toPairs.strict({ a: 1 } as { a?: number }) // => [['a', 1]] typed Array<['a', number]>
```
TkDodo, line 9: it looks a bit weird in the docs to have a type assertion. would this also work with `as const` ?

```suggestion
 *    R.toPairs.strict({ a: 1 } as const) // => [['a', 1]] typed Array<['a', number]>
```

eranhirsch: I got so fixated on the optionality aspect of things that I forgot that this is just a general example. I'll change it (the suggestion you provided has the wrong type for the values, it should be `1` and not `number`).

eranhirsch:
btw, any reason this is called 'toPairs' and not entries? Feels like most JS developers would try to find a method with that name when looking for this...

TkDodo:
> btw, any reason this is called 'toPairs' and not entries? Feels like most JS developers would try to find a method with that name when looking for this...

predates me, but I think because `lodash` also has `toPairs`, and I think this existed before `Object.entries` was a thing. We can think about renaming it in the next major.

@vlad-yakovlev FYI

eranhirsch:
I guess having _both_ toPairs and entries and having one simply alias the other is not a great API?

TkDodo:
it would surely be possible in a minor, but eventually, I'd like to settle on one naming :)

TkDodo:
the question is: do we want previous lodash users to feel like they can easily adjust to the lib, or should it be geared towards users who know the es6 api. previously, we were definitely geared towards getting adoption from lodash / ramda users, hence also the mapping table

eranhirsch:
Makes sense, anyway I love Remeda. I'll be happy to help with anything you need around here

TkDodo:
Thanks @eranhirsch. Let me know if there's a different way how I can reach you, preferably twitter or discord :)

eranhirsch:
What discord server(?) should I join? (I have to admit that I feel completely lost on Discord)

TkDodo:
doesn't have to be a specific server (we don't have one for remeda), but you can DM me: TkDodo#3079

PR #266, opened 2023-03-02
chore(lint): Add eslint with recommended rules

eranhirsch:
Adding eslint to the project.

It's configured to run only on the `src` dir (skipping the docs and the scripts).

It runs the eslint recommended config, and typescript's recommended, type-checked recommended, and strict rules, with almost no extra rule configs. It also uses the prettier config to remove any styling rules from lint, as recommended by prettier.

There are a bunch of TODOs in the config right now for rules that don't jive with the existing codebase. Some of these would be fixed in subsequent PRs trivially (autofixing and manual fixing one liners); but some require a little more care and thought as they change idioms or even the logic itself. **The goal of this PR is just to add lint going forward, without changing anything in the codebase itself**..

---

Make sure that you:


eranhirsch:
I removed all changes to actual code from the PR so that it's easier to review. Once this is pushed we can send out 2 more PRs to bring the source to a lint-satisfied state.

TkDodo:
building failed with:

```
8:53:06 AM: error @typescript-eslint/eslint-plugin@5.54.0: The engine "node" is incompatible with this module. Expected version "^12.22.0 || ^14.17.0 || >=16.0.0". Got "10.24.1"
```

I fixed this on a different PR by adding an `.nvmrc` file:
- https://github.com/remeda/remeda/pull/246

merged this just now, so please update

eranhirsch:
Done!

TkDodo (approved):

PR #271, opened 2023-03-04
chore(lint): autofixers

eranhirsch:
Enabling all rules with autofixers and then running them.

There is literally no manual changes done in this PR, so if we trust the autofixers (which I think is reasonable) we can feel confident in the changes.

For easier review I split the PR into 2 commits, one without the "@typescript-eslint/array-type" rule, and one with only that (as it changes a lot of files).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -50,13 +50,6 @@ module.exports = {
 
     // TODO: These all have autofixers
     '@typescript-eslint/array-type': 'off',
-    '@typescript-eslint/ban-tslint-comment': 'off',
-    '@typescript-eslint/consistent-indexed-object-style': 'off',
-    '@typescript-eslint/no-inferrable-types': 'off',
-    '@typescript-eslint/no-unnecessary-boolean-literal-compare': 'off',
-    '@typescript-eslint/prefer-reduce-type-parameter': 'off',
-    '@typescript-eslint/prefer-ts-expect-error': 'off',
-    'prefer-const': 'off',
 
     // TODO: These have trivial manual fixes
     '@typescript-eslint/ban-ts-comment': 'off',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,6 +26,14 @@ module.exports = {
     // etc...) via namespaces by design.
     '@typescript-eslint/no-namespace': 'off',
 
+    '@typescript-eslint/array-type': [
+      'warn',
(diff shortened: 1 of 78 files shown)


.eslintrc.js
```diff
@@ -48,16 +48,6 @@ module.exports = {
     // `function foo(...args: readonly unknown[])`)
     'prefer-rest-params': 'off',
 
-    // TODO: These all have autofixers
-    '@typescript-eslint/array-type': 'off',
```
TkDodo, line 52: Can we set this to prefer `Array<>` over `[]` ?

```
'@typescript-eslint/array-type': ['error', { default: 'generic', readonly: 'generic' }],
```

I'm aware it's a preferential choice 😅 but there are a couple of reasons why I think this syntax is "objectively better"

TkDodo: thanks! but why level `warn` and not `error` ?

eranhirsch: I had a sense that the winds are shifting towards "array" styling, to me it just makes it more obvious to non-native Typescript readers that this is a simple "native" array and not something special... (also you can't redefine the Array type and completely mess with people's head by accident)

eranhirsch: > thanks! but why level `warn` and not `error` ?

It's the default for this rule.

eranhirsch:
The Generic format for arrays is more verbose, and i'm guessing that meant that a few more lines were added to the codebase, i'm guessing codecov is having a hard time with that...

.eslintrc.js
TkDodo: but we're not enforcing zero warnings, right? In my experience, eslint warnings are often just ignored

eranhirsch: I never put much thought to why something would be warning and something else error. I personally would still comment in a PR which has a lint warning, and they are both suppressesable all the same using a lint disable comment.

My only thought of keeping it a warning is that it communicates better to anyone who'll reconsider this rule in the future that the original author didn't think this rule is critical so it might be ok to change it (versus a rule set as error), but that's pretty much it, i dunno...

I don't mind changing it, I actually didn't notice your code snippet in the original comment, I just saw you said you like generics and went ahead and changed it...

TkDodo: thanks 🙏

TkDodo (approved):

PR #274, opened 2023-03-06
chore(deps): bump versions

eranhirsch:
This includes minor updates to:

* typescript 4.7 -> 4.9, providing some neat typing improvements, perf benefits, and the satisifies operator
* prettier fixes, basically to support satisfies, and minor fixes.

* vitest - I don't know what changed, seem like bugfix updates

Tested:
```bash
yarn lint
yarn prettier
yarn test
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,15 +9,15 @@
   "devDependencies": {
     "@typescript-eslint/eslint-plugin": "^5.54.0",
     "@typescript-eslint/parser": "^5.54.0",
-    "@vitest/coverage-c8": "^0.27.2",
+    "@vitest/coverage-c8": "^0.29.2",
     "eslint": "^8.35.0",
     "eslint-config-prettier": "^8.6.0",
     "marked": "^0.4.0",
-    "prettier": "^2.8.0",
+    "prettier": "^2.8.4",
     "ts-node": "^8.4.1",
     "typedoc": "^0.11.1",
-    "typescript": "^4.7.2",
-    "vitest": "^0.28.4"
+    "typescript": "^4.9.5",
+    "vitest": "^0.29.2"
   },
   "scripts": {
     "clean": "rm -rf ./dist && mkdir dist",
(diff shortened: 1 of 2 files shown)


TkDodo:
> typescript 4.7 -> 4.9, providing some neat typing improvements, perf benefits, and the satisifies operator

I'm not so sure about that one. Please keep in mind that we probably shouldn't use latest TS features. There is no real policy here (maybe we should have one?) but if we start to use TS 4.9 features, it means that our consumers must also be on TS4.9 (unless we provide differential typings for different TS versions, which is possible, but harder to maintain).

I actually don't know what the current minimum version of TS is that we need. There is something going on in this PR where we talked about needing 4.7 at a minimum:

- https://github.com/remeda/remeda/pull/222#discussion_r1020774509

eranhirsch:
Those are really good points i didn't think about. I wonder how the TS team see this, there are advantages to the dev experience using the latest TS which would be odd to deny from package maintainers that want a broad user-base. Are there any compatibility mode flags in the tsconfig file that could be used to lock the output version?

TkDodo:
the only thing I know of is shipping different types depending on the typescript version like so: https://github.com/TanStack/query/blob/5e37b145a8836b00dd20860e8eb1f94c608fec81/package.json#L83-L85

the thing is, if we update the requirements in a minor, the lib will likely break for users that have a smaller TS version. In react-query, we only increment those requirements in majors. With the next one, we'll go from at leat 4.1 to at least 4.7. But it will be some time until we can use `satisfies` (4.9) or const inference (5.0) ...

eranhirsch:
Ok, so this isn't really important for now...

airjp73:
> I actually don't know what the current minimum version of TS is that we need. There is something going on in this PR where we talked about needing 4.7 at a minimum:
> 
> https://github.com/remeda/remeda/pull/222#discussion_r1020774509

Yeah, in one of the earlier versions of the PR I was using the `extends` keyword after an `infer` (e.g. `infer Foo extends string`). The current iteration doesn't need that at all.

PR #286, opened 2023-03-15
perf: improve performance of _reduceLazy

rodrigofariow:
Performance concerns regarding iterating over an array are not always relevant. In fact, most of the time it does not matter.
However, for a library like this one, where you don't know exactly how it will be used, it's better to assume the worst-case scenario.
My team, for instance, makes heavy use of [google-map-react](https://github.com/google-map-react/google-map-react) which requires you to iterate over thousands and thousands of potential markers.
The usage of `useMemo` there is difficult so we are often left with the worst-case scenario, which is to have an array of thousands of items be rendered every time you move the map around or zoom, for instance.

Back to the point, we want to try to use a single library for most things and right now we were "forced" to use custom functions because the difference was noticeable.

This example might be a bit off-topic but hopefully, it will give you some context on the point I'm trying to make.
I think there is more room for improvement on other functions but I started with `_reduceLazy` ~~and `flatMap`~~ to see if it aligns with the goals of the library.

You are most likely familiar with this topic but I will leave a useful link here anyway:
https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/

Cheers!


TiagoPortfolio (approved):
LGTM!

src/flatMap.ts
```diff
@@ -38,8 +38,16 @@ export function flatMap() {
   return purry(_flatMap, arguments, flatMap.lazy);
 }
 
+function _map<T, K>(array: Array<T>, mapFn: (value: T) => K): Array<K> {
+  const newArray = new Array<K>(array.length);
```
eranhirsch, line 42: I know unicorn eslint consider this a warning and suggest a different syntax. I don't know the reasoning behind this but perhaps it's worthwhile to conform to that:

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-new-array.md

src/flatMap.ts
```diff
@@ -38,8 +38,16 @@ export function flatMap() {
...
+  for (let i = 0; i < array.length; i++) {
+    newArray[i] = mapFn(array[i]);
+  }
+  return newArray;
+}
+
```
eranhirsch, line 48: Why reimplement map in the flatMap source and not use our own mapper (which uses the reduceLazy that was optimized). I feel like this complicates the code unnecessarily. At least I would add some documentation above this with a link to the article and some context on why it's better than mapper, otherwise a future maintainer might just refactor it back to a regular map, or even worse, if v8 ever becomes more efficient and using the build-in map would be better no one would clean this up

src/flatMap.ts
rodrigofariow: Due to the reasons explained on https://github.com/remeda/remeda/pull/286#discussion_r1140460891, I think it would be worth disabling that rule for this specific use case.

rodrigofariow:
@eranhirsch I removed the changes on `flatMap` since it's a bit out of scope and a very isolated use case.
If the maintainers agree, I'm willing to provide deeper changes with some new tests as well for these "internal" functions

src/flatMap.ts
eranhirsch: I think the only real change is doing `Array.from<K>({ length: array.length })` instead of `new Array<K>(array.length)`. I don't think it's related to the comment you linked to

src/_reduceLazy.ts
```diff
@@ -25,14 +25,17 @@ export function _reduceLazy<T, K>(
   array: Array<T>,
   lazy: (item: T, index?: number, array?: Array<T>) => LazyResult<K>,
   indexed?: boolean
-) {
-  return array.reduce((acc: Array<K>, item, index) => {
+): Array<K> {
+  const newArray: Array<K> = [];
+  // We intentionally use a for loop here instead of reduce for performance reasons. See https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/ for more info
+  for (let index = 0; index < array.length; index++) {
```
eranhirsch, line 31: You should use a `for...of` loop (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) here. It's stricter as it doesn't use an index into the array (see #291) and is transpiled to the same code on es5 anyway.

src/_reduceLazy.ts
```diff
@@ -25,14 +25,17 @@ export function _reduceLazy<T, K>(
   array: Array<T>,
   lazy: (item: T, index?: number, array?: Array<T>) => LazyResult<K>,
   indexed?: boolean
-) {
-  return array.reduce((acc: Array<K>, item, index) => {
+): Array<K> {
+  const newArray: Array<K> = [];
```
eranhirsch, line 29: This should probably be called "ret" or "output" or "reduced" or something like that, as newArray is not really describing the semantics. But that's just my pet peeve so feel free to ignore...

eranhirsch (approved):

src/_reduceLazy.ts
eranhirsch: Hey, sorry for the mistaken suggestions, you can ignore them, while we compile down to es5 they aren't relevant. Your solution is on-point in the current configuration.

PR #287, opened 2023-03-16
feat(map): strict mode

eranhirsch:
map can be used on more than just simple arrays. I added a `strict` mode to map which would work with tuples and other complex array types. This allows us to map over stuff like NonEmptyArray's without losing the non-empty-ness. It also transfers any labels inside the tuple to the output.

```ts
type Square = [x: number, y: number, width: number, height: number];

function foo(square: Square): Square {
  return map.strict(square, x => x * 2);
}

// Valid, and even maintains the tuple labels
foo(foo([1, 1, 1, 1]));
```

[It is inspired by how Promise.all was typed in es2015.promise lib.](https://github.com/microsoft/TypeScript/blob/9ccf47fec5c1c87061956ac6e1446677654a4d4b/src/lib/es2015.promise.d.ts#LL15-L21)

---

Make sure that you:


src/map.ts
```diff
@@ -3,20 +3,31 @@ import { _reduceLazy, LazyResult } from './_reduceLazy';
...
  *    R.map.indexed(array, fn)
+ *    R.map.strict(array, fn)
+ *    R.map.strict.indexed(array, fn)
  * @example
- *    R.map([1, 2, 3], x => x * 2) // => [2, 4, 6]
- *    R.map.indexed([0, 0, 0], (x, i) => i) // => [0, 1, 2]
+ *    R.map([1, 2, 3], x => x * 2) // => [2, 4, 6], typed number[]
+ *    R.map.indexed([0, 0, 0], (x, i) => i) // => [0, 1, 2], typed number[]
+ *    R.map.strict([0, 0] as const, x => x + 1) // => [1, 1], typed [number, number]
+ *    R.map.strict.index([0, 0] as const, (x, i) => x + i) // => [0, 1], typed [number, number]
```
TkDodo, line 26:
```suggestion
 *    R.map.strict.indexed([0, 0] as const, (x, i) => x + i) // => [0, 1], typed [number, number]
```

src/map.ts
```diff
@@ -76,6 +87,31 @@ export namespace map {
...
+
+  export function strict<T extends ReadonlyArray<unknown> | [], K>(
+    array: Readonly<T>,
+    fn: Pred<T[number], K>
+  ): Mapped<T, K>;
+  export function strict<T extends ReadonlyArray<unknown> | [], K>(
+    fn: Pred<T[number], K>
+  ): (array: Readonly<T>) => Mapped<T, K>;
+  export function strict(...args: ReadonlyArray<unknown>): unknown {
+    return purry(_map(false), args, map.lazy);
```
TkDodo, line 102: instead of re-implementing `map` inside `strict`, could we just invoke the regular `map` function - similar to how `keys` is implemented? I realize now that `groupBy` also doesn't do that ...

TkDodo (approved):

PR #288, opened 2023-03-17
feat(chunk): improve typing and perf

eranhirsch:
Two changes to the way chunk types it's output:
* Chunks are never empty
* If the input is non-empty, the output is non-empty

---


src/_types.ts
```diff
@@ -18,4 +18,4 @@ export type AssertEqual<Type, Expected> = Array<Type> extends Array<Expected>
     : never
   : never;
 
-export type NonEmptyArray<T> = [T, ...Array<T>] | [...Array<T>, T];
```
eranhirsch, line 21: I had to revert the changes here, there's a difference in how typescript treats the array when it's a return value (as in groupBy) and when it's used as an input value (like I did with chunk).

TkDodo (approved):

PR #290, opened 2023-03-18
fix(first): better typing

eranhirsch:
There are several situations were the type of the first value could be inferred more precisely than just `T[0] | undefined`; when the array is not empty, and when taking into consideration arrays used as tuples too.

---

### DIFF
--- a/src/first.test.ts
+++ b/src/first.test.ts
@@ -133,6 +133,20 @@ describe('strict typing', () => {
     expect(result).toEqual(undefined);
   });
 
+  test('array with last', () => {
+    const arr: [...Array<number>, number] = [1];
+    const result = first(arr);
+    expectTypeOf(result).toEqualTypeOf<number>();
+    expect(result).toEqual(1);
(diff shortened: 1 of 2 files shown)


TkDodo:
looks really good, thank you. Maybe as a follow-up, the same could be applied to the `last` function?

eranhirsch:
> looks really good, thank you. Maybe as a follow-up, the same could be applied to the `last` function?

I started working on `last` but it's more challenging because there's no way to get the equivalent of `T[0]` for the _last_ item in a tuple.

TkDodo:
We could reverse the tuple like the reverse function does and then take the first element?

eranhirsch:
I think I can get it to work with a bunch of `infer`. It's just a matter of mapping all the possible ways to build a tuple which would affect the type of last:
1. any[]
2. [any, any, any]
3. [...any[], any]
4. [any, ...any[]]
5. [any?, any?, any?]

etc...

PR #292, opened 2023-03-23
chore(git-hooks): Automatically lint and format files

eranhirsch:
Adding husky and lint-staged, and configuring 2 githooks to make sure files are always formatted, linted, and tested when pushed to github.


TkDodo:
hm, I don't think this is important enough to have checked on every dev's machine. Sometimes, test don't work locally, and you want to test them in CI. It will also prohibit larger refactorings where some things are known to break in-between, but then I can't push to back-up my work. Nothing broken can get into main due to our CI checks, so I'm not sure what the advantage of running everything locally on pre-commit/push, except that will slow things down by a couple of seconds every time...

eranhirsch:
I can see the problems with the push hook that tests, but I think the commit hook is still useful as it simply automates stuff that you should do anyway. I've already sent a few commits with unformatted code just because we don't have a .vscode settings file which turns on format on save for the project (and I have it off by default to not format projects which don't have formatting). I feel that the husky hook is the right way to enforce formatting instead of the format-on-save.

TkDodo (approved):

PR #293, opened 2023-03-23
add 'matches' function

yanick:
add a `matches` function like lodash's _.matches. Basically allows to do flexible, deep comparisons on objects.

---

Make sure that you:

### DIFF
--- a/src/matches.ts
+++ b/src/matches.ts
@@ -4,9 +4,12 @@ function _matches(target: any, condition: any) {
   if (typeof condition === 'function') return condition(target);
 
   if (typeof condition === 'object') {
-    return Object.entries(condition).every(([key, value]) =>
-      matches(target[key], value)
-    );
+    for (const key in condition) {
+      if (!_matches(target[key], condition[key])) {
+        return false;
+      }
+    }
+    return true;
   }
 
   return target === condition;
--- a/src/matches.ts
+++ b/src/matches.ts
@@ -1,7 +1,9 @@
 import { purry } from './purry';
 
 function _matches(target: any, condition: any) {
-  if (typeof condition === 'function') return condition(target);
+  if (typeof condition === 'function') {
+    return condition(target);
+  }
 
   if (typeof condition === 'object') {
     for (const key in condition) {
--- a/src/matches.ts
+++ b/src/matches.ts
@@ -1,0 +1,58 @@
+import { purry } from './purry';
+
+function _matches(target: any, condition: any) {
+  if (typeof condition === 'function') return condition(target);
+
+  if (typeof condition === 'object') {
+    return Object.entries(condition).every(([key, value]) =>
+      matches(target[key], value)
+    );
+  }
+
+  return target === condition;
+}
+
+type Matcher<I = any> = ((dataInput: I) => boolean) | any;
+
+/**
+ * Compares the input with the matcher and returns `true` if they match.
+ * The matcher can be a function (which will be fed the input and is expected
+ * to return a boolean), or a value. If the value is an object, the matching
+ * will be recursive.
+ * @param input the input data
+ * @param matcher matching function or value
+ * @signature
+ *   R.matches(inputData, matcher)
+ * @example
+ *    R.matches( 'potato', 'potato'); // => true
+ *    R.matches( 'potato', 'turnip'); // => false
+ *    R.matches( 'potato', vegetable => vegetable === 'potato'); // => true
+ *    R.matches({ a: 1, b :2 }, { a: 1 } ); // => true
+ *    R.matches({ 'a': 4, 'b': 5, 'c': 6 }, { 'a': 4, 'c': 6 }) // => true
+ * @data_first
+ * @category Object
+ */
+export function matches<I>(data: I, matcher: Matcher<I>): boolean;
+
+/**
+ * Compares the input with the matcher and returns `true` if they match.
+ * The matcher can be a function (which will be fed the input and is expected
+ * to return a boolean), or a value. If the value is an object, the matching
+ * will be recursive.
+ * @param input the input data
+ * @param matcher matching function or value
+ * @signature
+ *   R.matches(matcher)(inputData)
+ * @example
+ *   R.pipe({a:1,b:2}, R.matches({ a: 1 }) ) // => true
+ *   R.pipe({a:1,b:2}, R.matches({ b: (val) => val < 5 }) ) // => true
+ *   R.pipe({a:1,b:2}, R.matches({ c: 3 }) ) // => false
+ *   R.pipe( { 'a': 4, 'b': 5, 'c': 6 }, R.matches({ 'a': 4, 'c': 6 })) // => true
+ * @data_last
+ * @category Object
+ */
+export function matches(matcher: Matcher): (dataIn: any) => boolean;
+
+export function matches() {
+  return purry(_matches, arguments);
+}
--- a/mapping.md
+++ b/mapping.md
@@ -46,6 +46,7 @@ documentation when migrating._
 | `length`           | `size`             | `length`            |
 | `map`              | `map`              | `map`               |
 | `mapKeys`          | `mapKeys`          | `-`                 |
+| `matches`          | `matches`          | `_`                 |
 | `meanBy`           | `meanBy`           | `-`                 |
 | `merge`            | `assign`           | `merge`             |
 | `mergeAll`         | `assign`           | `mergeAll`          |
--- a/src/index.ts
+++ b/src/index.ts
@@ -55,6 +55,7 @@ export * from './map';
 export * from './mapKeys';
 export * from './mapToObj';
 export * from './mapValues';
+export * from './matches';
 export * from './maxBy';
 export * from './meanBy';
 export * from './merge';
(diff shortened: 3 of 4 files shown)


src/matches.ts
```diff
@@ -0,0 +1,56 @@
...
+function _matches(target: any, condition: any) {
+  if (typeof condition === 'function') return condition(target);
+
+  if (typeof condition === 'object') {
+    return Object.entries(condition).every(([key, value]) =>
+      matches(target[key], value)
+    );
+  }
+
+  return target === condition;
```
TkDodo, line 12: should we compare with `Object.is` here?

src/matches.ts
```diff
@@ -0,0 +1,56 @@
...
+ * will be recursive.
+ * @param input the input data
+ * @param matcher matching function or value
+ * @signature
+ *   R.matches(inputData, matcher)
+ * @example
+ *    R.matches( 'potato', 'potato'); // => true
+ *    R.matches( 'potato', 'turnip'); // => false
+ *    R.matches( 'potato', vegetable => vegetable === 'potato'); // => true
+ *    R.matches({ a: 1, b :2 }, { a: 1 } ); // => true
```
TkDodo, line 30: can we have an example like in the lodash docs, and maybe also a test for that example? like:

```
matches({ 'a': 4, 'b': 5, 'c': 6 }, { 'a': 4, 'c': 6 })) // => true
```

src/matches.ts
```diff
@@ -0,0 +1,56 @@
...
+  if (typeof condition === 'object') {
+    return Object.entries(condition).every(([key, value]) =>
+      matches(target[key], value)
+    );
+  }
+
+  return target === condition;
+}
+
+type Matcher<I = any> = ((dataInput: I) => boolean) | any;
```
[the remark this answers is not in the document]
yanick: [default for generic] For the type at line 52. 

[ | any ] Because a matcher can also be anything. E.g.

```
matches({ id: '123 })(obj);
matches(true)(isValid);
```

And I know that `((dataInput: I) => boolean) | any` could simply be `any`, but a little explicitness for the most used case shouldn't hurt.

src/matches.ts
yanick: I don't think so. The target can be anything that is not an object or a function at that point.

src/matches.ts
yanick: Absolutely! Gimme a moment... done!

eranhirsch:
If this is still relevant I would love to review and merge it, @yanick , do you want to rebase it and fix any lint issues that might have surfaced since this was written?

yanick:
Rebased!

src/matches.ts
```diff
@@ -0,0 +1,58 @@
+import { purry } from './purry';
+
+function _matches(target: any, condition: any) {
```
eranhirsch, line 3: Let's use `unknown` here, I'm planning on turning on stricter typescript and we won't be able to use `any` anywhere
```suggestion
function _matches(target: unknown, condition: unknown) {
```

src/matches.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (typeof condition === 'function') return condition(target);
```
eranhirsch, line 4: Prefer blocks
```suggestion
  if (typeof condition === 'function') {
    return condition(target);
  }
```

src/matches.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  if (typeof condition === 'object') {
+    return Object.entries(condition).every(([key, value]) =>
+      matches(target[key], value)
+    );
```
eranhirsch, line 9: 1. Notice that Object.entries is es2017. You can use a regular for loop to make it compatible for more users.
2. You are calling the purried vesion of matches here, but you can simply call the implementation, there's no point going through purrying here..
```suggestion
    for (const key in condition) {
      if (!_matches(target[key], condition[key])) {
        return false;
      }
    }
    return true;
```

src/matches.ts
eranhirsch: Object.is is not just for objects, it supports any JS thing and it might be a little more accurate for a general purpose function like this, but it's also es2015 so using === would support more users (though we probably want to move to es2015 anyway).

Here's the answer from chatgpt :P

> Object.is:
Object.is is a method introduced in ECMAScript 6 (ES6) that compares two values for equality, similar to the strict equality operator (===). It returns true if the values are the same, both in terms of type and value. However, it has some special cases:
It considers +0 and -0 as not equal, unlike === which treats them as equal.
It considers NaN values as equal, unlike ===, which considers them not equal.
> In general, you can use === for most equality comparisons. However, if you specifically need to handle the special cases mentioned above, Object.is provides a more precise comparison.

src/matches.ts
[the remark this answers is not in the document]
eranhirsch: I agree with @TkDodo, these types don't make much sense because they are only for readability, they don't provide any type *safety* for users. In fact this allows users to send invalid inputs that would fail silently (mainly just returning false for them) without the users ever knowing about it until it explodes in production.

Can we work on the type of Matches a little more?
1. I think we can start with the fact that I should be `object`. This function doesn't make sense for any other use-cases because they can simply use `equals`.
2. Then it should map the props of the input object and recursively allow more mappers for any object values.

Here's a crude first attempt:
```suggestion
type Matcher<T extends object> = {
  [K in keyof T]?: T[K] extends object ? Matcher<T[K]> : ((value: T[K]) => boolean) | T[K]
}
```

src/matches.ts
```diff
@@ -0,0 +1,58 @@
...
+ * @example
+ *    R.matches( 'potato', 'potato'); // => true
+ *    R.matches( 'potato', 'turnip'); // => false
+ *    R.matches( 'potato', vegetable => vegetable === 'potato'); // => true
+ *    R.matches({ a: 1, b :2 }, { a: 1 } ); // => true
+ *    R.matches({ 'a': 4, 'b': 5, 'c': 6 }, { 'a': 4, 'c': 6 }) // => true
+ * @data_first
+ * @category Object
+ */
+export function matches<I>(data: I, matcher: Matcher<I>): boolean;
```
eranhirsch, line 35: Why are we using I for types instead of T?
```suggestion
export function matches<T>(data: T, matcher: Matcher<T>): boolean;
```

src/matches.ts
```diff
@@ -0,0 +1,58 @@
...
+ *   R.matches(matcher)(inputData)
+ * @example
+ *   R.pipe({a:1,b:2}, R.matches({ a: 1 }) ) // => true
+ *   R.pipe({a:1,b:2}, R.matches({ b: (val) => val < 5 }) ) // => true
+ *   R.pipe({a:1,b:2}, R.matches({ c: 3 }) ) // => false
+ *   R.pipe( { 'a': 4, 'b': 5, 'c': 6 }, R.matches({ 'a': 4, 'c': 6 })) // => true
+ * @data_last
+ * @category Object
+ */
+export function matches(matcher: Matcher): (dataIn: any) => boolean;
```
eranhirsch, line 54: This is entirely untyped...
```suggestion
export function matches<T>(matcher: Matcher<T>): (dataIn: T) => boolean;
```

eranhirsch (changes requested):
This effectively doesn't provide any typing for the functions, so the user can botch the matcher and wouldn't know about it.

src/matches.ts
yanick: From the tsconfig of the project, it seems that Node 10, which supports es2018, is deemed low enough a baseline. But sure.

src/matches.ts
yanick: Sorry, I can't take seriously any argument quoting chatGPT. I just.. can't.

src/matches.ts
yanick: That breaks.

src/matches.ts
yanick: I feel this is more personal preference than anything else, but I don't care that much either way. I'll merge the suggestion.

src/matches.ts
yanick: > these types don't make much sense because they are only for readability, 

Isn't readability a good thing?

yanick:
Sorry, I realize that I have a very different style than the reviewer, and I end up being more worked up than I really should. 

I'll back away from this PR. Feel free to modify it, be inspired by it, or discard it as you see fit. :-)

src/matches.ts
eranhirsch: With auto-formatting I agree that it's less likely that someone would introduce another statement and the if would break, but it's also just a matter of consistency. Anyway, unless we turn on "curly" you are right that his is a preference... feel free to ignore

src/matches.ts
TkDodo: I was suggesting `Object.is` because it is mostly equivalent to triple equals, except that it handles edge cases around NaN / zero better. It has nothing to do with the things to compare being an object. It is also what react uses under the hood for comparisons when updating state.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is

eranhirsch:
Feel free to re-open if you want to continue working on this

PR #294, opened 2023-03-28
feat: add shuffle method

Demivan:
Adds `shuffle` method. Creates an array of shuffled values, using the [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher-Yates_shuffle).

---

Make sure that you:


src/shuffle.ts
```diff
@@ -0,0 +1,24 @@
...
+ * [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher-Yates_shuffle).
+ * @param items the array to shuffle
+ * @signature
+ *    R.shuffle(array)
+ * @example
+ *    R.shuffle(['one', 'two', 'three']) // => ['two', 'three', 'one']
+ * @category Array
+ */
+export function shuffle<T>(items: ReadonlyArray<T>): Array<T> {
+  const length = items.length
```
eranhirsch, line 12: I don't think we need to memoize the length prop of the array, I can't imagine this having an impact on perf over accessing items.length directly

eranhirsch (changes requested):
I feel this is over-verbose for what it needs to be.

eranhirsch:
Also I just noticed this doesn't provide a data-last version

```typescript
function shuffle<T>(): (data: ReadonlyArray<T>) => Array<T>;
```

Demivan:
@eranhirsch Thanks for review. I have implemented your comments.

eranhirsch:
The code has non-consistent semi-colons, can you run `yarn lint && yarn prettier` on the branch?

Demivan:
Done

eranhirsch (approved):

src/shuffle.ts
```diff
@@ -0,0 +1,41 @@
+import { purry } from './purry';
+
+/**
+ * Creates an array of shuffled values, using the
+ * [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher-Yates_shuffle).
```
eranhirsch, line 5: No need to document the algorithm used, that is an internal that isn't relevant to the caller, and would become stale if we ever implement a different shuffling algorithm.

src/shuffle.ts
```diff
@@ -0,0 +1,41 @@
+import { purry } from './purry';
+
+/**
+ * Creates an array of shuffled values, using the
```
eranhirsch, line 4: I'd phrase this as "Shuffles the input array" as this doesn't create an array per-se (like you'd expect a "randomArray" function would create one)

eranhirsch (approved):
nits about the docs

src/shuffle.ts
Demivan: I wanted to indicate that it does not mutate the original array. "Shuffles the input array" sounds like it is mutating.

eranhirsch: All remeda functions shouldn't mutate the input, that's a core concept of Remeda.

Demivan: Updated

PR #299, opened 2023-04-10
feat: sample utility

eranhirsch:
Following the discussion in #245. An implementation of a sampling function that gets an array-like and a sample size and returns a random subset of that array.

* Maintains the order of the original array.
* Does not repeat items in the output.
* If K = sampleSize, N = arraySize, runs in approx. O(klogk) for small Ks, and O(n) for large ones.

A lot of the work here was around the typing. I wanted to make sure we pass along as much typing information from the input to the output. See the tests for more details about the implementation.

---


Tirke:
@eranhirsch Would be awesome to have this merged 😄

TkDodo (approved):

PR #300, opened 2023-04-12
feat(sort): strictm mode

eranhirsch:
I wanted to be able to sort a NonEmptyArray and keep it non empty (so that it could be used, for example, in a pipe with `first` to get the smallest item in a sort, maybe in a pipe with groupBy :wink:).

I made the types as generic as possible, so any sort of tuple combination would work, including mixed types.

---

Make sure that you:

### DIFF
--- a/src/sort.ts
+++ b/src/sort.ts
@@ -1,3 +1,4 @@
+import { IterableContainer } from './_types';
 import { purry } from './purry';
 
 /**
@@ -40,4 +41,23 @@ function _sort<T>(items: Array<T>, cmp: (a: T, b: T) => number) {
   const ret = [...items];
   ret.sort(cmp);
   return ret;
+}
+
+interface Strict {
+  <T extends IterableContainer>(
+    items: T,
+    cmp: (a: T[number], b: T[number]) => number
+  ): StrictOut<T>;
+
+  <T extends IterableContainer>(cmp: (a: T[number], b: T[number]) => number): (
+    items: T
+  ) => StrictOut<T>;
+}
+
+type StrictOut<T extends IterableContainer> = {
+  -readonly [P in keyof T]: T[number];
+};
+
+export namespace sort {
+  export const strict: Strict = sort;
 }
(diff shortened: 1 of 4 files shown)


src/sort.ts
```diff
@@ -41,3 +56,22 @@ function _sort<T>(items: Array<T>, cmp: (a: T, b: T) => number) {
   ret.sort(cmp);
   return ret;
 }
+
+interface Strict {
```
TkDodo, line 60: I'm not sure if we should have a `strict` version because of this, as opposed to a regular overload?

For example, the `last` function has an overload for non-empty arrays:

https://github.com/remeda/remeda/blob/6db8a67211760ea957fed02ab1cc5e96e7b32f61/src/last.ts#L23

and `reverse` also tries to infer the best possible return type depending on input:

https://github.com/remeda/remeda/blob/6db8a67211760ea957fed02ab1cc5e96e7b32f61/src/reverse.ts#L3-L12

The strict versions were meant to be "a type that is not really correct, but better for some cases", like we did for `Object.keys`. We then also added one for `groupBy`, because the normal overloads broke some other cases.

However, I can see that we also merged a strict version for `map`, so I'm not really sure how we would want to go from there, consistency wise 😅

eranhirsch: I don't think we can solve this with an overload because the non-strict version completely overlaps the typing of the strict version. Typescript wouldn't be able to differentiate between them so you'd have to put the stricter variant _before_ the non-strict one otherwise typescript won't be able to pick it up, and that would mean that de-facto the non-strict variant would never be used.

This would result in users now getting the stricter return type, where they previously got the less strict Array<T>. If their code assumes anything on this return type it might break.

The only way I can think of to provide better typing without being concerned with backwards compatibility is the `strict` variant. I think what we could do is bump remeda to 2 sometime and make the strict versions the main version.

This btw is true for all strict versions, and is apparent from the way we define the runtime const for `strict`. Typescript is aware that our strict versions effectively overlap the non-strict versions, and are not in-fact a different typing (like overloads provide).
```ts
export namespace myFunc {
   // This isn't a typescript error although they have different types, that is because
   // `Strict extends typeof myFunc`
   export const strict: Strict = myFunc;
}
```

TkDodo (approved):

PR #301, opened 2023-04-12
feat: join utility

eranhirsch:
Adding a better typed wrapper around the builtin Array.prototype.join, with data-first and data-last signatures.

Similar to https://github.com/sindresorhus/type-fest/blob/main/source/join.d.ts

---

Make sure that you:

### DIFF
--- a/mapping.md
+++ b/mapping.md
@@ -1,4 +1,4 @@
-### Mapping for Lodash and Ramda
+# Mapping for Lodash and Ramda
 
 _Remeda functions are not necessarily drop-in replacements for the
 listed Lodash and Ramda functions. Just as the Lodash function may behave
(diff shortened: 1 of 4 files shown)


TkDodo (approved):

TkDodo:
seems like this feature doesn't work with TS 4.4, but works with 4.8 ?

eranhirsch:
Good thing we have the new tests, i'll check it out

TkDodo:
> Good thing we have the new tests

my thoughts exactly

TkDodo:
I think 4.7 and onwards is fine, and we can set stricter standards (like > 5.0) in a new major

TkDodo (approved):

PR #304, opened 2023-04-13
feat(keys): strict tuple handling

eranhirsch:
Keys on non-empty arrays should be non-empty, but if we're at it we can also type them explicitly until the spread item shows up.

I also made the strict component a simple retyping of the regular function, instead of a reimplementation, inline with how we did strict in other functions.

---

Make sure that you:


src/keys.ts
```diff
@@ -18,18 +18,63 @@
...
+
 export namespace keys {
-  export function strict<T extends Record<PropertyKey, unknown>>(
-    source: T
-  ): Array<
-    { [K in keyof T]-?: K extends string | number ? `${K}` : never }[keyof T]
-  > {
-    return keys(source) as any;
-  }
+  // @ts-expect-error [ts2322] - I don't know why i'm getting this, the typing
```
TkDodo, line 77: hmm, we've had the experience in the past that these comments get stripped out during building, and then bring problems to people that don't have `skipLibCheck` turned on.

That was reported here:
- https://github.com/remeda/remeda/issues/163

eranhirsch: I've created #306 so that we can test our output files and avoid these issues without catching them in CR. Let's ship that and rebase and check it out

eranhirsch: I've rebased on top of #306 and the build is passing fine, I don't know though if that check is enough now, and I'm not sure I know how to proceed.

The problem is that this ts-expect-error is on the assignment to the strict variable, which has no bearing on the output type definitions, and the output logic itself is in JS so doesn't care about ts issues. I don't know how to simulate the skipLibCheck scenario here to see if it's relevant or not.

also, as far as I understand skipLibCheck is `false` by default, including in our own build ci tests, so I //think// we are fine here? 🤷🏼

TkDodo:
one thing I'm struggling with here is the inferred types shown in the IDE:

Screenshot 2023-04-22 at 17 09 23

Example:

```
const result = R.keys.strict({ foo: true, bar: false });
```

It's always

```
ObjectKey<{
    foo: boolean;
    bar: boolean;
}>
```

instead of just `'foo' | 'bar'`

this used to work better with the [previous types](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAKjgQwM5wEpwGZQiOAcigFMQSATZQgKBoGMIA7VeU1AVwBt4BeTAHQBrEgE9UA1lGD0YACgDeOCBABccGFA4kANHABGyKOuzIuqEnAC+ASgDcddtxgCQyMHIAecXgD44CjRwcAD0IcERwQB6APxBcKQwHFBMcJ4C9AAWRgCC8gAMNjS2QA)

src/keys.ts
TkDodo: I also see no error, using the built remeda version from this PR (codesandbox preview build) in a project that has `skipLibCheck: false`. I think this means we're good ...

eranhirsch:
I'll take a look at how type-fest's Simplify works, it's meant for these situations.

Another option is to split arrays and objects. Semantically calling indices `keys` is misleading, and the only reason this works is because it's a JS schtick that array's are actually objects. In that regard I don't mind dropping support for arrays altogether and pointing people to use `map.strict.indexed((_, index) => index)` to achieve this on arrays.

PR #306, opened 2023-04-20
ci: typecheck our output type definitions

eranhirsch:
We are not checking the right things right now.

Our sources could use internally any typescript feature we want and we can benefit from modern syntax.

Instead, we want to make sure that our type definition files (.d.ts) that we generate don't have typescript issues in them (like stripped type-checking comments), and that we support the typescript versions that we want as a product.

I've split our CI job to now do both.


tsconfig.dist.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "http://json.schemastore.org/tsconfig",
+
+  "include": ["dist"],
+
+  "compilerOptions": {
+    "strict": true,
```
TkDodo, line 7: should we not turn on more strict options, or extend our normal tsconfig ? Or does this not matter because it's unlikely that something like that will diverge from the source checks?

eranhirsch: I thought about it, but feels like it's not the right place for it, the only thing that truly matters at this point is that an older version of typescript can compile with these files. That's why it should be kept separate from our base tsconfig which is more concerned with typescript as a developer experience tool than a build tool.

Most of the strict flags deal with code constructs, and these are only type definitions (.d.ts). I tried turning on allowJs and checkJs and they add noise.

It'd be great if you can double check me on this, though

PR #313, opened 2023-04-27
feat(guards): more control over undefined and null checks

eranhirsch:
When working with stricter typescript null and undefined are not the same, and sometimes you'd want to filter a type `{} | null | undefined` so that you still retain the null or undefined values.

To prevent any backwards compatibility issues I created a new variant for isDefined for the explicit case, and added an `isNonNull` for nulls.

I also split the guard tests so that they are colocated with the files following the split done recently.
---


src/isDefined.ts
```diff
@@ -13,3 +13,9 @@
 export function isDefined<T>(data: T): data is NonNullable<T> {
   return typeof data !== 'undefined' && data !== null;
 }
+
+export namespace isDefined {
+  export function strict<T>(data: T | undefined): data is T {
```
TkDodo, line 18: we'd also need docs here

src/isNonNull.ts
```diff
@@ -0,0 +1,3 @@
+export function isNonNull<T>(data: T | null): data is T {
+  return data !== null;
+}
```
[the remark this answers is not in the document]
eranhirsch: oops, i had another commit i never pushed with the docs and index changes 😨

src/isDefined.ts
eranhirsch: do we document variants? I couldn't find an example where we do, the documentation usually resides on the main object

TkDodo: ah you're right, it's enough if we have an example 👍

TkDodo (approved):

Issue #307, opened 2023-04-21
Missing function: isEmpty


paour:
I've seen a few references to this function in issues, but it's neither implemented, nor documented in `mapping.md` and I was wondering why.

It's all the more important because checking for empty objects (in particular) is a fraught subject in JS. https://www.samanthaming.com/tidbits/94-how-to-check-if-object-is-empty/

TkDodo:
I don't think with TypeScript, you would need the constructor check. An `isEmpty` function for objects should only accept objects, so the keys-length-check should suffice, no? We could add that to the one-liner helpers?

paour:
Array and string overloads would be nice.

eranhirsch:
I commented on the PR but just mirroring what I said there:
1. With stronger typing overall in Remeda the value of these predicates is in narrowing types, so you'd want a isEmpty(x) to narrow the *type* of `x` to it's trivial empty value.
2. But narrowing an array to [], or a string to "", isn't very useful as you most likely not act on the empty array itself (you'd probably have some fallback logic instead).
3. This means that most likely what you actually want is a `isNonEmpty`, as that would ensure that, for example, `myArray[0]` is defined.
4. The problem is that typescript doesn't do implicit typing for *negated type predicates*, so simply doing `!isEmpty(x)` is meaningless.
5. And defining isNonEmpty on anything *but arrays* is impossible in typescript as there's no way to type a non-empty string, or an object with at least one prop in it (and it's especially hard if you actually want to refine the input, and not just return a generic non-empty value).

TL;DR - I don't think this is very useful as a type predicate, and as a simple boolean predicate this is just a trivial one liner.

paour:
Fixed by #314

PR #314, opened 2023-04-28
feat: add isEmpty util function

michal-weglarz:
The basic implementation of `isEmpty` util function returns true if the passed input is empty, otherwise false. The criteria of "emptiness" have been mostly influenced by `Ramda`'s [isEmpty](https://ramdajs.com/docs/#isEmpty).

The supported values have been included both in typedoc as examples and in test cases. 
```ts
 R.isEmpty([]) //=> true
 R.isEmpty({}) //=> true
 R.isEmpty('') //=> true
 R.isEmpty(new Set()) //=> true
 R.isEmpty(new Map()) //=> true
 R.isEmpty([1, 2, 3]) //=> false
 R.isEmpty({ length: 0 }) //=> false
 R.isEmpty(null) //=> false
 R.isEmpty(1) //=> false
 R.isEmpty(true) //=> false
```

I have tried to experiment with narrowing types like this: 
```ts
type EmptyType =
  | ''
  | Array<never>
  | Record<string, never>
  | Set<never>
  | Map<never, never>;

export function isEmpty(data: unknown): data is EmptyType {
...
```

but unfortunately I haven't found a consistent way to express the empty state for different values using types in Typescript. I'm open to any suggestions though and I'll be happy to learn if there's a way to solve the issue:

```ts
// ✅ – seems good
const input: Array<number> = [];
if (isEmpty(input)) {
  //=> never[]
} else {
  //=> number[]
}

// ✅ – seems good
const input: Record<string, unknown> = {};
if (isEmpty(input)) {
  //=> Record<string, never>
} else {
  //=> Record<string, unknown>
}

// 🤔
const input = { a: 1, b: 2 };
if (isEmpty(input)) {
  //=> {
  //      a: number;
  //      b: number;
  //   } & EmptyType
} else {
  // => {
  //      a: number;
  //      b: number;
  //   }
}
``` 

The absence of this function in `Remeda` has been pointed out, for example, in this issue: #307.

---

Make sure that you:


eranhirsch:
I think there's value in this, especially since it's present in other libraries we'd like people to migrate from comfortably, but... I think the value of narrowing the typing to empty isn't as useful as having a "isNonEmpty" function. The problem with a isNonEmpty function is that it's only really possible with arrays, neither strings nor objects have a notion (in the typing system) of being non-empty.

So basically my TL;DR is:
1. the type-narrowing for isEmpty is not important, might as well return false and that's it
2. I'd add a isNonEmpty specifically for arrays.

To comment on the typing attempt you mention in the description, you can't narrow with a union type because you would effectively be *widening* the type to include all possible union members. You'd need to make the type conditional on the type of the input, e.g. 
```ts
type Empty<T> = T extends readonly unknown[]
  ? Array<never>
  : T extends Record<PropertyKey, unknown>
  ? Record<keyof T, never>
  : T extends string
  ? ""
  : never
```
_(don't copy-paste me here, I didn't check it in an IDE/playground)_

Secondly, your examples are working by accident, the type isn't actually narrowed usually in the "else" part of a condition. In Typescript (currently at least, v5.0) negating a type predicate doesn't negate it's type. It only works for unions. e.g.
```ts
// would work!
function foo(x: "bar" | "baz"): void {
  if (x === "bar") {
    // x is "bar"
  } else {
    // x is "baz"
  }
}

function foo(x: string): void {
    if (startsWithAnH(x)) {
        // type is ok here
    } else {
        // type is just string, it doesn't "remember" that it doesn't start with an h.
        if (x[0] === "h") {
          // This condition is valid, typescript wouldn't consider this unreachable...
        }
    }
}

function startsWithAnH(x: string): x is `h${string}` {
    return x[0] === "h";
}
```

paour (approved):
Just minor typos.

eranhirsch:
>  but for an array, index access is always nullable anyway, so I'm not sure what it really adds.

This is incorrect, if you narrow an array to be non empty you can make it look like `[T, ...Array<T>]`, and accessing the array at index 0 will be strongly typed `T`. We already do this with `groupBy`, `chunk`, `first`, etc...

eranhirsch (changes requested):
Let's not accept values which can't semantically be empty.

The whole point of having a strongly typed language is to prevent runtime errors by moving checks to the compile time. Sending nulls, numbers, Dates, or booleans to an `isEmpty` function should be considered a bug. If there's a legitimate use-case to return either `false` or `true` for those values it should be explicit.

Remeda is not a drop-in replacement for lodash or ramda, and not designed like this in many of it's utility functions. It doesn't matter what those libraries do here, they are not canon.

paour:
> Let's not accept values which can't semantically be empty.

I wholeheartedly agree for other data types. For nullish, I do see a case for treating it as empty, for example in a chain. But I can easily live with the stricter approach.

src/isEmpty.ts
```diff
@@ -0,0 +1,35 @@
...
+ *    R.isEmpty(new Set()) //=> true
+ *    R.isEmpty(new Map()) //=> true
+ *    R.isEmpty([1, 2, 3]) //=> false
+ *    R.isEmpty({ length: 0 }) //=> false
+ *    R.isEmpty(null) //=> false
+ *    R.isEmpty(1) //=> false
+ *    R.isEmpty(true) //=> false
+ * @category Function
+ */
+export function isEmpty<T>(data: T) {
```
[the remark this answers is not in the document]
michal-weglarz: It's a fair point! I'll adjust the input accordingly.

src/isEmpty.test.ts
```diff
@@ -0,0 +1,71 @@
+import { isEmpty } from './isEmpty';
+
+describe('isEmpty', () => {
+  test('returns false when data array contains items', () => {
+    expect(isEmpty([1, 2, 3])).toBe(false);
+  });
+
+  test('returns false when data is an empty array', () => {
```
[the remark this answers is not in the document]
michal-weglarz: Fixed. Thanks for noticing this!

src/isEmpty.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    expect(isEmpty(new Map())).toBe(true);
+  });
+
+  test('returns false for a non-empty Set', () => {
+    const mapObject = new Map();
+    mapObject.set('a', 1);
+    mapObject.set('b', 2);
+
+    expect(isEmpty(mapObject)).toBe(false);
+  });
```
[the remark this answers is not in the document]
michal-weglarz: Changed.

src/isEmpty.ts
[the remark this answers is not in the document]
michal-weglarz: I have added the `Data` type alias for the union type specifying allowed inputs.

src/isEmpty.ts
```diff
@@ -0,0 +1,39 @@
...
+ *    R.isEmpty([]) //=> true
+ *    R.isEmpty({}) //=> true
+ *    R.isEmpty('') //=> true
+ *    R.isEmpty(new Set()) //=> true
+ *    R.isEmpty(new Map()) //=> true
+ *    R.isEmpty([1, 2, 3]) //=> false
+ *    R.isEmpty({ length: 0 }) //=> false
+ * @category Function
+ */
+export function isEmpty(data: Data): boolean {
```
eranhirsch, line 28: Thanks for handling the typing, I still think your original approach to narrow the type too is still valid here, as any data refinement is better than none. You can type the return value based on T so that you can "link" the input and the output.

With this kind of typing you don't need to type the input itself, because any other value (like number) would fall to the `never` final type, and the code wouldn't compile. (you can test this with @ts-expect-error above calls to `isEmpty(3)`)

```suggestion
type Empty<T> = T extends ReadonlySet<unknown>
  ? ReadonlySet<never>
  : T extends ReadonlyMap<infer Key, unknown>
  ? ReadonlyMap<Key, never>
  : T extends ArrayLike<unknown>
  ? ArrayLike<never>
  : T extends string
  ? ""
  : T extends Record<PropertyKey, unknown>
  ? Record<keyof T, never>
  : never; 

export function isEmpty<T>(data: T): data is Empty<T> {
```

eranhirsch (changes requested):
Let's make this the best type experience we can

src/isEmpty.ts
michal-weglarz: Hmm, I like the idea, but I'm not exactly sure how is it possible to link the input and the output, since, for `data is Empty<T>`  type predicate, Typescript complains that: 
```ts
A type predicate's type must be assignable to its parameter's type.
  Type 'Empty<T>' is not assignable to type 'T'.
    'T' could be instantiated with an arbitrary type which could be unrelated to 'Empty<T>'.
```

michal-weglarz: I believe the below code works as intended, according to what you have described. 

Unfortunately, I'm not using `readonly` types here – I still need to work on them a little bit to avoid them being narrowed down to `never` when the `data` argument is empty, instead of, for example, `ReadonlySet<never>`.

```ts
type Empty<T> = T extends Map<infer Key, unknown>
  ? Map<Key, never>
  : T extends Set<unknown>
  ? Set<never>
  : T extends string
  ? ''
  : T extends Array<unknown>
  ? Array<never>
  : T extends Record<PropertyKey, unknown>
  ? Record<keyof T, never>
  : never;

export function isEmpty<T>(
  data: Empty<T> extends never ? never : T
): data is Empty<T> extends never
  ? never
  : Empty<T> extends T
  ? Empty<T>
  : never {
  if (data instanceof Set || data instanceof Map) {
    return data.size === 0;
  }
  if (isArray(data) || isString(data)) {
    return data.length === 0;
  }
  if (isObject(data)) {
    return Object.keys(data).length === 0;
  }
  return false;
}

```

Tests of narrowing:
```ts
// ✅
let input = 'test';
if (isEmpty(input)) {
  input; //=> ""
} else {
  input; //=> string
}

// ✅
const input1: Array<number> = [];
if (isEmpty(input1)) {
  input1; //=> never[]
} else {
  input1; //=> number[]
}

// ✅
const input2: Record<string, unknown> = {};
if (isEmpty(input2)) {
  input2; //=> Record<string, never>
} else {
  input2; //=> Record<string, unknown>
}

// ✅
const input3 = { a: 1, b: 2 };
if (isEmpty(input3)) {
  input3; //=> Record<"a" | "b", never>
} else {
  input3; // => {
  //      a: number;
  //      b: number;
  //   }
}

// ✅
const input4 = new Set(['a', 2]);
if (isEmpty(input4)) {
  input4; // => Set<never>
} else {
  input4; // => Set<string | number>
}

// ✅
const input5 = 23;
isEmpty(input5); // => Error: Argument of type number is not assignable to parameter of type never.
```

eranhirsch: Sorry for driving you crazy with this, working on Remeda is always a good way to practice typescript :)

I'm fine with dropping the readonly parts, maybe except the readonly Array which I know is very predictable.

Another approach is to simply use function overloading here, it might just be easier to read and maybe work better. Typescript would do the complex union typing for us via the overload.

Lastly, let's make arrays return `[]` and not `Array<never>`, i'm concerned that `Array<never>` can still be a sparse array with length !== 0, `[]` is fully unambiguous.

PS, i'm not sure that type-narrowing Sets and Maps is useful anyway, i'm not sure they have a good semantic of an empty collection as they would suffer from the same problem I mentioned with the array. I'm also not sure they are used frequently enough that we should care right now. We can always fix bugs that get reported later on.

```ts
export function isEmpty(data: string): data is "";
export function isEmpty(data: ReadonlyArray<unknown> | []): data is [];
export function isEmpty<Key extends PropertyKey>(data: Readonly<Record<Key, unknown>>): data is Record<Key, never>;
export function isEmpty(data: ReadonlySet<unknown>): data is Set<never>;
export function isEmpty<Key>)(data: ReadonlyMap<Key, unknown>): data is Map<Key, never>;
export function isEmpty(data: unknown): boolean {
  // IMPL
}
```

michal-weglarz: I've already learned a ton, so I really appreciate it! :) 

I pushed the changes that include function overloading, just as you've suggested and it make the code much more readable, that's for sure. I have only slightly changed a signature of `export function isEmpty<Key extends PropertyKey>(data: Readonly<Record<Key, unknown>>): data is Record<Key, never>`, because it made the function accept invalid input without an error. 
I have also removed type-narrowing for Sets and Maps – I agree with you that they might not be that relevant but if they turn out to be so, then we can always fix that later.

TkDodo (approved):

PR #315, opened 2023-05-04
Add switchCase function

aose-yuu:
## Description:

This pull request introduces a new function `switchCase` which provides a more functional approach to using switch statements in JavaScript/TypeScript. The function is designed to work seamlessly with Remeda's existing functions, supporting both data-first and data-last styles.

## Implementation details:

- The `switchCase` function has been implemented to handle switch-like cases in a functional style.
- Both data-first and data-last styles are supported, making it compatible with other Remeda library functions.
- TypeScript documentation (tsdoc) has been added for the switchCase function.
- Comprehensive test cases have been added for various scenarios, ensuring the function works as expected in different use cases.

## Example usage:

```typescript
import * as R from 'remeda';

const result1 = R.switchCase(
  1,
  {
    1: () => 'first',
    5: () => 'last',
  },
  () => 'not exists'
);
console.log(result1); // 'first'

const result2 = R.switchCase(
  {
    1: () => 'first',
    5: () => 'last',
  },
  () => 'not exists'
)(5);
console.log(result2); // 'last'

const result3 = R.pipe(
  R.range(1, 6),
  R.last(),
  R.switchCase(
    {
      1: () => 'first',
      5: () => 'last',
    },
    () => 'not exists'
  )
);
console.log(result3); // 'last'
```

By incorporating switchCase with other Remeda functions like R.range(), R.first(), and R.last(), this new function enables users to write cleaner and more functional code.

Please review the changes and let me know if any modifications are required or if you have any suggestions for improvement. Looking forward to your feedback!

---

Make sure that you:


src/switchCase.ts
```diff
@@ -0,0 +1,81 @@
...
+  cases: Record<string, () => R>,
+  defaultFn: () => R
+): (value: T) => R;
+
+export function switchCase() {
+  if (arguments.length === 2) {
+    return purry(_switchCaseWithCases, arguments);
+  } else {
+    return purry(_switchCase, arguments);
+  }
```
eranhirsch, line 80: This is what purry does internally, it looks at the arguments and dynamically builds the data-last function from the data-first one. You don't need any of this.
```suggestion
  return purry(_switchCase, arguments);
```

eranhirsch (changes requested):
This could be better typed, the cases object takes a simple record keyed on strings no matter what T is. This means that the cases object would not be able to enforce it is a valid mapping of the input values in cases like literal unions ("cat" | "dog") and would result in silently returning the default value. This also prevents any exhaustive checks from being performed on the cases object, reducing a lot of the value of using switches in the first place.

I'm also not sure about the API in general, and think people would encounter issues trying to use this in real-world scenarios. One common use of switch is further processing on discriminated unions:
```ts
function foo(animal: { type: "cat", color: string } | { type: "dog", name: string }): string {
  switch (animal.type) {
    case "cat": return `I'm a ${animal.color} cat`;
    case "dog": return `I'm ${animal.name} the dog`;
  }
}
```

So to be useful this would need to take a prop that would define the pivot, and the case functions would need to take the input object. I'm not sure how easy it would be to type that...

Finally, I have a nit about the name, I think this should be called just switch, the "case" part is an implementation detail.

eranhirsch:
@aose-yuu I'm closing this PR as it's been stale for almost 3 months. If you still want to pursue this feel free to reopen the PR and address the comments

Issue #41, opened 2020-01-26
add mergeDeep


geekflyer:
Hi,

great work with this library!

I was wondering if you could add a mergeDeep (mergeDeepLeft or mergeDeepRight) to the library? That is especially useful to merge defaults into nested config objects etc.

Thanks

lstkz:
Hi,

Sorry for the delay. Would you be interested in submitting a PR?

lstkz:
Here is a pending PR https://github.com/remeda/remeda/pull/69

hgezim:
Any movement on this?

TkDodo:
there was no movement, so I closed the PR. if you want to pick it up - please go ahead :)

yanick:
Fwiw, I ported `updeep`, a small utility package that focuses on deep merging of immutable objects to use Remeda as its underlying engine: https://github.com/yanick/updeep-remeda/

eranhirsch:
Being handled in #317

PR #317, opened 2023-05-13
feat(merge): add mergeDeep

sairus2k:
This pull request continues the work initiated in PR #69. It includes the implementation of a 'mergedeep' feature, which is designed to deeply merge JavaScript objects.

Key updates in this pull request include:

1.    Rebased the branch on the current master to ensure compatibility and ease of merge.
2.   Fixed all lint errors to maintain code quality and readability.
3.   Added mergedeep into index.ts and mapping.md

Please review these changes and provide your valuable feedback.

solves: #41

---

Make sure that you:


davydkov:
Any ideas when this gonna be merged?

src/index.ts
```diff
@@ -8,6 +8,7 @@ export * from './compact';
 export * from './concat';
 export * from './countBy';
 export * from './createPipe';
+export * from './deepMege';
```
TkDodo, line 11: There's a typo here: `deepMege` vs `deepMerge`.

Also, to allow importModularization (e.g. in [nextJs](https://nextjs.org/docs/architecture/nextjs-compiler#modularize-imports)), it is necessary that each re-export is done from a separate file with the exact same name, so `deepMergeLeft` and `deepMergeRight` would need to be split apart please.

TkDodo:
@eranhirsch can you also take a look please?

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+  // If we here we know that a is primitive value if it's defined we chose a over b
+  if (isDefined(a)) {
+    return a;
+  }
+
+  return b;
+}
+
+/**
+ * Merging object from left to right
```
eranhirsch, line 92: Let's keep it simple, why do we need to handle multiple merges here, let's simply provide a deepMerge function that only handles 2 objects, taking multiple objects in a single merge is an optimization that doesn't make this code easier to read or test.

eranhirsch (changes requested):
This looks really interesting, but it's also really confusing, and the code doesn't make it any simpler. I have a feeling that there are edge-cases that I both don't understand, and don't work as expected.

You don't use typing to protect against making mistakes, and instead use `unknown` and `any` in the most sensitive parts of the impl. The tests are over generic and large, and don't help to bisect the logic properly.

sairus2k:
@eranhirsch Thank you for your review and constructive feedback.

I take your feedback to heart and will work on improving the readability of the function and ensuring that it handles all edge cases correctly.

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+      }
+
+      // If 'obj' is an array, check if any of its elements are
+      // an object that has been seen already.
+    } else if (Array.isArray(obj)) {
+      for (const item of obj) {
+        if (detectCycle(item)) {
+          return true;
+        }
+      }
```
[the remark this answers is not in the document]
sairus2k: done

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+  : T extends Array<infer U>
+  ? DeepPartialArray<U>
+  : T extends object
+  ? DeepPartialObject<T>
+  : T | undefined;
+export type DeepPartialObject<T> = { [P in keyof T]?: DeepPartial<T[P]> };
+
+function isCyclic(object: unknown) {
+  const seenObjects = new WeakMap(); // use to keep track of which objects have been seen.
+  function detectCycle(obj: unknown) {
```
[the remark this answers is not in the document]
sairus2k: done

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+      Object.prototype.toString.call(obj) == '[object Object]'
+    ) {
+      if (seenObjects.has(obj)) {
+        return true;
+      }
+
+      // If 'obj' hasn't been seen, add it to 'seenObjects'.
+      // Since 'obj' is used as a key, the value of 'seenObjects[obj]'
+      // is irrelevant and can be set as literally anything you want. I
+      // just went with 'undefined'.
```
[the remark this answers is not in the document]
sairus2k: Replaced WeakMap with WeakSet

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+export type DeepPartialObject<T> = { [P in keyof T]?: DeepPartial<T[P]> };
+
+function isCyclic(object: unknown) {
+  const seenObjects = new WeakMap(); // use to keep track of which objects have been seen.
+  function detectCycle(obj: unknown) {
+    // If 'obj' is an actual object (i.e., has the form of '{}'), check
+    // if it's been seen already.
+    if (
+      isObject(obj) &&
+      Object.prototype.toString.call(obj) == '[object Object]'
```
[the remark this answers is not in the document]
sairus2k: Removed this code

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    return false;
+  }
+
+  return detectCycle(object);
+}
+
+/**
+ * Merging @param a and @param b recursively @param a is most important
```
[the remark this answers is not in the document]
sairus2k: It means that "a" is dominant parameter. Rewrote jsdoc.

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+ */
+function recursiveMerge(a: unknown, b: unknown): unknown {
```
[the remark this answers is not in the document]
sairus2k: done

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+  if (isObject(a) && isObject(b)) {
+    const output = { ...a };
+    const keys = uniq([...Object.keys(a), ...Object.keys(b)]);
+    for (const k of keys) {
+      // If object cyclic I will not try to merge objects
+      if (isCyclic(a[k])) {
+        output[k] = a[k];
+        continue;
+      }
+      output[k] = recursiveMerge(a[k], b[k]);
```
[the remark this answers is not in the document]
sairus2k: fixed

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+import { isArray } from './isArray';
+import { isObject } from './isObject';
+import { uniq } from './uniq';
+import { clone } from './clone';
+
+export type DeepPartialArray<T> = Array<DeepPartial<T>>;
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type DeepPartial<T> = T extends Function
+  ? T
+  : T extends Array<infer U>
```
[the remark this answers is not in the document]
sairus2k: done

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+  ? DeepPartialArray<U>
+  : T extends object
+  ? DeepPartialObject<T>
+  : T | undefined;
```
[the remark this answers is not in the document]
sairus2k: Replaced with `unknown`

src/deepMege.ts
sairus2k: done

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+ * (truthy plain value) + ob = (truthy plain value)
+ * (truthy plain value) + undefined = (truthy plain value)
+ * A(truthy plain value) + B(truthy plain value) = A(truthy plain value)
+ *
+ * Handles circular references
+ * @returns {T}
+ */
+export function deepMergeLeft<T extends object>(
+  target: T,
+  ...sources: Array<DeepPartialObject<T>>
```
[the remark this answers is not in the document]
sairus2k: done

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+): T {
+  let output = { ...target } as unknown;
+  for (const source of sources) {
+    output = recursiveMerge(output, source);
+  }
+  return output as any;
```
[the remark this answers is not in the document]
sairus2k: Removed type casting

src/deepMege.ts
```diff
@@ -0,0 +1,149 @@
...
+ * (truthy plain value) + undefined = (truthy plain value)
+ * A(truthy plain value) + B(truthy plain value) = B(truthy plain value)
+ * Handles circular references
+ * @returns {T}
+ */
+export function deepMergeRight<T extends object>(
+  target: T,
+  ...sources: Array<DeepPartialObject<T>>
+): T {
+  let output = clone(target) as unknown;
```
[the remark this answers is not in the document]
sairus2k: Because test for deepMergeLeft failed for some reason if we use clone there. But I think that we do not need deep clone here because we already have shallow clones in recursiveMerge for each nested object.

src/deepMerge.test.ts
```diff
@@ -0,0 +1,315 @@
+import { deepMergeLeft, deepMergeRight } from './deepMege';
```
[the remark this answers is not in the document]
sairus2k: Simplified tests a lot. Most of test cases moved to `_recursiveMerge.test.ts`

src/deepMerge.test.ts
```diff
@@ -0,0 +1,315 @@
...
+      d: {
+        e: {
+          a?: number; // C3 nested object
+        };
+        d?: () => void; // C4 function
+      };
+      y?: {
+        a: number;
+      };
+    }
```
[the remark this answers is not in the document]
sairus2k: done

src/deepMerge.test.ts
```diff
@@ -0,0 +1,315 @@
...
+  };
+  test('Expecting correctly merged arrays', () => {
+    expect(deepMergeLeft(d1, d2)).toEqual({
+      a: [1, 2, 3, { a: 1 }, 7, { a: 1 }, { a: 2 }],
+    });
+
+    expect(deepMergeRight(d1, d2)).toEqual({
+      a: [7, 2, 3, { a: 1 }, { a: 2 }, 1, { a: 1 }],
+    });
+  });
```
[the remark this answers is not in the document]
sairus2k: Made it more consistent

src/index.ts
sairus2k: Fixed typo and split into two files

sairus2k (commented):
I made all requested changes. Please take a look. @TkDodo @eranhirsch

src/_recursiveMerge.ts
```diff
@@ -0,0 +1,44 @@
...
+      return uniq(a.concat(b)) as T;
+    }
+    return a;
+  }
+  const output = { ...a };
+  const keys = uniq(Object.keys(a).concat(Object.keys(b)));
+  for (const k of keys) {
+    // @ts-expect-error We've already ensured that "a" and "b" are objects at runtime.
+    if (!isCyclic(a[k])) output[k] = recursiveMerge(a[k], b[k]);
+  }
```
eranhirsch, line 42: I think you are over-complicating what you need to do here, this is also the reason why the typing becomes harder. The type for recursiveMerge should be:
```ts
recursiveMerge(a: object, b: object): T {
```

This ensures that you are only dealing with merging for objects, anything else (`undefined`, strings, numbers, booleans, etc...) simply won't be callable.

Now, you simply do the merge of objects like this:
```suggestion
  // At this point the output is already merged, simply not deeply merged.
  const output = { ...a, ...b };
  // now just scan the output and look for values that should have been deep-merged
  for (const k in b) {
    const aValue = a[k];
    const bValue = b[k];
    if (
      typeof aValue === "object" &&
      !isCyclic(a[k]) &&
      typeof bValue === "object"
    ) {
      // These are the only keys that need recursive merging. At this point
      // we already know that both of them are objects, so they match the type
      // of recursiveMerge.
      output[k] = resursiveMerge(aValue, bValue);
    }
  }
```

src/deepMergeRight.ts
```diff
@@ -0,0 +1,25 @@
...
+ * A(truthy plain value) + B(truthy plain value) = B(truthy plain value)
+ * Handles circular references
+ * @returns {T}
+ */
+export function deepMergeRight<T extends object>(
+  target: DeepPartialObject<T>,
+  source: T
+): T {
+  return recursiveMerge(source, target);
+}
```
eranhirsch, line 25: Let's drop this, it's not that important to have a merge for both sides, and if someone really needs it they can always create a simple lambda that would offer it.

src/deepMergeLeft.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Handles circular references
+ * @returns {T}
+ */
+export function deepMergeLeft<T extends object>(
+  target: T,
+  source: DeepPartialObject<T>
+): T {
+  return recursiveMerge(target, source);
+}
```
eranhirsch, line 27: This simply calls the recursiveMerge function, then why do we need it? simply use the recursiveMerge function. I would simply named it deepMerge and drop the left/right concept. It's enough to have one direction, and anyone who needs the other direction can simply call the function with the params reversed.

Also, please follow the format we have for other files and create the data-first and data-last signatures and use purry to call the implementation

eranhirsch (changes requested):
This is a great improvement, but there's still a few things that I would want to see changed.

* The semantics of merging need to be really well defined, otherwise users would run into odd situations. This is especially important for array merging.
* Some typing issues.
* The whole concept of the left/right merge is redundant, we only need one mergeDeep function, users can build more complex logic on top of it, and it just makes using the function more cumbersome because users would now need to understand what left and right mean.
* Make sure you implement both the data-first and data-last signatures so users can use this in a pipe

dartess:
@sairus2k Hi! How are you doing with that? Do you need any help? I am looking forward to this PR to complete the migration and can help if needed.

src/_recursiveMerge.ts
```diff
@@ -0,0 +1,44 @@
...
+export type DeepPartialArray<T> = ReadonlyArray<DeepPartial<T>>;
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type DeepPartial<T> = T extends Function
+  ? T
+  : T extends ReadonlyArray<infer U>
+  ? DeepPartialArray<U>
+  : T extends object
+  ? DeepPartialObject<T>
+  : unknown;
+export type DeepPartialObject<T> = { [P in keyof T]?: DeepPartial<T[P]> };
```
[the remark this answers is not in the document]
sairus2k: 1. I think, we should use merge-deep, not partial-deep from type-fest. I added type-fest as a dependency, because merge-deep is too complex
2. It was used in two files. Now it is removed

src/_recursiveMerge.ts
```diff
@@ -0,0 +1,44 @@
...
+    typeof a !== 'object' ||
+    typeof b !== 'object' ||
+    a === null ||
+    b === null
+  ) {
+    return a ?? (b as T);
+  }
+  if (Array.isArray(a)) {
+    if (Array.isArray(b)) {
+      return uniq(a.concat(b)) as T;
```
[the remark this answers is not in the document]
sairus2k: done

src/deepMergeRight.ts
sairus2k: dropped

src/deepMergeLeft.ts
sairus2k: done

src/deepMerge.ts
```diff
@@ -0,0 +1,68 @@
...
+  ) {
+    return a ?? b;
+  }
+  if (Array.isArray(a)) {
+    if (Array.isArray(b) && !isCyclic(a)) {
+      return a.concat(b);
+    }
+    return a;
+  }
+  const output = { ...a };
```
dartess, line 61: when is it really necessary? it seems that the values for all keys will be overwritten during the loop. I think you can go the other way:

1. easy way: use empty object for output.
```ts
const output = {};
```

2. another way: copy all values through the spread, and let only the shared keys go into the loop.
```ts
const output = { ...a, ...b }; // order is not important 
const keys = intersection(Object.keys(a), Object.keys(b));
for (const k of keys) {
  // @ts-expect-error We've already ensured that "a" and "b" are objects at runtime.
  if (!isCyclic(a[k])) output[k] = deepMerge(a[k], b[k]);
}
return output;
```

3. yet another way: skip keys that are not in `b`: this means that they are only in `a`, which means they are already in the output:

```ts
for (const k of keys) {
  if (!(k in b)) {
    continue;
  }
  // @ts-expect-error We've already ensured that "a" and "b" are objects at runtime.
  if (!isCyclic(a[k])) output[k] = deepMerge(a[k], b[k]);
}
```

in the second and third options, we will save on the number of calls to the `deepMerge` function

src/deepMerge.ts
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Recursively merges two values, `a` and `b`.
+ * @param source - The second value to merge.
+ * @signature
+ *    R.deepMerge(b)(a)
+ * @example
+ *    R.deepMerge({ foo: 'baz', y: 2 })({ foo: 'bar', x: 1 }) // => { foo: 'bar', x: 1, y: 2 }
+ *    R.pipe(
+ *      { foo: 'bar', x: 1 },
+ *      R.difference({ foo: 'baz', y: 2 }),
```
dartess, line 32: `R.deepMerge`

mapping.md
```diff
@@ -17,6 +17,8 @@ documentation when migrating._
 | `compact`          | `compact`          | `-`                 |
 | `concat`           | `concat`           | `concat`            |
 | `createPipe`       | `-`                | `-`                 |
+| `deepMergeLeft`    | `-`                | `mergeDeepLeft`     |
```
dartess, line 20: `deepMerge`

src/deepMerge.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(deepMerge('foo', 'bar')).toBe('foo');
+    expect(deepMerge('foo', undefined)).toBe('foo');
+    expect(deepMerge(undefined, 'bar')).toBe('bar');
+  });
+
+  test('should merge objects', () => {
+    const a = { foo: 'bar', x: 1 };
+    const b = { foo: 'baz', y: 2 };
+    expect(deepMerge(a, b)).toEqual({ foo: 'bar', x: 1, y: 2 });
+  });
```
dartess, line 27: It seems that all tests should pass the usual "merge". Maybe add tests with nested objects?

src/deepMerge.ts
```diff
@@ -0,0 +1,68 @@
...
+    typeof a !== 'object' ||
+    typeof b !== 'object' ||
+    a === null ||
+    b === null
+  ) {
+    return a ?? b;
+  }
+  if (Array.isArray(a)) {
+    if (Array.isArray(b) && !isCyclic(a)) {
+      return a.concat(b);
```
dartess, line 57: I think it's worth adding tests for merge arrays. tests should demonstrate the expected behavior.

mapping.md
dartess, line 20: `_.merge` from lodash works recursively

src/deepMerge.test.ts
sairus2k: Added case for nested objects

src/deepMerge.ts
sairus2k: done

src/_recursiveMerge.ts
```diff
@@ -0,0 +1,44 @@
+import { uniq } from './uniq';
+import { isCyclic } from './_isCyclic';
+
+export type DeepPartialArray<T> = ReadonlyArray<DeepPartial<T>>;
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type DeepPartial<T> = T extends Function
+  ? T
+  : T extends ReadonlyArray<infer U>
+  ? DeepPartialArray<U>
```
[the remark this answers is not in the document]
sairus2k: It was replaced with type from `type-fest`

src/_recursiveMerge.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export function recursiveMerge<T>(a: T, b: unknown): T {
+  if (
+    typeof a !== 'object' ||
+    typeof b !== 'object' ||
+    a === null ||
+    b === null
+  ) {
+    return a ?? (b as T);
+  }
```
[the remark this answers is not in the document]
sairus2k: I think it is fixed with `type-fest` typings

src/_isCyclic.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    // If 'input' hasn't been seen, add it to 'seenValues'.
+    seenValues.add(input);
+
+    if (Array.isArray(input)) {
+      // If 'input' is an array, check if any of its elements are an object that has been seen already.
+      return input.some(hasCycle);
+    } else {
+      // Recurse through the object, looking for more circular references.
+      return Object.values(input).some(hasCycle);
```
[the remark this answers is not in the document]
sairus2k: Added support for Map and Set. Also added some more test cases

src/_recursiveMerge.ts
sairus2k: done

src/deepMerge.ts
sairus2k: Chose the second option

src/_isCyclic.ts
```diff
@@ -0,0 +1,53 @@
...
+ * ```
+ */
+export function isCyclic(value: unknown): boolean {
+  const seenValues = new WeakSet(); // use to keep track of which objects have been seen.
+  let depth = 0;
+
+  function hasCycle(input: unknown) {
+    depth++;
+
+    // If max depth is exceeded, return false
```
dartess, line 20: it won't work as intended. the exit will not be from the root call, in whose scope the variable is defined, but from one of the recursive calls. Thus, we will simply stop one of the iterations without stopping the function as a whole, and continue to count again in order to skip another one after 1000 iterations.

Demo: https://codesandbox.io/s/xenodochial-gagarin-hynlcd?file=/src/index.ts

dartess: It can be solved with try/catch and custom exception:

```js
const maxDepthReached = Symbol("maxDepthReached"); // on top level

...

    if (depth > MAX_DEPTH) {
      throw maxDepthReached;
    }

...

  try {
    return hasCycle(value);
  } catch (exception) {
    if (exception === maxDepthReached) {
      return false;
    }
    throw exception;
  }
```

demo: https://codesandbox.io/s/pedantic-phoebe-5xqcxc?file=/src/index.ts

dartess: Instead of a symbol, you can use another way to define your exception, for example, extend the error class and check instancesof. The variant with the symbol looks simpler.

src/_isCyclic.ts
```diff
@@ -0,0 +1,53 @@
+const MAX_DEPTH = 1000;
```
dartess, line 1: By the way, why is this function / this limit needed?

Given the limit, this function is not "fair". If the limit is exceeded, the function will say that the object is not cyclic, although it can be cyclic. Why not remove the limit?

If we are worried that the cycle check might take a long time because the check is called in a dipmerge, there is a simple solution - do not call the cycle check. Instead of pre-checking, add a runtime check that will check nodes as they are copied, and will respond to an attempt to process circular references. Then the check will become part of the implementation and will not be needed as a separate function.

src/deepMerge.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  // At this point the output is already merged, simply not deeply merged.
+  const output = { ...b, ...a };
+
+  // now just scan the output and look for values that should have been deep-merged
+  for (const k in b) {
+    const aValue = a[k];
+    const bValue = b[k];
+
+    if (isRecordOrArray(aValue) && !isCyclic(a[k]) && isRecordOrArray(bValue)) {
```
dartess, line 82: Another point that speaks not in favor of the current implementation.
We check `a[k]` for cyclicity. We obviously do it recursively. Then we recursively pass `a[k]` to `_deepMerge` and internally check the child nodes of `a[k]` for cyclicity here. But they are already checked in the root call! Thus, we have tons of extra checks for recursive calls.

src/deepMerge.ts
```diff
@@ -0,0 +1,90 @@
...
+      return a.concat(b);
+    }
+    return a;
+  }
+  if (Array.isArray(b)) {
+    return a;
+  }
+  if (Array.isArray(b)) {
+    return a;
+  }
```
dartess, line 72: code is duplicated, typo or mistake

src/_isCyclic.ts
sairus2k: Removed depth checking

src/_isCyclic.ts
sairus2k: Removed this

src/deepMerge.ts
sairus2k: Moved logic that detects cycles right into deepMerge function

src/deepMerge.ts
sairus2k: removed

src/deepMerge.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export function deepMerge() {
+  return purry(_deepMerge, arguments);
+}
+
+export function _deepMerge(a: UnknownRecordOrArray, b: UnknownRecordOrArray) {
+  const seenValues = new WeakSet(); // use to keep track of which objects have been seen.
+
+  function isArrayCyclic(arr: Array<unknown>) {
+    return arr.some(value => value && seenValues.has(value));
```
eranhirsch, line 55: Why does the flasy check matter here? the only thing that matters is what we put in the `seenValues` set, if we don't put falsy values in it it doesn't matter.
```suggestion
    return arr.some(value => seenValues.has(value));
```

src/deepMerge.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  function merge(a: UnknownRecordOrArray, b: UnknownRecordOrArray) {
+    if (
+      typeof a !== 'object' ||
+      typeof b !== 'object' ||
```
eranhirsch, line 61: These are always false if a and b are only records and arrays.
```suggestion
```

src/deepMerge.ts
```diff
@@ -0,0 +1,112 @@
...
+    seenValues.add(b);
+
+    if (Array.isArray(a)) {
+      if (Array.isArray(b)) {
+        if (isArrayCyclic(a)) {
+          return b;
+        }
+        if (isArrayCyclic(b)) {
+          return a;
+        }
```
eranhirsch, line 81: ~~Can you add documentation here, I'm not sure I understand why if one array contains a recursive element then the merge result is simply the other array...~~

EDIT: Thinking about this again, because we don't deep merge the array items I think it's best to simply not do anything with recursive array items, as the merge logic wouldn't run into an infinite loop so it doesn't really matter.
```suggestion
```

src/deepMerge.ts
```diff
@@ -0,0 +1,112 @@
...
+    const output = { ...b, ...a };
+
+    // now just scan the output and look for values that should have been deep-merged
+    for (const k in b) {
+      const aValue = a[k];
+      const bValue = b[k];
+
+      if (
+        isRecordOrArray(aValue) &&
+        !seenValues.has(a[k] as object) &&
```
eranhirsch, line 100: why is the cast needed here?
```suggestion
        !seenValues.has(a[k]) &&
```

eranhirsch (approved):
The logic seems good to me, and the implementation is solid, I still have some nits and concerns I would like to see addressed, but I'm approving this so that this PR has the right momentum to get shipped.

I'll be honest, the cyclical checks concern me, I'm not sure it's an important facet of deep merging to even handle, and it *feels* error-prone, especially around how they are handled in the arrays. I think it's worthwhile simplifying the whole thing and just dropping this concept. If someone has a cyclic/recursive data structure and they want to deep-merge it they need to make a lot of decisions on what it means to their data-structure anyway, and they'd be better off building the merge logic themselves. This is an open source project, they can copy this implementation into their project and expand it with the cyclic/recursive logic they need.

src/deepMerge.ts
```diff
@@ -0,0 +1,112 @@
...
+ * @example
+ *    R.deepMerge({ foo: 'bar', x: 1 }, { foo: 'baz', y: 2 }) // => { foo: 'bar', x: 1, y: 2 }
+ * @data_first
+ * @category Object
+ * @pipeable
+ */
+export function deepMerge<
+  Target extends UnknownRecordOrArray,
+  Source extends UnknownRecordOrArray
+>(target: Target, source: Source): MergeDeep<Target, Source>;
```
eranhirsch, line 25: I just noticed, you need to add `{arrayMergeMode: 'spread'}` to the type to cover the concat logic you put in. This also needs to be added in the data-last signature. 
```suggestion
>(target: Target, source: Source): MergeDeep<Target, Source, { arrayMergeMode: "spread" }>;
```

src/deepMerge.ts
sairus2k: We need falsy check here because of ts error:
```
TS2345: Argument of type  unknown  is not assignable to parameter of type  object 
```

src/deepMerge.ts
sairus2k: Another ts error:
```
TS2345: Argument of type  unknown  is not assignable to parameter of type  object 
```

src/deepMerge.ts
```diff
@@ -0,0 +1,112 @@
...
+ * @param source - The second value to merge.
+ * @signature
+ *    R.deepMerge(a, b)
+ * @example
+ *    R.deepMerge({ foo: 'bar', x: 1 }, { foo: 'baz', y: 2 }) // => { foo: 'bar', x: 1, y: 2 }
+ * @data_first
+ * @category Object
+ * @pipeable
+ */
+export function deepMerge<
```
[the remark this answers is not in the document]
sairus2k: Renamed to `mergeDeep`

src/deepMerge.ts
sairus2k: As for now, without those check the function throws `RangeError: Maximum call stack size exceeded` for tests "should not merge left/right cyclic array"

src/mergeDeep.ts
```diff
@@ -0,0 +1,117 @@
...
+ *      R.mergeDeep({ foo: 'baz', y: 2 }),
+ *    ) // => { foo: 'bar', x: 1, y: 2 }
+ * @data_last
+ * @category Object
+ * @pipeable
+ */
+export function mergeDeep<
+  Target extends UnknownRecordOrArray,
+  Source extends UnknownRecordOrArray
+>(source: Source): (target: Target) => MergeDeep<Target, Source>;
```
eranhirsch, line 55:
```suggestion
>(source: Source): (target: Target) => MergeDeep<Target, Source, { arrayMergeMode: 'spread' }>;
```

src/mergeDeep.ts
```diff
@@ -0,0 +1,117 @@
...
+      if (Array.isArray(b)) {
+        if (isArrayCyclic(a)) {
+          return b;
+        }
+        if (isArrayCyclic(b)) {
+          return a;
+        }
+        return a.concat(b);
+      }
+      return a;
```
eranhirsch, line 89: This is incorrect in regards to the type you defined, when there's a mismatch between one side being an array and the other being something else, the latter value is always used, never the former: https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgWQKZQOaoCKtWOAXzgDMoIQ4AiGATzFQFoTUBnGKgbgChu6G4ADTgBeFOiy58AHm4B6OXCUA9APzclSAIYAuOACMIEADaotAOyIAaDYji645gK4h96ANoBda7e1QoWrRomKjIEAAmqHpUrGBQZuFURNwAfDx89KhwAJqi4iFSYLIKSnBqvvZ67FDA5hhePpqVBkamFo12Wv6BwVhhkdGx8VqJyWm8-FkAWgCMeb04eEXyiirqTQ7VtfXehDYbes6uHrv7nd1BEqERUdRDCUmEqemTcFMATPNXhcWrZeudKowGp1Bp7CoOI5uKBgs5+AKXEL9W5UeJgYxaADGqEezwmmTeAGYvgUlr9SuUDnAtqDThDDi5oWC8UA

```suggestion
      return b;
```

src/mergeDeep.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    if (Array.isArray(b)) {
+      return a;
```
eranhirsch, line 92: Similarly, this is also incorrect
```suggestion
      return b;
```

src/mergeDeep.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    // At this point the output is already merged, simply not deeply merged.
+    const output = { ...b, ...a };
```
eranhirsch, line 96: This is the opposite of how the type works, latter values are always preferred over former ones
```suggestion
    // At this point the output is already merged, simply not deeply merged.
    const output = { ...a, ...b };
```

src/mergeDeep.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    // now just scan the output and look for values that should have been deep-merged
+    for (const k in b) {
+      const aValue = a[k];
+      const bValue = b[k];
+
+      if (
+        isRecordOrArray(aValue) &&
+        !seenValues.has(a[k] as object) &&
+        isRecordOrArray(bValue)
```
eranhirsch, line 106: We never recurse into arrays, this should only check for records as those are the only type that we actually perform a deep merge for. I'd even go as far as limit the whole logic of this method to just objects without any special handling of arrays (beyond spreading them if both values are arrays).
```suggestion
        isRecord(aValue) &&
        !seenValues.has(a[k] as object) &&
        isRecord(bValue)
```

eranhirsch (changes requested):
I'm sorry but the function isn't implemented correctly in regards to the type it says it returns. Make sure you test not just the implementation but also if the typing fits the runtime results. You can use vitest's `expectTypeOf`

src/mergeDeep.ts
sairus2k: fixed

src/mergeDeep.ts
sairus2k: fixed

RedHatter:
@sairus2k Are you still working on this? Would really like to see this merged.

sairus2k:
@RedHatter I am no longer actively working on this feature.
Since you're interested in seeing this get merged, feel free to fork the branch and continue development on your own fork if you'd like.
I'm closing this pull request for now since it seems I won't be able to prioritize finishing it myself in the near future.

eranhirsch:
hey @sairus2k, I noticed I didn't go over your most recent changes yet, I'm sorry it slipped under my radar. I'll reopen the PR and have a look soon. If I see there are still open issues I might just push fixes myself. Sorry for this taking so long, I can imagine how frustrating it probably is to work on something for so long and not see it get shipped.

sairus2k:
Hey @eranhirsch , no worries at all. I understand that things can get busy.

I wanted to bring to your attention that this pull request introduces the very first regular dependency, which is an important consideration. Also, the TypeScript config change could potentially have wider impacts that we should evaluate.

Unfortunately, due to time constraints, I won't be able to continue working on this pull request. Thank you for your understanding.

cjquines:
heyy @eranhirsch , what's the status on this?

eranhirsch:
Sorry for the delays, I'm just catching up this week on a backlog of 2 months. I hope to get to his in the coming couple of weeks.

dartess:
@eranhirsch hey. The code has changed so many times that I'm a little lost. But if you see certain problems that need to be resolved, I'm still can help.

eranhirsch:
I'll send out fixes for this and would appreciate help with a review. Thanks @dartess

eranhirsch:
Oh boy!

Type-fest is both esm-only, requiring us to change our build slightly, but more importantly, it is also not backward compatible and defines its supported typescript version to 5.3 and its node version to 20 (21?). This makes it hard to import type-fest into our project. I couldn't get it to work, so I eventually copy-pasted the required parts of type-fest directly into the project.

I'd be happy to change things if someone has a better idea.

src/mergeDeep.test.ts
```diff
@@ -0,0 +1,135 @@
+import { expect, test } from 'vitest';
```
cjquines, line 1: **nit:** it doesn't look like any other test files import from vitest

src/mergeDeep.ts
```diff
@@ -0,0 +1,95 @@
...
+ * @dataFirst
+ * @category Object
+ */
+export function mergeDeep<
+  Destination extends Record<string, unknown>,
+  Source extends Record<string, unknown>,
+>(destination: Destination, source: Source): MergeDeep<Destination, Source>;
+
+/**
+ * Merges the `source` object into the `destination` object, while recursively merging any matching nested objects. When both objects have a value for a key and either value is not an object, the value from `source` would overwrite the value from `destination` (similar to a simple spread `{ ...destination, ...source }`).
```
cjquines, line 23: **nit:** the "either value is not an object" phrasing is a bit awkward?

**suggestion:**

```suggestion
 * Merges the `source` object into the `destination` object. When both objects have a value for a key:
 * - if both values are objects, they are recursively merged;
 * - otherwise, the value from `source` will be used (similar to a simple spread `{ ...destination, ...source }`).
```

src/mergeDeep.ts
```diff
@@ -0,0 +1,95 @@
+import { purry } from './purry';
+import { MergeDeep } from './type-fest/merge-deep';
+
+/**
+ * Merges the `source` object into the `destination` object, while recursively merging any matching nested objects. When both objects have a value for a key and either value is not an object, the value from `source` would overwrite the value from `destination` (similar to a simple spread `{ ...destination, ...source }`).
```
cjquines, line 5: (see comment for jsdoc below)

cjquines (approved):
> copy-pasted the required parts of type-fest directly into the project

i was gonna suggest this :sweat_smile:

dartess (approved):

dartess:
@eranhirsch hey! I finally replaced the self-written utility (copied from the previous PR) with a new version of remeda and found myself in debugging hell for a while.

The trick turned out to be that I couldn’t deep freeze the new value to the class instance. According to the previous logic, the old value was simply replaced in the instance. But your deepmerge simply replaced the entire instance, and I lost everything except the new values.

For now, I wrapped the instance in `R.clone(...)` before merging, this solved my problem. I just wanted to make sure - is this intended behavior, or should I open an issue with repl?

eranhirsch:
I'm having a hard time following your description of the issue, can you provide an example?

dartess:
I'm sorry about than, I'm still not familiar with English.

Sure:

https://codesandbox.io/p/devbox/amazing-jones-87z6zf?file=%2Findex.test.js%3A15%2C25 

Merging with a class instance occurs correctly only at the top level.

eranhirsch:
I now understand what you reported (and i'm sorry for the long delay, there were a lot of other things on the table to go through first).

I don't look at classes when considering implementations of object-based utility functions as they are kind of a different beast, and I think it's incorrect semantically to consider classes as simple objects (even when they have a "public" interface) because they should be used for different things, one edge-case I can think of is computed properties which are based on an internal state and implemented via a setter and getter that do non-trivial logic.

PR #318, opened 2023-05-20
feat: add findEntry

zorzysty:
This PR introduces `findEntry` function.

---

Make sure that you:

### DIFF
--- a/mapping.md
+++ b/mapping.md
@@ -24,6 +24,7 @@ documentation when migrating._
 | `equals`         | `isEqual`        | `equals`            |
 | `filter`         | `filter`         | `filter`            |
 | `find`           | `find`           | `find`              |
+| `findEntry`      | `-`              | `-`                 |
 | `findIndex`      | `findIndex`      | `findIndex`         |
 | `findLast`       | `findLast`       | `findLast`          |
 | `findLastIndex`  | `findLastIndex`  | `findLastIndex`     |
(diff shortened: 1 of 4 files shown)


eranhirsch:
This could be achieved with 

```ts
find(toPairs(obj), fn)
```

Is there an advantage of combining them in a new function?

zorzysty:
> This could be achieved with
> 
> ```ts
> find(toPairs.strict(obj), fn)
> ```
> 
> Is there an advantage of combining them in a new function?

Yeah, to me, it's mostly about simplicity and not requiring users to know about `toPairs.strict` function and figuring out how to combine it with `find` when they are only looking for "find in object" in the docs. 

Also type for `key` is a bit better:
obraz

eranhirsch:
> Also type for `key` is a bit better:

That's why you would use `toPairs.strict`, which makes sure that the types are the most accurate in JS (which has it's own peculiarities, like the fact that a Record<number, any> would not have number keys.

We also have a section for helpful one-liners for these situations: https://github.com/remeda/remeda/blob/master/mapping.md#helpful-one-liners

zorzysty:
Oh, I see I missed `.strict` in my example. 
Come to think of it, this kinda proves my first point that `findEntry` is just more convenient and easier to use :)

Anyway, I don't have other arguments, so feel free to close this PR if you don't find it useful.

TkDodo:
let's opt for closing this one. It's not clear to me that `findEntry` operates on an object and not an array, so we'd have to clarify the naming somehow. The provided solution by @eranhirsch to combine `find` and `toPairs` seems really good.

PR #320, opened 2023-05-24
fix(prop): Partially Revert #238

eranhirsch:
`prop` is broken when used inside a pipe, typescript can't infer the internal types:

```ts
const input = [{ a: 1 }];
const works = R.sortBy(input, R.prop('a')); // typed typeof input;
const doesntWork = R.pipe(input, R.sortBy(R.prop('a'))) // typed Record<any, any>;
```

//

### DIFF
--- a/src/prop.ts
+++ b/src/prop.ts
@@ -7,6 +7,6 @@
  * @data_last
  * @category Object
  */
-export function prop<K extends string = string>(name: K) {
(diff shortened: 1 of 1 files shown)


eranhirsch:
@TkDodo , @yanick  - I couldn't fix the typing for prop to work for my example. My typescript foo isn't strong enough

yanick:
Checking...

eranhirsch:
Fixed in #321

PR #321, opened 2023-05-24
improve on the type of prop

yanick:
change the type to be able to deal with the standalone form (where we don't know the source object type at all) and the form used in `pipe` (where we have an inference of the source).

For #320

### DIFF
--- a/src/prop.ts
+++ b/src/prop.ts
@@ -7,6 +7,13 @@
  * @data_last
  * @category Object
  */
-export function prop<K extends string = string>(name: K) {
-  return <T extends Record<any, any>>(obj: T) => obj[name];
+
+interface PropInterface {
+  <T, K extends keyof T>(name: K): (obj: T) => T[K];
+  <K extends string = string>(name: K): <T extends Record<any, any>>(
+    obj: T
+  ) => T[K];
 }
+
+export const prop: PropInterface = (name: string) => (obj: Record<any, any>) =>
+  obj[name];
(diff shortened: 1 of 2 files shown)


yanick:
@eranhirsch  @TkDodo  I think that fixes #320

src/prop.ts
```diff
@@ -7,6 +7,13 @@
  * @data_last
  * @category Object
  */
-export function prop<K extends string = string>(name: K) {
-  return <T extends Record<any, any>>(obj: T) => obj[name];
+
+interface PropInterface {
```
eranhirsch, line 11: I'd call this PropFunc, interface is just the "tool" here, it could (should?) be a type too

src/prop.ts
```diff
@@ -7,6 +7,13 @@
...
+  <T, K extends keyof T>(name: K): (obj: T) => T[K];
+  <K extends string = string>(name: K): <T extends Record<any, any>>(
```
eranhirsch, line 13: Can this just be ʼobjectʼ instead of record? Its a little more broad and correct (for example you can use prop("length“) on arrays.

If not i prefer the Record<PropertyKey, unknown> syntax as it's more eslint friendly

eranhirsch (approved):
Great idea, thanks!

eranhirsch:
Btw you don't need the type definition at all, it's what function signature overloading would provide for you natively

yanick:
> Btw you don't need the type definition at all, it's what function signature overloading would provide for you natively.

So it works without? Nifty. In that case, I'll let you remove the definitions and amend the PR with the variable names you prefer.

PR #322, opened 2023-05-25
Invert

LandonSchropp:
@TkDodo Sorry for the delayed response on #236. (I had a baby around that time, and this completely fell off my radar.) I've pushed up all of the changes you've suggested except for the more refined type. I wasn't sure if you saw my comment there, so I decided to reopen this.

I spent quite a while battling the compiler, but I couldn't find a way to make it happy with a generic type.

Here are the two main variations of the `Inverted<T>` type I spent time trying to get work. (I tried several other variants as well.) @texastoland Could you take a look at my branch and see if you can get either working?

``` typescript
type Inverted<T extends Invertable> = Record<Extract<T[keyof T], PropertyKey>, T[keyof T]>

type Inverted<Obj extends Record<PropertyKey, PropertyKey> | PropertyKey[]> = {
  [Value in Extract<Obj[keyof Obj], PropertyKey>]: keyof {
    [Key in keyof Obj as Obj[Key] extends Value ? Key : never]: never
  }
}
```

I also played with a few different implementations of `_inverted`, but all failed with a similar error when running `tsc`.

``` typescript
function _invert<T extends Invertable>(object: T): Inverted<T> {
  return Object.entries(object).reduce((accumulator, [key, value]) => {
    accumulator[value] = key;
    return accumulator;
  }, {} as Inverted<T>);
}
```

``` typescript
src/invert.ts:46:5 - error TS7053: Element implicitly has an 'any' type because expression of type 'any' can't be used to index type 'Inverted<T>'.

46     accumulator[value] = key;
       ~~~~~~~~~~~~~~~~~~
```

Do you have any other suggestions or improvements? I promise not to take 5 months to implement them this time. 😬

### DIFF
--- a/src/invert.test.ts
+++ b/src/invert.test.ts
@@ -1,0 +1,55 @@
+import { invert } from './invert';
+import { pipe } from './pipe';
+
+describe('invert', function () {
+  describe('with parameter', () => {
+    test('empty object', () => {
+      expect(invert({})).toEqual({});
+    });
+
+    test('no duplicate values', () => {
+      expect(invert({ a: 'd', b: 'e', c: 'f' })).toEqual({
+        d: 'a',
+        e: 'b',
+        f: 'c',
+      });
+    });
+
+    test('duplicate values', () => {
+      expect(invert({ a: 'd', b: 'e', c: 'd' })).toEqual({ e: 'b', d: 'c' });
+    });
+
+    test('numeric values', () => {
+      expect(invert(['a', 'b', 'c'])).toEqual({ a: '0', b: '1', c: '2' });
+    });
+  });
+
+  describe('without parameter', () => {
+    test('empty object', () => {
+      expect(pipe({}, invert())).toEqual({});
+    });
+
+    test('no duplicate values', () => {
+      expect(pipe({ a: 'd', b: 'e', c: 'f' }, invert())).toEqual({
+        d: 'a',
+        e: 'b',
+        f: 'c',
+      });
+    });
+
+    test('duplicate values', () => {
+      expect(pipe({ a: 'd', b: 'e', c: 'd' }, invert())).toEqual({
+        e: 'b',
+        d: 'c',
+      });
+    });
+
+    test('numeric values', () => {
+      expect(pipe(['a', 'b', 'c'], invert())).toEqual({
+        a: '0',
+        b: '1',
+        c: '2',
+      });
+    });
+  });
+});
--- a/src/invert.test.ts
+++ b/src/invert.test.ts
@@ -1,55 +1,53 @@
 import { invert } from './invert';
 import { pipe } from './pipe';
 
-describe('invert', function () {
-  describe('with parameter', () => {
-    test('empty object', () => {
-      expect(invert({})).toEqual({});
-    });
+describe('data first', () => {
+  test('empty object', () => {
+    expect(invert({})).toEqual({});
+  });
 
-    test('no duplicate values', () => {
-      expect(invert({ a: 'd', b: 'e', c: 'f' })).toEqual({
-        d: 'a',
-        e: 'b',
-        f: 'c',
-      });
-    });
-
-    test('duplicate values', () => {
-      expect(invert({ a: 'd', b: 'e', c: 'd' })).toEqual({ e: 'b', d: 'c' });
-    });
-
-    test('numeric values', () => {
-      expect(invert(['a', 'b', 'c'])).toEqual({ a: '0', b: '1', c: '2' });
+  test('no duplicate values', () => {
+    expect(invert({ a: 'd', b: 'e', c: 'f' })).toEqual({
+      d: 'a',
+      e: 'b',
+      f: 'c',
     });
   });
 
-  describe('without parameter', () => {
-    test('empty object', () => {
-      expect(pipe({}, invert())).toEqual({});
-    });
+  test('duplicate values', () => {
+    expect(invert({ a: 'd', b: 'e', c: 'd' })).toEqual({ e: 'b', d: 'c' });
+  });
 
-    test('no duplicate values', () => {
-      expect(pipe({ a: 'd', b: 'e', c: 'f' }, invert())).toEqual({
-        d: 'a',
-        e: 'b',
-        f: 'c',
-      });
-    });
+  test('numeric values', () => {
+    expect(invert(['a', 'b', 'c'])).toEqual({ a: '0', b: '1', c: '2' });
+  });
+});
 
-    test('duplicate values', () => {
-      expect(pipe({ a: 'd', b: 'e', c: 'd' }, invert())).toEqual({
-        e: 'b',
-        d: 'c',
-      });
-    });
+describe('data last', () => {
+  test('empty object', () => {
+    expect(pipe({}, invert())).toEqual({});
+  });
 
-    test('numeric values', () => {
-      expect(pipe(['a', 'b', 'c'], invert())).toEqual({
-        a: '0',
-        b: '1',
-        c: '2',
-      });
+  test('no duplicate values', () => {
+    expect(pipe({ a: 'd', b: 'e', c: 'f' }, invert())).toEqual({
+      d: 'a',
+      e: 'b',
+      f: 'c',
+    });
+  });
+
+  test('duplicate values', () => {
+    expect(pipe({ a: 'd', b: 'e', c: 'd' }, invert())).toEqual({
+      e: 'b',
+      d: 'c',
+    });
+  });
+
+  test('numeric values', () => {
+    expect(pipe(['a', 'b', 'c'], invert())).toEqual({
+      a: '0',
+      b: '1',
+      c: '2',
     });
   });
 });
--- a/src/invert.ts
+++ b/src/invert.ts
@@ -11,16 +11,24 @@ type Inverted = { [index: string]: string };
  *    R.invert(object)
  * @example
  *    R.invert({ a: "d", b: "e", c: "f" }) // => { d: "a", e: "b", f: "c" }
- *    R.invert({}) // => {}
- *    R.pipe(
- *      { a: "d", b: "e", c: "f" }
- *      R.invert()
- *    ); // => { d: "a", e: "b", f: "c" }
- *
+ * @data_first
  * @category object
  * @pipeable
  */
 export function invert<T>(object: Invertable): Inverted;
+
+/**
+ * Returns an object whose keys are values are swapped. If the object contains duplicate values,
+ * subsequent values will overwrite previous values.
+ * @param object the object
+ * @signature
+ *    R.invert()(object)
+ * @example
+ *    R.pipe({ a: "d", b: "e", c: "f" }, R.invert()); // => { d: "a", e: "b", f: "c" }
+ * @data_last
+ * @category object
+ * @pipeable
+ */
 export function invert<T>(): (object: Invertable) => Inverted;
 
 export function invert() {
--- a/src/invert.ts
+++ b/src/invert.ts
@@ -36,14 +36,14 @@ export function invert() {
   return purry(_invert, arguments);
 }
 
-function _invert<T extends object>(object: T): Inverted<T> {
-  const result: Record<PropertyKey, keyof T> = {};
+function _invert(
+  object: Readonly<Record<PropertyKey, PropertyKey>>
+): Record<PropertyKey, PropertyKey> {
+  const result: Record<PropertyKey, PropertyKey> = {};
 
   for (const key in object) {
-    // @ts-expect-error We ensure that the value is a valid type in the definition of Invertable
-    // above.
     result[object[key]] = key;
   }
 
-  return result as Inverted<T>;
+  return result;
 }
--- a/src/invert.ts
+++ b/src/invert.ts
@@ -1,7 +1,8 @@
 import { purry } from './purry';
 
-type Invertable = Record<PropertyKey, PropertyKey> | Array<PropertyKey>;
-type Inverted = Record<PropertyKey, PropertyKey>;
+type Inverted<T extends object> = T[keyof T] extends PropertyKey
+  ? Record<T[keyof T], keyof T>
+  : never;
 
 /**
  * Returns an object whose keys are values are swapped. If the object contains duplicate values,
@@ -15,7 +16,7 @@ type Inverted = Record<PropertyKey, PropertyKey>;
  * @category object
  * @pipeable
  */
-export function invert(object: Invertable): Inverted;
+export function invert<T extends object>(object: T): Inverted<T>;
 
 /**
  * Returns an object whose keys are values are swapped. If the object contains duplicate values,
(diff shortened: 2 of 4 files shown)


texastoland:
@LandonSchropp I reworked it in this [playground](https://www.typescriptlang.org/play?target=99&jsx=0#code/JYWwDg9gTgLgBAKjgQwM5wEpwGZQiOAIigFMQSATZQgKBpgE8wS4A1ZAGwFcSB5bADy8ARgCs4JAB4wSAOwroA3gF8AfHAC8cEaIDaAaxIMI2bWIC6dRszgAxYBxlQAQg3bcSQsROlyFcAGFkWQBJWQA3ElgAGjZOHh8ZeXR3Hn4vUVV1LUUaODgAegK4EGQwdABpIzgYCDhZEkioOGBTYBh0cPiWCggSVFkAcnhSmABjAAs4jzy4XSqGFtk4Q2NTHRR0HXmjc0S-FO64AH44BbgALnrGqPMr7YXLZSsmFiDQiKj4LQwSMegKAIAAp4ZiwBgLWIgiBgxgLdQAHzg0NhEKMuksNCkkFgNVecDCTRgGX2yUCwUJX2ycFy+SKJTKhw8NTqyBW1VaLRgg3QbK6HGAFHZDFmulSLGAy3F6R06jQ0wSUiS-hRXzRi1O4su1yad2FJjsDicrnFGVi4tUNGeNGwXFkYxgwAgy0lRIE-1kqHgGyVB3JHyJqgAFBAxH8YPcxABKK6U2AZdS0wrFAACHQAtNjw5moHhmhQSMx5HIxsB+nAJshInAOBAIKgWNZ+rNSDAuFBlhgAHSlMAAFQgOiDOnDXdWqBDYYdUdiQdWUc06l0odE4YMu1i-J45ijVpeNgAMiQYE50FpdIQAIKEWKEZw3ogBB+EAAiz4AouNCJYm3AjyeolQAAmMIC0kcstDjYl-1PS16XyAA9Y4aHg-I0PQjC0NQCYIC4DghRyOAAFYrkIDg5AAcxgCZCAAbjgS9SIABjouBnFIgBGViAlIoDWJfUiAGZWM-MZSIAFlY54gA). Happy to explain on a call. I'll make a few suggestions inline too.

@TkDodo A few observations working in the codebase:

1. I didn't see any tests for types. The 3 functions I tried inferred correctly. IMO it'd be enough to add `satisfies` everywhere: `expect(input satisfies Input)toEqual(output satisifies Output)`.
2. My types and the others I looked at reinvented patterns from generics utility libraries. I even saw copy pasta from [type-fest](https://github.com/sindresorhus/type-fest). Adding it as a dev dependency seems consistent with the vision for Remada.
3. `strict` functions don't make sense to me. What does it mean to be "not strict"? Strict types can always be assigned to looser types. I understand if it's about compilation speed but it isn't benchmarked.

I love the concept and just wanted to share 🙏🏼

LandonSchropp:
Thanks for the reply @texastoland! I read through your example. I thought the code was well-written, and I learned a few new things about TypeScript. 🙂

I did try to incorporate your solution, but I noticed that it doesn't work without `@ts-expect-error`. I wasn't sure if it made sense to suppress any type errors, although I noticed that this was a pattern used elsewhere in this codebase. @TkDodo, do you have a sense of what would be best here?

Here's the code:

```
import { mapToObj } from './mapToObj';

// A utility helper that extracts the type of the object's values.
type ValueOf<Obj extends {}> = Obj[keyof Obj]

// An invertable value
type Invertable = Record<PropertyKey, PropertyKey> | PropertyKey[]

// Maps Key to never if its value doesn't match Value
type FilterByValue<Obj extends Invertable, Value extends ValueOf<Obj>> = {
  [Key in keyof Obj as Obj[Key] extends Value ? Key : never]: Obj[Key]
}

// The result of inverting an object.
export type Inverted<Obj extends Invertable> = {
  [Value in ValueOf<Obj> as Value extends PropertyKey ? Value : never]: keyof FilterByValue<Obj, Value>
}

/**
 * Returns an object whose keys are values are swapped. If the object contains duplicate values,
 * subsequent values will overwrite previous values.
 * @param object the object
 * @signature
 *    R.invert(object)
 * @example
 *    R.invert({ a: "d", b: "e", c: "f" }) // => { d: "a", e: "b", f: "c" }
 * @data_first
 * @category object
 * @pipeable
 */
export function invert<Obj extends Invertable>(object: Obj): Inverted<Obj>;

/**
 * Returns an object whose keys are values are swapped. If the object contains duplicate values,
 * subsequent values will overwrite previous values.
 * @param object the object
 * @signature
 *    R.invert()(object)
 * @example
 *    R.pipe({ a: "d", b: "e", c: "f" }, R.invert()); // => { d: "a", e: "b", f: "c" }
 * @data_last
 * @category object
 * @pipeable
 */
export function invert<Obj extends Invertable>(): (object: Obj) => Inverted<Obj>;

export function invert() {
  return purry(_invert, arguments);
}

// function _invert<Obj extends Invertable>(object: Obj): Inverted<Obj> {
//   return Object.entries(object).reduce<Inverted<Obj>>(
//     (accumulator, [key, value]) => {
//       accumulator[value] = key;
//       return accumulator;
//     },
//     {} as Inverted<Obj>
//   );
// }

function _invert<const Obj extends Invertable>(object: Obj): Inverted<Obj> {
  // @ts-expect-error dependencies have loose types
  return mapToObj(Obj.keys(object), (key) => [object[key], value])
}
```

One last thing worth calling out: In the version I wrote, `tsc` produces an error, but VSCode doesn't. (I normally use Neovim, so I didn't know if this was normal behavior or not for VSCode.) Since VSCode is displaying other errors, could that be due to a TypeScript version mismatch? If so, maybe bumping TypeScript in this project would make the error go away?

LandonSchropp:
@texastoland Oh, one other quesiton. Why is it necessary to map the value to `never`? Why doesn't something like this work instead?

``` typescript
type FilterByValue<Obj extends Invertable> = Record<ValueOf<Obj>, keyof Obj>;
```

texastoland:
> Why doesn't something like this work instead?
>
> ```ts
> type FilterByValue<Obj extends Invertable> = Record<ValueOf<Obj>, keyof Obj>;
> ```


```ts
// your type is basically
type Loose = { [key: Value1 | Value2]: Key1 | Key2 }
// as opposed to
type Strict = { [key: Value1]: Key1, [key: Value2]: Key2 }
```

The goal of tricky generics is to preserve relationships between types. In the case of `FilterBy`:

```ts
type FilterByValue<Obj extends CanInvert, Value extends ValueOf<Obj>> = {
  // maps Key to never if its value doesn't match Value
  [Key in keyof Obj as Obj[Key] extends Value ? Key : never]: Obj[Key]
}
```

Each `Key` is either discarded (the effect of `Key ... as never`) or maintains its original value (`Obj[Key] extends Value`).

LandonSchropp:
@texastoland Gotcha. Thanks for the explanation!

src/invert.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+function _invert(object: Invertable): Inverted {
+  return Object.entries(object).reduce<Inverted>(
+    (accumulator, [key, value]) => {
+      accumulator[value] = key;
+      return accumulator;
+    },
+    {}
+  );
```
eranhirsch, line 45: Prefer a simple for loop instead of a reducer as the code is actually less readable, and there might even be perf implications.

src/invert.ts
```diff
@@ -0,0 +1,46 @@
+import { purry } from './purry';
+
+type Invertable = Record<PropertyKey, PropertyKey> | Array<PropertyKey>;
```
eranhirsch, line 3: You don't actually need invertable, you can infer the types in the output

src/invert.ts
```diff
@@ -0,0 +1,46 @@
...
+type Inverted = Record<PropertyKey, PropertyKey>;
```
eranhirsch, line 4: Here is what I came up with for typing

https://www.typescriptlang.org/play?#code/PTAEDUEMDsEsBt6VAFwJ4AdbQOYCg90MBTUASWgDdiAnFYgEwB4AVUYgD3ugYGdQA9gCMAVsQDGKAHygAvKBYBtANbE0AgGYKAuuy7Ee-AAo0BJOmgDSavKFAB+UACUJAmsyWr1WltoA0oF6aClK2oABcoNDE1DQA3AREpACCcuRUtPTMAN6gkJEAjHGgQpEATKAAvlIJSaAAQmkUsVlMLuJuzAAGABbEiAIA+gAk2bwoNNg4lV0BXQDubvAMI2MTUzNSNYSYpADCTRl0jEy5+SUCAvDEMFXbeCCgAOqwKD2ouwC0GsTjibugPYCHivWDAyDwaxoXhMeqQXjEAJAkEoMHQGTyABywMxAFdEJAhNcmGFsmE7IooaBsIE1ME4QjtJEGcRKWpdJxuHxAcCGKDgQ5QFTItFYgk7JUVHStCztHhtnVkXzUeD4EZYOJlLD4YieSi0RjQOrNSS7Cy-GElfzoBCoTDzXrlQb5QkdiR0i1GC83ixdgAxX4oVh6Ln8YRiSSGzzSnQhgzckxmTJWGx2Ry5RRGanQBRS7w6Jm0-NWlU2tUarUsAJGGSVMIimK0BJAA

src/invert.ts
LandonSchropp: I'm happy to cut it, but don't I need `Record<PropertyKey, PropertyKey>` somewhere to indicate that the values are themselves valid keys? Or am I missing something?

src/invert.ts
LandonSchropp: ✅ Done

src/invert.ts
LandonSchropp: TypeFest is really neat! I wish I would have know about that sooner.

src/invert.ts
LandonSchropp: I'm happy to cut it, but don't I then need ` T extends Record<PropertyKey, PropertyKey>` to ensure a caller of the function doesn't provide an object whose values aren't valid property keys?

I've left it for now, but please let me know what you'd like me to do as an alternative and I'll drop it. 🙂

LandonSchropp:
@eranhirsch Sorry for the delay! I've pushed up the changes you suggested.

1. I've removed the array case. To be honest, I'm not sure why I included it in the first place. ¯\\\_(ツ)\_/¯
2. Letting the library handle the details of the typing makes perfect sense to me, and I see your point about that being an advantage of using the library. 👍
3. Ah, the example you provided helps to clarify things. I didn't think about the return type having _all_ of the keys being extremely helpful information for the compiler.

Please let me know what you think! Thanks in advance.

src/invert.test.ts
```diff
@@ -0,0 +1,55 @@
+import { invert } from './invert';
+import { pipe } from './pipe';
+
+describe('invert', function () {
```
eranhirsch, line 4: No need to wrap the whole test suite with a describe, it'll show up as the file name. Use describe when you have more than one category of tests.

src/invert.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  describe('with parameter', () => {
```
eranhirsch, line 5: I think the convention is calling these "data-first" and "data-last"... it wasn't clear to me when reading this what parameter you are referring to

src/invert.ts
```diff
@@ -0,0 +1,49 @@
...
+  return purry(_invert, arguments);
+}
+
+function _invert<T extends object>(object: T): Inverted<T> {
+  const result: Record<PropertyKey, keyof T> = {};
+
+  for (const key in object) {
+    // @ts-expect-error We ensure that the value is a valid type in the definition of Invertable
+    // above.
+    result[object[key]] = key;
```
eranhirsch, line 45: Can you add which ts error we are expecting here... it just makes it easier if some new ts feature in the future can solve this to know that that was the problem encountered here... Here it's probably 2322?
```suggestion
    // @ts-expect-error [tsXXXX] We ensure that the value is a valid type in the definition of Invertable
    // above.
    result[object[key]] = key;
```

src/invert.ts
```diff
@@ -0,0 +1,49 @@
...
+export function invert<T extends object>(): (object: T) => Inverted<T>;
+
+export function invert() {
+  return purry(_invert, arguments);
+}
+
+function _invert<T extends object>(object: T): Inverted<T> {
+  const result: Record<PropertyKey, keyof T> = {};
+
+  for (const key in object) {
```
eranhirsch, line 42: I'm guessing that TS would have issues with `undefined` when trying to access `object[key]` because it can't tell if it exists in the input object.

I think we should use `Object.entries` here (it's not ES5, but we use it in several functions already). It would provide us a pair `[key, value]` where `value` is always well defined.

src/invert.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @category object
+ * @pipeable
+ */
+export function invert<T extends object>(): (object: T) => Inverted<T>;
+
+export function invert() {
+  return purry(_invert, arguments);
+}
+
+function _invert<T extends object>(object: T): Inverted<T> {
```
eranhirsch, line 39: Because of all the casts the implementation doesn't need to take the generic. You can type this a bit stricter and gain some help from TS. Now the values of this object are viable property keys and it might work without any error suppressions
```suggestion
function _invert(object: Readonly<Record<PropertyKey, PropertyKey>>): Record<PropertyKey, PropertyKey> {
```

eranhirsch (approved):
I think we can remove the error supression, maybe make this a little stricter, but if not this is also fine.

src/invert.test.ts
LandonSchropp: You're right. I'll change them.

src/invert.ts
LandonSchropp: With the other changes you mentioned, TypeScript isn't producing any errors here. I'm happy to refactor this to use `Object.entries` if you'd like, but the compiler seems happy without it.

src/invert.ts
LandonSchropp: `@ts-expect-error` is no longer necessary here. 🎉

LandonSchropp:
@eranhirsch Thank you for the review! I've updated my PR with your feedback.

src/invert.test.ts
LandonSchropp: ✅ Done

src/invert.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @param object the object
+ * @signature
+ *    R.invert()(object)
+ * @example
+ *    R.pipe({ a: "d", b: "e", c: "f" }, R.invert()); // => { d: "a", e: "b", f: "c" }
+ * @data_last
+ * @category object
+ * @pipeable
+ */
+export function invert<T extends object>(): (object: T) => Inverted<T>;
```
eranhirsch, line 33: (i think) you can move the type parameter to the returned function, it would make invert's data-last method a little more resilient to weird typing situations. I might be wrong though... i'm not sure.
```suggestion
export function invert(): <T extends object>(object: T) => Inverted<T>;
```

eranhirsch (approved):
Looks great! Thanks for addressing the comments

LandonSchropp:
Thank you!

Issue #281, opened 2023-03-10
Implement intersectWith


azhiv:
It would be awesome to have this utility function available. Simply `intersect` is not enough, consider the following example:
```ts
const arr = [{ id: 1, name: 'Box' }, { id: 3, name: 'Present' }];
const ids = [3, 5];

intersectWith(arr, ids, (item, id) => item.id === id);
// [{ id: 3, name: 'Present' }]
```

TkDodo:
would you like to contribute it?

michal-weglarz:
@azhiv @TkDodo Since I haven't noticed any interest in this so far, I've decided to give it a shot myself in the PR #325

PR #325, opened 2023-05-28
feat: add intersectionWith util function

michal-weglarz:
The new function `intersectionWith` has been added, accepting two arrays of any type, returning a list of intersecting values based on a custom comparator function which is invoked to compare elements of those arrays. 

`inersectionWith` can be:
- used with 'data first' approach
- used with 'data last' approach
- evaluated lazily
- pipeable

The equivalent functions in other libraries:
- [Ramda: `innerJoin`](https://ramdajs.com/docs/#innerJoin)
- [Lodash: `intersectionWith`](https://lodash.com/docs/4.17.15#intersectionWith)

Types of parameters passed to `intersectionWith` comparator are inferred from `array` and `other` types. However, there's problem with 'data last' approach, where the type of first parameter cannot be inferred properly and defaults to `unknown`. It is not an issue occurring only here though – e.g. `differenceWith` struggles with this as well. 

Additionally, unfortunately 'differenceWith' didn't fit in an existing column in `mapping.md` and because of that I needed to enlarge all of them. This is why you can see almost the entire file as modified. 

The idea for creating this util function has come from issue #281.

---

Make sure that you:


TkDodo:
since we have an `intersection` method, I think `intersectionWith` is fine too 👍

PR #329, opened 2023-05-30
chore(omit): Optimized for optional keys

eranhirsch:
In hot paths where the omit is used to filter out props which are rare (like normalization paths) the current impl creates too many redundant objects both while computing the omit (the set and the new object) and for the output.

In this implementation the object is only created if needed.

We also optimize the case of a single prop to remove to use spread so that the effort of creating the new object is left to the engine, instead of our impl.

Finally, we implement the the omit via built-ins so that we don't need to deal with our own loops. This might be optimized by the engine (at least when we move to more modern es versions)

### DIFF
--- a/src/omit.ts
+++ b/src/omit.ts
@@ -40,6 +40,10 @@ function _omit<T extends object, K extends keyof T>(
   data: T,
   propNames: ReadonlyArray<K>
 ): Omit<T, K> {
+  if (propNames.length === 0) {
+    return { ...data };
+  }
+
   if (!propNames.some(propName => propName in data)) {
     return { ...data };
   }
--- a/src/omit.ts
+++ b/src/omit.ts
@@ -2,34 +2,52 @@ import { purry } from './purry';
 
 /**
  * Returns a partial copy of an object omitting the keys specified.
- * @param object the object
- * @param names the property names
+ * 
+ * If the properties are optional and they don't exist in the input object then
+ * the same object is returned (unlike when using destructuring, which returns a
+ * new clone of the object).
+
+ * @param data the object
+ * @param propNames the property names
  * @signature
  *    R.omit(obj, names);
  * @example
  *    R.omit({ a: 1, b: 2, c: 3, d: 4 }, ['a', 'd']) // => { b: 2, c: 3 }
+ * 
+ *    const obj: { a?: number, b: number } = { b: 2 };
+ *    const result = R.omit(obj, ['a']);
+ *    result === obj // => true
  * @data_first
  * @category Object
  */
-export function omit<T extends Record<PropertyKey, any>, K extends keyof T>(
-  object: T,
-  names: ReadonlyArray<K>
+export function omit<T, K extends keyof T>(
+  data: T,
+  propNames: ReadonlyArray<K>
 ): Omit<T, K>;
 
 /**
  * Returns a partial copy of an object omitting the keys specified.
- * @param object the object
- * @param names the property names
+ *
+ * If the properties are optional and they don't exist in the input object then
+ * the same object is returned (unlike when using destructuring, which returns a
+ * new clone of the object).
+ *
+ * @param data the object
+ * @param propNames the property names
  * @signature
  *    R.omit(names)(obj);
  * @example
  *    R.pipe({ a: 1, b: 2, c: 3, d: 4 }, R.omit(['a', 'd'])) // => { b: 2, c: 3 }
+ *
+ *    const obj: { a?: number, b: number } = { b: 2 };
+ *    const result = R.pipe(obj, R.omit(['a']));
+ *    result === obj // => true
  * @data_last
  * @category Object
  */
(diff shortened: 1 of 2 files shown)


src/omit.ts
```diff
@@ -1,49 +1,80 @@
...
-  return Object.entries(object).reduce<any>((acc, [name, value]) => {
-    if (!set.has(name)) {
-      acc[name] = value;
-    }
-    return acc;
-  }, {}) as Omit<T, K>;
+  if (!propNames.some(propName => propName in data)) {
+    // Avoid creating a clone of the input object when the output is unchanged.
+    return data;
+  }
```
TkDodo, line 64: I'm not sure if returning the actual input is desired. From a user's perspective, I'd want the function to behave consistently - so either always return a new object or the same. Consider:

type FooBar = { foo: string, bar?: number }

const toFoo = (input: FooBar) => omit(input, ['bar'])

const myFooBar: FooBar = { foo: 'foo' }

const myFooOnly = toFoo(myFooBar)
```

here, I'd still expect `myFooOnly` to be a new object, not the same reference as `myFooBar`. If I were to mutate `myFooOnly` later, it should not impact `myFooBar`.

Also, it now behaves differently if I have:

```
const myFooBar2: FooBar = { foo: 'foo', bar: 5 }

const myFooOnly2 = toFoo(myFooBar2)
```

because now, `myFooOnly2` is a new copy because it happened to include the `bar` field ...

Comment (eranhirsch) [reply src/omit.ts:64]: fair points. I'm thinking maybe we should create some react-friendly util function that would return object A if they are equal, and object B if they are different (via deep equality) so that you could do something like:
```ts
setXXX((current) => {
  return pipe(
    current,
    ...,
    ...,
    ifChanged(current),
  );
});
```

wdyt?

TkDodo: yeah that would keep the original reference, and we could do that. additionally (or maybe instead), what would you think about providing something equivalent to the `structuralSharing` feature of react-query (code pointer: https://github.com/TanStack/query/blob/765f6196e900c1dd826b040773fcf97839ca60d0/packages/query-core/src/utils.ts#L217-L250)

it also gives you back the `original` if they are deep equal, however, if they are not, it still tries to keep as many references as possible. For example, if you'd compare & share these two arrays:

```
[{ foo: true }, { bar: false }, { baz: { nested: true } }]
[{ foo: true }, { bar: false }, { baz: { nested: true }, new: 'value' }]
```

You would get back an array with 3 values, where the first and second entry are the same reference as before, and only the last one is a new object. And even there, the `nested` object would also be the same reference as the original one.

Our implementation in Query only works with JSON compatible values though.

src/omit.ts
```diff
@@ -28,22 +29,37 @@ export function omit<T extends Record<PropertyKey, any>, K extends keyof T>(
...
-  const set = new Set(names as Array<string>);
-  return Object.entries(object).reduce<any>((acc, [name, value]) => {
-    if (!set.has(name)) {
-      acc[name] = value;
-    }
-    return acc;
-  }, {}) as Omit<T, K>;
+  if (propNames.length === 0) {
+    return { ...data };
+  }
```
TkDodo, line 45: could we make this case impossible on type level? What's the point of calling omit with an empty array of propNames? Making it a `NonEmptyArray` on type level would rid us of that path.

eranhirsch: I thought about it too. My thinking here was that someone might want to dynamically omit fields, maybe based on some UI component that defines the omitted parts, and that might contain a trivial empty case. Or if you want to optionally omit something and don't want to do:
```ts
const x = pipe(
  input,
  ...,
  omit(<CONDITION> ? [] : ["prop"]),
);

// Instead of 
const y = pipe(
  input,
  ...,
  <CONDITION> ? identity : omit(["prop"]),
  ...,
);

src/omit.ts
eranhirsch: I think we can start with the simple impl; the difference from react-query being that these are data processing pipes in memory and not fetched from the network and reparsed from scratch, so it's likely that the mutated data would already be a shallow clone.

I just don't want to add too much processing on hot paths, where the whole thing is a perf optimization. The main use-case would be after filters and sorts, where you'd want to prevent a re-render on trivial cases.

src/omit.ts
TkDodo: I would argue "don't call `omit` if you don't want to omit anything" ? In a pipe, the second solution looks fine. It conceptually quite similar to why we don't allow null/undefined to be passed to array methods - the check needs to be explicitly moved to the outside, _before_ the invocation of e.g. `filter`

eranhirsch: OK, but that creates an issue with how we change the signature of `omit` now that it's in the wild without breaking people's code :) ?

eranhirsch:
@TkDodo - are there any remaining open issues with this PR that you'd want to see addressed?

eranhirsch:
pinging this again @TkDodo

Issue #214, opened 2022-09-05
Support for IfElse, and Cond


EfstathiadisD:
One of the most beautiful helpers in Ramada are 'Cond' and 'IfElse'.

Is there any chance we can bring them into remeda? Basically the only thing that bugs me. But I understand it's early days.

Happy to Submit a PR if you provide some instructions.. Thanks for the hard work. 😊

TkDodo:
what would you need from us to make a PR ?

thelawclaw11:
This is something I'd like to see as well. I'll work on getting a PR ready!

cjquines:
i beleive this is closed by #331 --- i think `conditional` works fine for `ifElse`:

```ts
R.conditional([
  [R.isTruthy, trueCase],
  [R.not(R.isTruthy), falseCase],
]);
```

PR #331, opened 2023-05-31
feat: add conditional

thelawclaw11:
I added the cond function present in Lodash and Ramda. I found myself reaching for it the other day in a React project and thought I'd add it.


src/cond.ts
```diff
@@ -0,0 +1,49 @@
...
+ *    R.cond(value, pairs)
+ * @example
+ *    R.cond(
+ *      "Hello",
+ *      [[s => s === "Hello", s => "World"], [s => s === "Foo", s => "Bar"]],
+ *    ) // "World"
+ * @data_first
+ * @category Function
+ */
+export function cond<T, K>(value: T, pairs: Array<Pair<T, K>>): K | undefined;
```
eranhirsch, line 28: The pairs array should be readonly
```suggestion
export function cond<T, K>(value: T, pairs: ReadonlyArray<Pair<T, K>>): K | undefined;
```

src/cond.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @example
+ *    R.pipe(
+ *      "Hello",
+ *      R.cond([[s => s === "Hello", s => "World"], [s => s === "Foo", s => "Bar"]])
+ *    ) // "World"
+ * @data_last
+ * @category Function
+ */
+export function cond<T, K>(
+  pairs: Array<Pair<T, K>>
```
eranhirsch, line 44: Same here
```suggestion
  pairs: ReadonlyArray<Pair<T, K>>
```

src/cond.ts
```diff
@@ -0,0 +1,49 @@
+import { purry } from './purry';
+
+type Pair<T, K> = [(value: T) => boolean, (value: T) => K];
+
+function _cond<T, K>(value: T, pairs: Array<Pair<T, K>>) {
```
eranhirsch, line 5: And here...
```suggestion
function _cond<T, K>(value: T, pairs: ReadonlyArray<Pair<T, K>>) {
```

src/cond.ts
```diff
@@ -0,0 +1,49 @@
+import { purry } from './purry';
+
+type Pair<T, K> = [(value: T) => boolean, (value: T) => K];
```
eranhirsch, line 3: 1. The tuple should be readonly
2. For better documentation you can name each field in the tuple
```suggestion
type Pair<T, K> = readonly [predicate: (value: T) => boolean, mapper: (value: T) => K];
```

src/cond.test.ts
```diff
@@ -0,0 +1,32 @@
+import { cond } from './cond';
+import { pipe } from './pipe';
+
+describe('data first', () => {
+  test('should return value of first pair', () => {
+    const value = cond('Jokic', [
+      [s => s === 'Murray', s => 'point guard'],
```
eranhirsch, line 7: Use an empty params list to avoid this lint
```suggestion
      [s => s === 'Murray', () => 'point guard'],
```

eranhirsch (changes requested):
Thanks for adding this! I've added a few nits about typing.

One thing I'm not sure about is the naming. I'm not a big fan of abbreviations and here I'm not sure it communicates well enough what this function does or when it's good for. Even just calling it "conditionally" would be a good start.

@TkDodo - WDYT?

src/cond.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @param pairs pairs of functions
+ * @signature
+ *    R.cond(value, pairs)
+ * @example
+ *    R.cond(
+ *      "Hello",
+ *      [[s => s === "Hello", s => "World"], [s => s === "Foo", s => "Bar"]],
+ *    ) // "World"
+ * @data_first
+ * @category Function
```
TkDodo, line 26: from that description and example, I don't fully understand what this function is doing 😅

TkDodo (commented):
please also fix the lint warnings in the tests

TkDodo:
There's also an open issue mentioning `cond`:

- https://github.com/remeda/remeda/issues/214

eranhirsch:
@thelawclaw11 - Are you planning on continuing work on this?

thelawclaw11:
> @thelawclaw11 - Are you planning on continuing work on this?

Hey sorry I've been unresponsive. Yes I'm planning to work on this. Work has been crazy but I should have some free time this weekend or next.

src/cond.ts
cjquines: perhaps something like

```tsx
const cond = R.cond([
  [R.isString, s => `${s} is a string`],
  [R.isNumber, n => `${n} comes before ${n+1}`],
]);
cond("hello") //=> hello is a string
cond(5) //=> 5 comes before 6
```

cjquines:
`cond` is a traditional name: for example both lodash and ramda use it; i think it comes from [lisp](https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/node84.html).

of course, there is no need to follow tradition here

eranhirsch:
@thelawclaw11, I continued your work because I wanted to ship this. Could you take a look and see if it meets your requirements?

src/conditional.ts
```diff
@@ -0,0 +1,198 @@
...
+   * @param then - You only need to provide the transformer, the predicate is implicit.
+   */
+  export const defaultCase = <In>(then: (data: In) => unknown) =>
+    [() => true, then] as const;
+
+  /**
+   * An even simpler catch-all case that simply returns undefined, to match the behavior of other frameworks.
+   * @see defaultCase.
+   */
+  export const LEGACY_DEFAULT_CASE = defaultCase(() => undefined);
```
TkDodo, line 197: just a minor nit: why is this named `LEGACY`? does that mean it shouldn't be used? If so, why are we adding it ?

eranhirsch: Lodash and Ramda return `undefined` unconditionally because they have to take the final "else" case into account in the typing. I didn't want the return type to include `undefined` if it could be avoided as it could be messy for the rest of the pipe to handle it. I added the "defaultCase" function to make it more ergonomic, but then I wanted to make migration from other frameworks even easier, so I added the default behavior as an even easier drop-in const.

I called it "legacy" because I assumed that people coming from another framework would first use this to transition smoothly without changing behavior, and then they could tackle these cases in a later refactor. For those people, the existing behavior is the legacy/old/previous way things were done; similar to a "_FIX_ME" added after a deprecation or refactor.

I added a note about it in the function documentation:
> Unlike similar implementations in other frameworks, this does not return a default value when no cases match (like `undefined`), throwing an exception instead. To implement traditional default behavior, use `conditional.LEGACY_DEFAULT_CASE` as the final case or `conditional.defaultCast(then)` which allows for customized default handling.

does that make sense? if no, can you suggest an alternative?

PR #333, opened 2023-06-05
feat: length

lucasteles:
> Adds simple `length` operator

---

Make sure that you:

### DIFF
--- a/src/length.ts
+++ b/src/length.ts
@@ -1,0 +1,31 @@
+import { purry } from './purry';
+
+type Enumerable<T> = ArrayLike<T> | Iterable<T>;
+
+/**
+ * Counts values of the collection or iterable.
+ * @param items The input data.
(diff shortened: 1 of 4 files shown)


eranhirsch:
Why `count` and not `length`? I think names which are closer to their vanilla JS implementations are easier for users to find and understand.

src/length.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @category Array
+ */
+export function length<T>(items: Enumerable<T>): number;
+export function length<T>(): (items: Enumerable<T>) => number;
+
+/**
+ * Counts values of the collection or iterable.
+ * @param fn The predicate.
+ * @signature
+ *    R.count(fn)(array)
```
TkDodo, line 21: it still mentions `count` here in the docs

Issue #332, opened 2023-06-04
Feature Request: swap element in Array 


hyousi:
https://ramdajs.com/docs/#swap 

I this the `swap` function is extremely useful because it's a common use case in drag-and-drop. 

Ramda provides the `swap` function in the latest docs. It has been 8 years since the first feature request about the swap in lodash-js, but it's still working in progress...

eranhirsch:
I think this is a great addition, would you want to try sending a PR for it?

PR #342, opened 2023-06-30
feature: swap 

pukuba:
#332 
I have added a swap function. It's implemented identically to the one in Ramda, you can refer to this [link](https://ramdajs.com/docs/#swap) for more details.

---

Make sure that you:


eranhirsch (changes requested):
The typing here is a little loose, it's good for arrays, strings, and simple records, but it would widen types for tuples and for general objects, causing them to lose a lot of precision. This goes into more advanced typescript that isn't as easy to read though, especially for tuples.

This also doesn't enforce that for objects, the keys provided are actually keys of the object, so users would lose the safety of compile time errors when the wrong prop/key is used in a swap. You want to make sure that the keys provided use `keyof T` to couple everything.

Here is a typescript playground with examples of what I'm saying here. It types swap correctly for all cases.

https://www.typescriptlang.org/play?#code/C4TwDgpgBA8gRgKwgY2AZQO4EMwB4AqANFANICMUEAHsBAHYAmAzlANYQgD2AZlEaQCZKNeszYcefAHxQAvFADeAKChQA2iSgBLOuK698AXQBcfDRWq1GLTQH5BUUySGXRNqPfKPShgNxKAX38lUEgoAEEAJ0isEEwcXBU+YSsxSIgsBk46ABsQKABXOlY6Tgw6NUMoAB91Q0IkgElGagsRayg6AoBbOAhIhtVmhmoXdrEu3v7BqAAFTiYtYC1slLcodMzsvMLi0vLKmrq5OpmYSK0Acx0sHLWOzazc-KKSsoqq2sP5fCUZH-uaQyTx2ah03H6EWA80Wy2yxAAdEjwZCAEoQJjAQxJeyTPqRQEsYatJKqewAenJUAA6tBkFhdNt8iiCcAABbQULQLDAKDIbKYhm8sALJYrOhMUmqPhS0x4yGuDrEqgCKUUqm0vkMqBM7R0CGsjlQLlQHl8gXAIVQEWw8WS6XS34O0xqKUOmFi7JqABEOXol3Z3qqirEyrIbodHiglNgBQJ6WQccWq02yA5DCgYeITE4UAwdO1BSYnKNJrNytVkarsAu1zotzUFex1elpg9cIqvv9geD4yJLRVEYd6tj8ZQSfFGwyaYgGYr2dz+a1uiLJc54G5vLDQ8j5yuNxyjYHZGbLe8MZgcaniciyd0WhYWCgDC03AN9GFoo7C7z0FXxtLDdTRyR58jNdkzRtT06B3B1wmhL9xRmKsY01SIimvCcADcIB2VZ2WgdJMR1XgCONAowD9YgdGQdJug-HRLjmRDsh3JEESiGI4mwPB0UxYgs0zAcBGINRXn2OhESRdtxXqGt93rHIpAjU9TBjRpSKNLBolibQmDoAByXkn2ACi-V-ZcjL1A1jQ3R8WCYSBkFfLRkGtFiJSlfBgm4IpUEnJgeIIYgvBDFh2H0PgQrGVJwokAwpAACiSBgeSwUwiCSCKyCcMgZgigQnFVABKDLCSgTELjoS4cWjKlKsYx90igBAi15BrqqYYgCPSAyWFKPNdOAXN0jAIiPwqnjIAzNNtKwVB+hYHRTQq4AqqYrkpQ6mrVDKsKpy2Z5djeA5PjqWrQr7ToenxNUHH2+VIgjexONieI8H4chouUqs1KpfA2QfCq2U4AocgYQzeT6dymEWOA-QAQh3b0AFFUVRGBUSKzpOGM66pkib1ZTqvhAYckGwYhqzoZtOHEYjVH0cx3Kcbxx6id25J9s4RAUGAWr4CQVB3uC0g8sEH7vEZjGsagABhBkrMCnAAKBrkEe9Hy-I7SacES1LLVMHmheAYhstMbazY4QrVvW4rFCSfkJU4P0ERyThLn1tKrZAcWCuK-xVHSUzIl0BQgkCJQlCd4iiLB4B8AoyjoHkZWwGS1QfQ5HJ3e9YhvQwThInBvOoG9elgFL70skuKusDgZARm4INTRYGPTaSESkgAFiUAOlBjaUAD1bCj9upyYePXvyVOeMStRxZEqAAGZiG74gAFY5KX7v+8H1QR7Hi1FC1U3nw9qSESI4AAhONPkv34fR9UBRT9MNaCggYga9MMgBFXygfowBsiFKYb02dc5QFvlgNuFoZjlx5HnJI1cPZE37koIAA

pukuba:
Thank you for your review @eranhirsch!

I have implemented the changes to ensure a correct swap for all cases.
``` typescript
swap([1, 2, 3] as const, 0, -1) -> [3, 2, 1]
swap("hello" as const, 0, 3) -> "lelho"
swap([1, 2, 3] as const, 0, 10) -> "readonly [1, 2, 3]"
```

src/swap.ts
```diff
@@ -0,0 +1,329 @@
...
+    return _swapObject(item, key1, key2);
+  }
+}
+
+function _swapList<T>(
+  item: ReadonlyArray<T>,
+  index1: number,
+  index2: number
+): Array<T> {
+  const length = item.length;
```
eranhirsch, line 268: you don't need to memoize the length

src/swap.ts
```diff
@@ -0,0 +1,329 @@
...
+
+  if (positiveIndexA < 0 || positiveIndexA > length) {
+    return result;
+  }
+  if (positiveIndexB < 0 || positiveIndexB > length) {
+    return result;
+  }
+  if (positiveIndexA === positiveIndexB) {
+    return result;
+  }
```
eranhirsch, line 289: I don't think we need to optimize for this very specific case, the swap logic is sound without this.
```suggestion
```

src/swap.ts
```diff
@@ -0,0 +1,329 @@
...
+  result = Array.prototype.concat.apply(
+    [],
+    [
+      result.slice(0, positiveMin),
+      [result[positiveMax]],
+      result.slice(positiveMin + 1, positiveMax),
+      [result[positiveMin]],
+      result.slice(positiveMax + 1, length),
+    ]
+  );
```
eranhirsch, line 300: This is inefficient and redundant, why do you need to concat all these sub-arrays when the array is mutable. Simply swap the 2 items.

```suggestion
  result[positiveIndexA] = item[positiveIndexB];
  result[positiveIndexB] = item[positiveIndexA];
}

File: src/swap.ts [hunk 4]
```diff
@@ -0,0 +1,329 @@
...
+  const copy = clone(item);
+  if (
+    Object.prototype.hasOwnProperty.call(copy, key1) &&
+    Object.prototype.hasOwnProperty.call(copy, key2)
+  ) {
+    const tmp = copy[key1];
+    copy[key1] = copy[key2];
+    copy[key2] = tmp;
+  }
+  return copy;
```
eranhirsch, line 328: You should use spread operators and destructuring 
```suggestion
  const { [key1]: value1, [key2]: value2, ...rest } = obj;
  return { [key1]: value2, [key2]: value1, ..rest };
```

src/swap.ts
```diff
@@ -0,0 +1,329 @@
...
+  if (isArray(item)) {
+    return _swapList(item, key1, key2);
+  } else if (isString(item)) {
+    return _swapString(item, key1, key2);
+  } else {
+    return _swapObject(item, key1, key2);
+  }
+}
+
+function _swapList<T>(
```
eranhirsch, line 263: Why is it called list here? These are arrays...

eranhirsch (changes requested):
I'll be honest and say that I think this is either AI generated code, or copy-pasted from some other library. I'd be happy to include a comment somewhere in the code if this was copy-pasted from some other project, so they can get credit and so we can check on their implementation in the future if we encounter any bugs. If this is AI generated I'd also add a comment about it.

I'm lacking documentation for all these types, explaining what they do and how they work. It took me a while to read through and I'd like future maintainers to have an easier time.

Also some of the types aren't written ideally, and I wrote comments for them inline.

Lastly, the implementation of the function itself is inefficient and relies on concepts and style of JS that we try to avoid.

I'll be happy to see this ship, and that's why I put the time to add comments, pointers, and suggestions as much as possible, so please don't consider this a rejection. I'll also understand if you don't want to pursue this any further.

src/swap.ts
```diff
@@ -0,0 +1,329 @@
+import { purry } from './purry';
+import { isString } from './isString';
+import { isArray } from './isArray';
+import { clone } from './clone';
+
+type ParseInt<T> = T extends `${infer N extends number}` ? N : never;
+
+type Length<T extends ReadonlyArray<any>> = T extends { length: infer L }
+  ? L
+  : never;
```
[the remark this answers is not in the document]
pukuba: I'm not very skilled with TypeScript type operations, so I largely used code that I copied and pasted from a blog. Because it was code from the [blog](https://itnext.io/implementing-arithmetic-within-typescripts-type-system-a1ef140a6f6f), I didn't thoroughly check this part. I'm sorry for any confusion. I promise to pay more attention and verify everything more closely in the future.

src/swap.ts
```diff
@@ -0,0 +1,329 @@
...
+  ? Length<U>
+  : never;
+
+type NonNegativeInteger<T extends number> = number extends T
+  ? never
+  : `${T}` extends `-${string}` | `${string}.${string}`
+  ? never
+  : T;
+
+type Absolute<T extends number> = `${T}` extends `-${infer R}` ? R : `${T}`;
```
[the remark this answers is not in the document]
pukuba: This code is not working correctly
```typescript
type Absolute<T extends number> = `${T}` extends `-${infer R}`
  ? R extends number
    ? R
    : never
  : T;

type x = Absolute<-1>
        -> never
  ``` 

I also think that it's more appropriate for the Absolute type to return a number, so I'm going to change the code!

```typescript
type Absolute<T extends number> = `${T}` extends `-${infer R}`
  ? R extends `${infer N extends number}`
    ? N
    : never
  : T;
```
This code works correctly. I'll change it to this code!

src/swap.ts
```diff
@@ -0,0 +1,329 @@
...
+        ...ArraySwap<Rest, Index1, Index2, [unknown, ...Position], Original>
+      ]
+  : T;
+
+type SafeNegativeIndex<
+  T extends ReadonlyArray<unknown> | [],
+  Index extends number
+> = Subtract<Length<T>, ParseInt<Absolute<Index>>> extends never
+  ? never
+  : Subtract<Length<T>, ParseInt<Absolute<Index>>>;
```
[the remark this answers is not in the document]
pukuba: Oh! It's my mistake. Thank you

src/swap.ts
```diff
@@ -0,0 +1,329 @@
...
+      ? T
+      : Join<
+          ArraySwap<
+            StringToChars<T>,
+            SafeIndex<StringToChars<T>, K1>,
+            SafeIndex<StringToChars<T>, K2>
+          >
+        >
+    : T
+  : T;
```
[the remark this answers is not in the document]
pukuba: Oh, it's beautiful. Thank you!

src/swap.ts
pukuba: To keep the order the same as the input object, I'll set the code to this
```typescript
  const { [key1]: value1, [key2]: value2 } = obj;
  return {
    ...obj,
    [key1]: value2,
    [key2]: value1,
  };
  ```

File: src/swap.ts (hunk 6 above)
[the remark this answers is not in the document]
Comment (eranhirsch) [reply src/swap.ts]: The problem with using a blog (and AI in fact) is that the code represents what's the right way to do things in that specific period of time. The speed at which typescript moves, a blog post from 2020 just isn't current enough, and because it's not actual code that runs, just blobs of text, nothings there to make sure it stays up-to-date or even continues working.

File: src/swap.ts [hunk 10]
```diff
@@ -0,0 +1,291 @@
...
+  : 0 extends A
+  ? true
+  : 0 extends B
+  ? false
+  : isLessThan<Subtract<A, 1>, Subtract<B, 1>>;
+
+type BuildTuple<
+  L extends number,
+  T extends ReadonlyArray<unknown> = []
+> = T['length'] extends L ? T : BuildTuple<L, [...T, unknown]>;
```
eranhirsch, line 19: types usually describe the result, and not how they do it (they aren't functions). I think a better name for this would be TupleOfLength

src/swap.ts
```diff
@@ -0,0 +1,291 @@
...
+type Subtract<A extends number, B extends number> = BuildTuple<A> extends [
+  ...infer U,
+  ...BuildTuple<B>
+]
+  ? U['length']
+  : never;
+
+type IsNonNegativeInteger<T extends number> = number extends T
+  ? false
+  : `${T}` extends `-${string}` | `${string}.${string}`
```
eranhirsch, line 30: I know that for completeness the fractional part makes sense, but we don't actually need it here because it's not an input we need to care about. Nobody would try to swap item and index 9.75 and if they do I don't care about the type.
```suggestion
  : `${T}` extends `-${string}`
```

src/swap.ts
```diff
@@ -0,0 +1,291 @@
...
+  ? false
+  : true;
+
+type Absolute<T extends number> = `${T}` extends `-${infer R}`
```
eranhirsch, line 34: You can mix the two checks, but my bigger concern is that this isn't supported in older typescript versions, I think it was added in 4.7 or even 4.8, and we currently want to support a broader range of typescript versions. 

I think the best approach right now so you can start shipping stuff, and so people can use it is to split the `object` case from the array and string cases. I think that object has a much simpler type, and a much simpler impl, so it's an easy win and i'd be happy to see it ship asap. TL;DR - I'd split this function into `swapIndices` and `swapProperties`.

src/swap.ts
```diff
@@ -0,0 +1,291 @@
...
+    ? Join<
+        Rest,
+        Head extends string
+          ? Accumulator extends ''
+            ? Head
+            : `${Accumulator}${Head}`
+          : never
+      >
+    : Accumulator
+  : Accumulator;
```
eranhirsch, line 60: I just remembered we already have our version of the join type, checkout our join function!

Export the Joined type there and use it here...

https://github.com/remeda/remeda/blob/master/src/join.ts#L6-L20

src/swap.ts
```diff
@@ -0,0 +1,291 @@
...
+  ? false
+  : true;
+
+type Absolute<T extends number> = `${T}` extends `-${infer R}`
+  ? R extends `${infer N extends number}`
+    ? N
+    : never
+  : T;
+
+type StringToChars<T extends string> = string extends T
```
eranhirsch, line 40: Similar to my comment above, I think type's should describe the result, not what they do. I'd call this `CharactersTuple`

src/swap.ts
```diff
@@ -0,0 +1,291 @@
...
+  : 0 extends B
+  ? false
+  : isLessThan<Subtract<A, 1>, Subtract<B, 1>>;
+
+type BuildTuple<
+  L extends number,
+  T extends ReadonlyArray<unknown> = []
+> = T['length'] extends L ? T : BuildTuple<L, [...T, unknown]>;
+
+type Subtract<A extends number, B extends number> = BuildTuple<A> extends [
```
eranhirsch, line 21: Similar to my other comments, if we want to be extra strict about how we name types, this maybe should be named `Difference` as that's the result of subtraction...

src/swap.ts
```diff
@@ -0,0 +1,291 @@
...
+  T extends IterableContainer,
+  Index extends number
+> = isLessThan<Index, T['length']> extends true ? Index : never;
+
+type SafeIndex<
+  T extends IterableContainer,
+  Index extends number
+> = IsNonNegativeInteger<Index> extends false
+  ? SafeNegativeIndex<T, Index>
+  : SafePositiveIndex<T, Index>;
```
eranhirsch, line 108: I'd flip the order here..
```suggestion
> = IsNonNegativeInteger<Index> extends true
  ? SafePositiveIndex<T, Index>
  : SafeNegativeIndex<T, Index>;
```

eranhirsch (changes requested):
This looks really good, thanks for putting the time to go over my comments. I'm concerned that the `infer K extends number` syntax is not supported in all typescript versions we support. To be extra-safe I would want to split the 2 different concepts of swap we have here into 2 functions: `swapIndices` and `swapProps` (see my comment near the Absolute type).

Other then that I just have stylistic nits.

src/swap.ts
```diff
@@ -0,0 +1,284 @@
...
+    : T
+  : T;
+
+type Swap<
+  T extends ReadonlyArray<unknown> | string,
+  K1 extends number,
+  K2 extends number
+> = T extends string
+  ? string extends T
+    ? string
```
eranhirsch, line 127: This check should be part of SwapString

src/swap.ts
```diff
@@ -0,0 +1,284 @@
...
+ *   swap('abc', 0, 1) // => 'bac'
+ *
+ * @category Array
+ *
+ * @returns Returns the manipulated array or string.
+ *
+ * @data_first
+ */
+export function swap<
+  T extends ReadonlyArray<unknown> | string,
```
eranhirsch, line 156:
```suggestion
  T extends IterableContainer | string,
```

src/swap.ts
```diff
@@ -0,0 +1,284 @@
...
+    ? number extends K2
+      ? ReadonlyArray<T[number]>
+      : SafeIndex<T, K2> extends never
+      ? T
+      : SwapArrayInternal<T, SafeIndex<T, K1>, SafeIndex<T, K2>>
+    : T
+  : T;
+
+type Swap<
+  T extends ReadonlyArray<unknown> | string,
```
eranhirsch, line 122:
```suggestion
  T extends IterableContainer | string,
```

src/swap.ts
```diff
@@ -0,0 +1,284 @@
...
+> = IsNonNegativeInteger<Index> extends true
+  ? SafePositiveIndex<T, Index>
+  : SafeNegativeIndex<T, Index>;
+
+type SwapString<T extends string, K1, K2> = Joined<
+  SwapArray<CharactersTuple<T>, K1, K2>,
+  ''
+>;
+
+type SwapArray<T extends ReadonlyArray<unknown>, K1, K2> = K1 extends number
```
eranhirsch, line 107:
```suggestion
type SwapArray<T extends IterableContainer, K1, K2> = K1 extends number
```

src/swap.ts
```diff
@@ -0,0 +1,284 @@
...
+type ObjectSwap<T, K1 extends keyof T, K2 extends keyof T> = {
+  [K in keyof T]: T[K1 extends K ? K2 : K2 extends K ? K1 : K];
+};
+
+type SwapArrayInternal<
+  T extends IterableContainer,
+  Index1 extends number,
+  Index2 extends number,
+  Position extends IterableContainer = [],
+  Original extends ReadonlyArray<unknown> = T
```
eranhirsch, line 63: This is the other way around
```suggestion
  Position extends ReadonlyArray<unknown> = [],
  Original extends IterableContainer = T
```

src/swap.ts
```diff
@@ -0,0 +1,284 @@
...
+ *   swap(0, -1)('abc') // => 'cba'
+ *
+ * @category Array
+ * @returns Returns the manipulated array or string.
+ * @data_last
+ */
+export function swap<K1 extends number, K2 extends number>(
+  key1: K1,
+  key2: K2
+): <T extends ReadonlyArray<unknown> | string>(data: T) => Swap<T, K1, K2>;
```
eranhirsch, line 197:
```suggestion
): <T extends IterableContainer | string>(data: T) => Swap<T, K1, K2>;
```

src/swap.ts
```diff
@@ -0,0 +1,284 @@
...
+ *
+ * @example
+ *   swap('a', 'b')({a: 1, b: 2, c: 3}) // => {a: 2, b: 1, c: 3}
+ *
+ * @returns Returns the manipulated object.
+ */
+export function swap<K1 extends PropertyKey, K2 extends PropertyKey>(
+  key1: K1,
+  key2: K2
+): <T extends { [P in K1 | K2]: any }>(data: T) => ObjectSwap<T, K1, K2>;
```
eranhirsch, line 215: This is an old remanent, it should be modernized like the data-first impl
```suggestion
export function swap<T extends object, K1 extends keyof T, K2 extends keyof T>(
  key1: K1,
  key2: K2
): (data: T) => ObjectSwap<T, K1, K2>;
```

eranhirsch (approved):
This is great, *I still think it's worthwhile to split the object version from the array/string version*, as it would result in less code imported for each case and less typescript complexity for typescript to handle.

I've added some final nits. Would you mind going over them and sending a final fix before I merge this?

eranhirsch:
I ran the typescript checks now (the PR needed to be approved to run them). As I feared, the `${infer N extends number}` syntax is not supported in older typescript versions.

I don't strictly mind this, but it's one of the reasons I want to split `swapProperties` which works on objects and doesn't need this syntax from `swapIndices` which does.

@TkDodo  - wdyt about how we should handle the older typescript versions here...?

TkDodo:
> @TkDodo - wdyt about how we should handle the older typescript versions here...?

we could plan a v2.0 and put it there? I think we need clear restrictions on which TS versions / node versions and js versions we support, and test against all of those in our pipelines.

eranhirsch:
@pukuba - I've made some changes so we can ship this, i'll wait with merging so you can acknowledge the changes and see that everything still looks OK to you. Ping me if you have any questions, otherwise i'll ship this in a day or 2

src/swapProps.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @returns Returns the manipulated object.
+ *
+ * @data_last
+ */
+export function swapProps<
+  T extends object,
+  K1 extends keyof T,
+  K2 extends keyof T
+>(key1: K1, key2: K2): (data: T) => SwappedProps<T, K1, K2>;
```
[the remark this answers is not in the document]
eranhirsch: This was an intentional change. data-last functions are primarily built for pipes, there's not a lot of value in having a standalone function factory that takes some params but leaves the actual object they work on generic. If we want to enforce that the prop names used are strictly coupled to the object they are manipulating (as with the data-first impl) then we have to configure the keys in relation to the main type. Using PropertyKey means they can take any value.

the TL;DR is that in order to get IDE auto-complete, and to get compile time protection against using the wrong key, we have to type the function like this. Otherwise this is valid and won't surface the error:
```ts
pipe(
  { a: "hello", b: "world" },
  swapProps("aTypo", "b")
);
```ts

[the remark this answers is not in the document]
pukuba: Thank you so much for all the excellent reviews!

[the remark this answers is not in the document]
eranhirsch: Thanks for staying patient and contributing to the project!

PR #349, opened 2023-08-07
Fix sortBy example

dartess:
Fix error in example

`(x) => x.color` sort `asc` by default, so `green` will be first


eranhirsch (approved):
Thank You!

Issue #350, opened 2023-08-07
Unexpected `dropLast` behavior with negative `n`.


dartess:
```
R.dropLast([1, 2, 3], -1); // [1]
```
It seems the function should not take into account negative values.
I can do PR for that.

eranhirsch:
Because we can't limit the typing to only non-negative integers the only way to solve this would be by either *throwing* or *ignoring*. To make sure we don't break anyone's flow I would take negative numbers as if they are 0 (that is, drop nothing). It's important we return a new array as that is the expected pattern for all remeda functions (similar to how `array.slice()` would still return a new copy).

Do you want to send a PR for this?

dartess:
Yes i will do it

PR #352, opened 2023-08-08
fix(dropLast): Handle non-positive lengths

dartess:
Closed https://github.com/remeda/remeda/issues/350
Added a pinch of tests.
Also this change fixed the work with `NaN` as `N`.


eranhirsch (approved):
Thanks for doing this!

Issue #353, opened 2023-08-09
Request for new binary search based indexing functions


envil:
It would be nice to have binary search function implemented like in lodash:

https://lodash.com/docs/4.17.15#sortedIndex
https://lodash.com/docs/4.17.15#sortedIndexBy
https://lodash.com/docs/4.17.15#sortedIndexOf
...

eranhirsch:
Sounds good. Would you like to contribute these? I'd be happy to review, approve, and merge

envil:
Sure, I will give it a try. Thanks.

eranhirsch:
hey, just pinging to let you know I had an implementation of this in one of my projects so i'm sending a PR to include this.

eranhirsch:
This is done, would be part of the next remeda release

envil:
Great work, thank you!

PR #357, opened 2023-08-17
feat (sortedIndex): added functions for binary searching arrays

eranhirsch:
Solves #353.

Adds binary search based functions for finding an index in an array.

4 utility functions to match the lodash offering, based on `sortedIndexWith` which does the actual binary search.

---

Make sure that you:


src/sortedIndexWith.ts
```diff
@@ -0,0 +1,108 @@
...
+  ): number;
+  export function indexed<T>(
+    predicate: (item: T, index: number) => NonNullable<unknown>
+  ): (data: ReadonlyArray<T>) => number;
+  export function indexed(): unknown {
+    return purry(sortedIndexWithImplementation, arguments);
+  }
+}
+
+export function sortedIndexWithImplementation<T>(
```
TkDodo, line 87: because this function is exported here as well, it will be publicly available, as the `index.ts` file re-exports everything.

I would suggest to either:
- move this function to `_ sortedIndexWithImplementation.ts`
- only export specific things from the index barrel, and not everything from this file

I think I'd prefer the first one

TkDodo (approved):

TkDodo:
I'm afraid the commit message had a wrong format (one blank too many after `feat`) so it won't auto-release :/

`feat (sortedIndex): added functions for binary searching arrays`

eranhirsch:
any way to work around it without editing the push locally and force pushing to master 😨

TkDodo:
I merged the docs change PR in with a `fix` prefix. Just realized it should've been `feat` but whatever. sorry 😅 . at least there will be a release

Issue #354, opened 2023-08-10
Problem with `mapValues` types


dartess:
For some strange reason, I'm getting a key type compatibility error. In the callback, the key is described as a string. I can't pass it into `mapValues` because I get the error:

```
TS2345: Argument of type
(order: Order, id: string) => BotOrder
is not assignable to parameter of type
(value: Order, key: string | number) => BotOrder
Types of parameters  id  and  key  are incompatible.
Type  string | number  is not assignable to type  string 
Type  number  is not assignable to type  string 
```

I tried to simplify the example and reduced it to this demonstration of the problem:

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbziAhmAaigNgVwKYDOcAvnAGZQQhwBEUeIeAJijQNwCwAUNzAJ5g8cAPJQmeKHAC8iOMCYAuOADscIAEYSSnLgGMIygvGjioBAEp59YpZetMAPEajBlAcwA0IsRIB80xGIdfUNjHzMASWVxAA9mJSQAbQBrPD4lZ1c3AF0lUVMSAIQg7lKuVAxsfAIACm44OBMJCysTD3q4GqaoL3kASml-VSwsdq4GiYaAeim4AD0Afm4+nQrMXEI68cbwgijY5jGGrvDepgGpIZwRo8np2cXlnSA

Since I can't figure out where the difference between an index signature and a Record comes from, I suggest fixing the bug at the `remedy` type level. To do this, it is enough to add intersections with strings to the keys. This will be correct, because the implementation uses `Object.keys`, which always returns only strings anyway.

I can do a PR for this.

eranhirsch:
This isn't related to remeda, I don't know why `keyof` for an indexed object return `string | number`.

https://www.typescriptlang.org/play?#code/MYewdgzgLgBATgU1HAJgLhgbxgbQNIwCWYM0cxA5gLoYBEKArgLZMCetMAvjALxacBuALAAoUJFjEUCAB4J0WXAGsErDGUo0Y9Zmw7c+mQaNFRWABwTwkIVHlUReMFaxAAzGGcvvryFMJEAekCYUIA9AH5TCyspWXtWRz4XHy8EHzi5f1Fg0JhIoA

mapValues simply types the key as `keyof T`, it's doing what it's supposed to do here

dartess:
Yes but it can be easily fixed in remeda via `keyof T & string`

eranhirsch:
[...]

PR #358, opened 2023-08-18
improve mapValues types

dartess:
Closes https://github.com/remeda/remeda/issues/354

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -21,3 +21,10 @@ export type NonEmptyArray<T> = [T, ...Array<T>];
  * @see This was inspired by the type-definition of Promise.all (https://github.com/microsoft/TypeScript/blob/1df5717b120cddd325deab8b0f2b2c3eecaf2b01/src/lib/es2015.promise.d.ts#L21)
  */
 export type IterableContainer<T = unknown> = ReadonlyArray<T> | [];
+
+// Inspired and largely copied from `sindresorhus/ts-extras`:
+// @see https://github.com/sindresorhus/ts-extras/blob/44f57392c5f027268330771996c4fdf9260b22d6/source/object-keys.ts
+export type ObjectKeys<T extends Record<PropertyKey, unknown>> = `${Exclude<
+  keyof T,
+  symbol
+>}`;
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -24,7 +24,4 @@ export type IterableContainer<T = unknown> = ReadonlyArray<T> | [];
 
 // Inspired and largely copied from `sindresorhus/ts-extras`:
 // @see https://github.com/sindresorhus/ts-extras/blob/44f57392c5f027268330771996c4fdf9260b22d6/source/object-keys.ts
-export type ObjectKeys<T extends Record<PropertyKey, unknown>> = `${Exclude<
-  keyof T,
-  symbol
->}`;
+export type ObjectKeys<T extends object> = `${Exclude<keyof T, symbol>}`;
--- a/src/mapValues.test.ts
+++ b/src/mapValues.test.ts
@@ -1,6 +1,5 @@
 import { mapValues } from './mapValues';
 import { pipe } from './pipe';
-import { describe } from 'vitest';
 
 describe('data first', () => {
   test('mapValues', () => {
--- a/src/mapValues.test.ts
+++ b/src/mapValues.test.ts
@@ -34,3 +34,27 @@ describe('data last', () => {
     });
   });
 });
+
+describe('mapValues key types', () => {
+  test('interface', () => {
+    mapValues({} as { foo: unknown; bar: unknown }, (_, key) =>
+      expectTypeOf(key).toEqualTypeOf<'foo' | 'bar'>()
+    );
+  });
+
+  test('mapped type', () => {
+    mapValues({} as { [K in string]: unknown }, (_, key) => {
+      expectTypeOf(key).toEqualTypeOf<string>();
+    });
+  });
+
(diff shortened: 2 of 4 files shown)


src/_types.ts
```diff
@@ -21,3 +21,10 @@ export type NonEmptyArray<T> = [T, ...Array<T>];
  * @see This was inspired by the type-definition of Promise.all (https://github.com/microsoft/TypeScript/blob/1df5717b120cddd325deab8b0f2b2c3eecaf2b01/src/lib/es2015.promise.d.ts#L21)
  */
 export type IterableContainer<T = unknown> = ReadonlyArray<T> | [];
+
+// Inspired and largely copied from `sindresorhus/ts-extras`:
+// @see https://github.com/sindresorhus/ts-extras/blob/44f57392c5f027268330771996c4fdf9260b22d6/source/object-keys.ts
+export type ObjectKeys<T extends Record<PropertyKey, unknown>> = `${Exclude<
```
eranhirsch, line 27: Any reason to switch from `object` to `Record<PropertyKey, unknown>`. If the typing works for the objects we want, I don't mind us being broader and allowing unusual usages to be applicable too (and if they are buggy, to get an issue report).

https://www.typescriptlang.org/play?#code/C4TwDgpgBA8gRgKwgY2AaQiAzjANgEwB4AVKCAD2AgDt8soB7RFYAPigF4oADAEgG8AouWS4ArvgiEA1pgYAzKMQA0ULCAC2cBrlYBfbgG4AUKEixmqDNgByEAO4kylGnSgAlFAwBORAAreDJDeoNaqYtTS1Az21KzsXHxCIuKSMnKKKmqa2roGJsam4NAAwgCGuLhlcLjQXPzGUFAAFACUAFxQAEQAFhCVDF0mTWWdXfY+BEPGeoXIDNRYwFDkneWV1bWcFkioAHRlWFgAlgDm1M1tnOy9-biDqvxQo90T3lPP9POLwHqtBd8llBvBAsGJcMsuOQ2iYAPSwppNAB6AH5jICdBA9vdTs1yAd-sZ4YiSaTUYV0QsgSBtvBdsA9rJsHj-lBiRx2ABtADkZW5AF1CmZoNYcPJ1lUahA8PhaZZ0JgcARCBLNhBWHCEci0ULilBRTBxRVJbU7PY5fTRWaVca1RrCkA

```suggestion
export type ObjectKeys<T extends object> = `${Exclude<
```

eranhirsch (approved):
Very cool, see my suggestion, I think we should use `object`, it's broader and applicable to more cases and seems to work just fine

src/_types.ts
dartess: It doesn't matter at the moment, since the T sent always matches `extends Record<PropertyKey, unknown>`. But it might be useful in the future.

TkDodo:
could you please add a type-test that fails with the current version of the function but passes on this PR? Thanks.

dartess:
@TkDodo I don't see any type tests in this package. do you want me to take on the task of setting it up?

eranhirsch:
@dartess, here are example of typing tests we have already. We use vitest's `expectTypeOf`:

https://github.com/search?q=repo%3Aremeda%2Fremeda%20expectTypeOf&type=code

dartess:
@eranhirsch cool, I didn't know vitest could do that. I will add tests soon.

dartess:
@TkDodo done!

src/mapValues.test.ts
```diff
@@ -34,3 +34,27 @@ describe('data last', () => {
...
+
+  test('indexed signature', () => {
+    mapValues(
+      // eslint-disable-next-line @typescript-eslint/consistent-indexed-object-style -- must be "indexed signature"
+      {} as { [key: string]: unknown },
+      (_, key) => {
+        expectTypeOf(key).toEqualTypeOf<string>();
+      }
+    );
+  });
```
eranhirsch, line 59: Just to make sure we cover all our bases, can you add tests for:
* `{ [key in number]: unknown }`
* `{ [key: number]: unknown }`
* `{ [key in `prefix${string}`]: unknown }`
* `{ [key: `prefix${string}`]: unknown }`

eranhirsch (approved):
Looks great

src/mapValues.test.ts
dartess: good, also add tests for symbols

src/mapValues.test.ts
```diff
@@ -34,3 +34,51 @@ describe('data last', () => {
...
+    mapValues({} as { [K in number]: unknown }, (_, key) => {
+      expectTypeOf(key).toEqualTypeOf<`${number}`>();
+    });
+    mapValues({} as { [K in `prefix${string}`]: unknown }, (_, key) => {
+      expectTypeOf(key).toEqualTypeOf<`prefix${string}`>();
+    });
+    mapValues({} as { [K in symbol]: unknown }, (_, key) => {
+      expectTypeOf(key).toEqualTypeOf<never>();
+    });
+  });
```
eranhirsch, line 66: nit: I prefer using a `describe` here and breaking each assert to a separate test so that we get fine-tuned test results with the specific issue. Up to you if you want to change it.

eranhirsch (approved):
Perfection

src/mapValues.test.ts
```diff
@@ -1,5 +1,6 @@
 import { mapValues } from './mapValues';
 import { pipe } from './pipe';
+import { describe } from 'vitest';
```
eranhirsch, line 3: Hmmm... Why is this needed, we have global types for vitest, the file already used describe without it

PR #359, opened 2023-08-18
refactor(docs)!: use create-vite and drop parcel and version bump all dependencies

u-yas:
1. The current remeda docs are pretty legacy. No version of any tool was updated at all. So I changed all versions to the latest.

2. Since parcel is used, I thought about the future, and vite was used in the root project, so I used vite in the docs folder and removed parcel completely. There shouldn't be a strong reason to keep using parcel when you have vite.

3. This docs is deployed with gh-pages, so you shouldn't need express or anything like that. So I deleted it.

https://github.com/remeda/remeda/assets/52757302/46480045-af8c-4812-a36f-38ad9c067cd3

---

Make sure that you:


docs/package.json
```diff
@@ -1,33 +1,37 @@
...
+    "@types/react": "^18.2.20",
+    "@types/react-dom": "^18.2.7",
+    "@types/react-router-dom": "^5.3.3",
+    "@types/react-syntax-highlighter": "^15.5.7",
+    "@typescript-eslint/eslint-plugin": "^6.0.0",
+    "@typescript-eslint/parser": "^6.0.0",
+    "@vitejs/plugin-react": "^4.0.3",
+    "cross-env": "^7.0.3",
+    "eslint": "^8.45.0",
+    "eslint-plugin-react-hooks": "^4.6.0",
```
eranhirsch, line 31: you install this plugin but don't use it

docs/package.json
```diff
@@ -1,33 +1,37 @@
...
-    "parcel": "^1.9.4",
-    "typescript": "^3.8.3"
+    "@types/react": "^18.2.20",
+    "@types/react-dom": "^18.2.7",
+    "@types/react-router-dom": "^5.3.3",
+    "@types/react-syntax-highlighter": "^15.5.7",
+    "@typescript-eslint/eslint-plugin": "^6.0.0",
+    "@typescript-eslint/parser": "^6.0.0",
+    "@vitejs/plugin-react": "^4.0.3",
+    "cross-env": "^7.0.3",
```
eranhirsch, line 29: you dropped cross-env from the scripts, you should drop it from the dependencies too

docs/package.json
```diff
@@ -1,33 +1,37 @@
...
   "repository": "git@github.com:remeda/docs.git",
   "author": "Łukasz Sentkiewicz",
   "license": "MIT",
   "scripts": {
-    "start": "node index",
-    "watch": "parcel serve index.html",
-    "build": "cross-env NODE_ENV=production parcel build index.html --public-url ./  --out-file index.html && cp _redirects dist/_redirects && cp dist/index.html dist/404.html"
+    "watch": "vite",
+    "build": "tsc && vite build",
+    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
```
eranhirsch, line 10: Put `reportUnusedDisableDirectives` in the config itself

eranhirsch, line 10: 1. Put `reportUnusedDisableDirectives` in the config itself
2. We don't need to mention the extensions here, they should be part of the config too, and generally this isn't required
3. We don't need the 0 warnings here, it's something we can turn on in the CI to specifically prevent these from falling through, for the user it doesn't matter if something is a warning or an error
```suggestion
    "lint": "eslint",
```

docs/src/App.tsx
```diff
@@ -1,24 +1,25 @@
...
-      <Router>
-        <div className="wrapper">
-          <Header />
-          <Route exact path="/" component={Home} />
-          <Route path="/docs" component={Docs} />
-        </div>
-      </Router>
-    );
-  }
+  },[])
```
eranhirsch, line 12: This is most likely a debug print that was left behind, we can drop it

docs/src/Header.tsx
```diff
@@ -9,12 +8,12 @@ export function Header() {
...
       <ul className="navbar-nav mr-auto">
         <li className="nav-item ">
-          <NavLink className="nav-link" to="/" exact>
+          <NavLink className="nav-link" to="/" >
             Home
           </NavLink>
         </li>
         <li className="nav-item  ">
-          <NavLink className="nav-link" to="/docs" activeClassName="active">
+          <NavLink className={({isActive}) => `${isActive ? "active ":"" }nav-link`} to="/docs" >
```
eranhirsch, line 16: These spaces are hard to catch and easy to break, it's ok for the className string to contain redundant spaces, prefer this:
```suggestion
          <NavLink className={({isActive}) => `${isActive ? "active": ""} nav-link`} to="/docs" >
```

docs/src/index.tsx
```diff
@@ -1,5 +1,9 @@
...
+import React from 'react'
+import ReactDOM from 'react-dom/client'
+import {App} from './App.tsx'
 
-ReactDOM.render(<App />, document.getElementById('js-main'));
+ReactDOM.createRoot(document.getElementById('root')!).render(
+  <React.StrictMode>
+    <App />
+  </React.StrictMode>,
+)
```
eranhirsch, line 9: Drop the namespaces and import the specific functions we need from the modules

```suggeston
import { StrictMode } from 'react'
import { createRoot } from 'react-dom/client'
import {App} from './App.tsx'

createRoot(document.getElementById('root')!).render(
  <StrictMode>
    <App />
  </StrictMode>,
)
```

docs/tsconfig.node.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "composite": true,
+    "skipLibCheck": true,
+    "module": "ESNext",
+    "moduleResolution": "bundler",
+    "allowSyntheticDefaultImports": true
+  },
+  "include": ["vite.config.ts"]
+}
```
eranhirsch, line 10: I don't understand why we need this file? We can include vite.config.ts in the regular project

eranhirsch (approved):
This is really good work! I'd prefer we drop the eslint parts from it though and introduce lint separately. It will both reduce the complexity and size of this PR, and it will allow us a smaller "surface area" to debug if something goes wrong with this migration.

I don't have much experience with the docs site so i'll also let @TkDodo take a look.

Notice my few suggestions and questions, mainly the one about the `tsconfig.node.json` file which I'm not sure about.

docs/package.json
```diff
@@ -1,33 +1,37 @@
...
-  "main": "index.js",
   "repository": "git@github.com:remeda/docs.git",
   "author": "Łukasz Sentkiewicz",
   "license": "MIT",
   "scripts": {
-    "start": "node index",
-    "watch": "parcel serve index.html",
-    "build": "cross-env NODE_ENV=production parcel build index.html --public-url ./  --out-file index.html && cp _redirects dist/_redirects && cp dist/index.html dist/404.html"
+    "watch": "vite",
+    "build": "tsc && vite build",
```
eranhirsch, line 9: one more thing, you dropped the copying of the `_redirects` file to the dist folder, but I think that would cause the site to break if someone tries to access a deep link into the docs directly (and not by going to `/` and browsing from there). Can you validate that this isn't broken right now...?

eranhirsch (changes requested):

docs/package.json
u-yas: I will check gh-pages document!

u-yas:
I'm glad that you received this pr positively!  I will check and correct the review later!

docs/tsconfig.node.json
u-yas: This stackoverflow may have the best answer as to why tsconfig.node.json is needed. In short, this file is the tsconfig used by vite itself. On the other hand, react in the src folder refers to tsconfig.json

https://stackoverflow.com/questions/72027949/why-does-vite-create-two-typescript-config-files-tsconfig-json-and-tsconfig-nod

docs/src/index.tsx
u-yas: thank you! It's just like what you said. Use named imports!

docs/src/Header.tsx
u-yas: thank you! that is more appropriate

docs/package.json
u-yas: cross-env, @typescript-eslint/* , eslint, these modules can be removed. Change eslint config from root to extend

docs/package.json
u-yas: It's just like what you said

docs/package.json
u-yas: I'm trying to remove the eslint module from the docs folder. So I'm thinking of writing the actual lint code in root's package.json. From the docs folder, I'm thinking of replacing it with the following scirpts

docs/packages.json
```json
"lint": "cd ../ && yarn docs:lint",
```

./packages.json
```json
"docs:lint": "eslint docs/ -c docs/.eslint.cjs"
```

docs/src/App.tsx
u-yas: Similar code existed in the old code, so I implemented it to keep compatibility. If you don't mind, I will try to remove it!

u-yas: https://github.com/remeda/remeda/pull/359/commits/77377cf9c944e8cdf42878bb00135abc1564e783

deleted

docs/src/index.tsx
u-yas: https://github.com/remeda/remeda/pull/359/commits/92b0107dca6585598b21af04b84d1f0734788498

used named import it!

docs/package.json
u-yas: @eranhirsch 

I'm sorry, I misunderstood, but is https://remedajs.com/ deployed with Netlify?

I looked at .github/workflows/pages.yml and thought it was deployed on github pages, but looking at the response header on remedajs.com, I can see that it came from netlify.

image

.eslintrc.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
 
   parser: '@typescript-eslint/parser',
   parserOptions: {
-    tsconfigRoodDir: __dirname,
+    tsconfigRootDir: __dirname,
```
u-yas, line 7: typo found 😢

docs/index.html
```diff
@@ -1,24 +1,28 @@
 <!DOCTYPE html>
 <html lang="en">
+  <head>
+    <meta charset="utf-8" />
```
u-yas, line 4: I ran prettier --write

docs/package.json
u-yas: https://github.com/remeda/remeda/pull/359/commits/e0717c77f2e8290edbf8646b2cbe14436967e9d8

eslint that can be organized in the root folder is organized in the root folder

docs/package.json
u-yas: removed!

docs/package.json
u-yas: https://github.com/remeda/remeda/pull/359/commits/5d990066c590583963e2f82b5f0fa97fa9e23367

For the time being, I inserted the _redirects file inside the dist folder.

u-yas:
We have responded to your review and made corrections.
Also, since prettier --write was applied, please be aware that the file changed may be large.

https://github.com/remeda/remeda/pull/359/commits/8bfd37315a5e6d1c6449a320da1c4c8b4b1d90f7

docs/.eslintrc.cjs
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  root: false,
```
eranhirsch, line 2: You don't need this.
```suggestion
```

docs/.eslintrc.cjs
```diff
@@ -0,0 +1,15 @@
...
+  env: { browser: true, es2020: true },
+  extends: ['../.eslintrc.js', 'plugin:react-hooks/recommended'],
```
eranhirsch, line 4: you don't need this
```suggestion
  extends: ['plugin:react-hooks/recommended'],
```

docs/tsconfig.node.json
eranhirsch: yea, I just noticed i have this file too in one of my recent vite projects

eranhirsch (commented):
This looks great, as I've written in a previous review, I prefer you drop the changes to eslint so that we focus in this PR on the other changes. It adds complexity I have a hard time tracking and validating, and it's not strictly required at all for this PR.

Regarding deploying to Netlify: tbh, i've never worked on the docs site so I don't really know how it works! :) It looks like we have 2 deployments, one to gh-pages (which requires 404.html to be defined), and the other to netlify (which requires _redirects to be defined). Most likely the gh-pages one was an experiment that got abandoned and left behind. So... i'll leave finalizing this review to @TkDodo

TkDodo:
typechecking on docs is failing

TkDodo:
> Most likely the gh-pages one was an experiment that got abandoned and left behind. So... i'll leave finalizing this review to @TkDodo

I also don't know much about the docs deploy - I know it deploys to netlify, but @lstkz has the account for that.

eranhirsch:
I think we can drop the netlify version of the docs and use gh-pages as it keeps everything closer together. We lose the ability to have a sandbox for PRs for the docs site (I think that's what netlify was used here for, as they spin a vm for each PR), but that seems a very specific use-case that we don't necessarily even need.

TkDodo:
> I think we can drop the netlify version of the docs and use gh-pages as it keeps everything closer together.

I don't know, we'd have to setup custom urls and make sure the remeda url points to the gh-pages rather than netlify. Wouldn't we need access to the netlify account for that?

TkDodo:
I'd rather stay on netlify than move to gh-pages please

u-yas:
> typechecking on docs is failing

I've fixed. and I confirmed that it works with the following command

```
yarn compile && \
  yarn compile:dist && \
  yarn build && \
  yarn lint && \
  yarn docs:lint && \
  yarn docs:build 
```

package.json
```diff
@@ -36,6 +36,7 @@
     "lint": "eslint src/**/*.ts --fix",
     "lint:check": "eslint src/**/*.ts",
     "docs:install": "cd docs && yarn",
+    "docs:lint": "eslint -c docs/.eslintrc.cjs --ext .ts --ext .tsx docs/src ",
```
TkDodo, line 39: question: are we calling this in our pipelines somewhere ?

u-yas: No, I don't think the pipeline specifically calls that command.  Simply, there were commands such as docs:install in package.json, so I thought it would be more convenient to have a lint command, so I wrote it there.  Is it okay to just list this only in the docs folder?

eranhirsch (approved):
This looks great, I'm fine with moving ahead with this if @TkDodo doesn't object.

Check out my comments inline, just a few more cleanups

docs/index.html
```diff
@@ -1,24 +1,28 @@
...
-  <div id="js-main" class="main">
-
-  </div>
-  <script src="./src/index.tsx"></script>
-</body>
-
-</html>
\ No newline at end of file
+  <body class="bg-light">
+    <div id="js-main" class="main"></div>
```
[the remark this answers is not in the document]
u-yas: It was the old entry point. It has been deleted!

https://github.com/remeda/remeda/pull/359/commits/e9ffe778ea48a8a340e16ea6e50cc03c18efa475

docs/package.json
```diff
@@ -1,33 +1,33 @@
...
-    "@types/react-router-dom": "^4.3.0",
-    "compression": "^1.7.2",
-    "connect-history-api-fallback": "^1.5.0",
-    "express": "^4.17.3",
-    "react": "^16.3.2",
-    "react-dom": "^16.3.2",
-    "react-router": "^4.3.1",
-    "react-router-dom": "^4.3.1",
-    "react-syntax-highlighter": "^7.0.4"
+    "compression": "^1.7.4",
```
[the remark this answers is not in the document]
u-yas: thank you for checking!

https://github.com/remeda/remeda/pull/359/commits/e4ef88c3bfe7534912d58673fae35cd692e9b9da

docs/package.json
```diff
@@ -1,33 +1,33 @@
...
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-router": "^6.15.0",
```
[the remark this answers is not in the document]
u-yas: deleted ! thank you!

https://github.com/remeda/remeda/pull/359/commits/e4ef88c3bfe7534912d58673fae35cd692e9b9da

docs/package.json
```diff
@@ -1,33 +1,33 @@
...
   "author": "Łukasz Sentkiewicz",
   "license": "MIT",
   "scripts": {
-    "start": "node index",
-    "watch": "parcel serve index.html",
-    "build": "cross-env NODE_ENV=production parcel build index.html --public-url ./  --out-file index.html && cp _redirects dist/_redirects"
+    "watch": "vite",
+    "build": "tsc && vite build && cp _redirects dist/_redirects",
+    "lint": "cd ../ && yarn docs:lint",
+    "start": "vite preview"
```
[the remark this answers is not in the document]
u-yas: Previously, the script name was start, so I tried to maintain compatibility, but it's true that the default naming of vite is easier to understand :+1: 

https://github.com/remeda/remeda/pull/359/commits/e4ef88c3bfe7534912d58673fae35cd692e9b9da

.prettierignore
```diff
@@ -2,7 +2,7 @@ node_modules/
 
 /coverage/
 /dist/
-/docs/
+/docs/dist/
```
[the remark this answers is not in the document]
u-yas: fixed .prettierignore and .eslintignore!
https://github.com/remeda/remeda/pull/359/commits/c0916a33cc600b57c15ea08e1f6b0691c65c2b41

and

https://github.com/remeda/remeda/pull/359/commits/9eb29ed3b56b0d4331769c2e7048d36a6a6744e1

Issue #347, opened 2023-07-21
lint for correctness


TkDodo:
we can use:
- attw: https://arethetypeswrong.github.io/?p=remeda%401.24.0
  - ✅ all good right now
- publint: https://publint.dev/remeda@1.24.0

`publint` currently shows lots of warnings

we could include both tools in our CI pipelines

eranhirsch:
These are really cool, I tried looking into the suggestion that publint raises but it's too complex for me, do you know how to fix it?

PR #360, opened 2023-08-18
build!: tsup

louiss0:
This is a Pull request to implement tsup inside the repo. Tsup is a bundler that allows its user to output different file types for different modules. This is a PR to fix Issue #347. Tsup will not only create JS files but types for those JS files as well. 

Tsup does not allow to you to separate ESM files from CJS ones, it can only change the file names.

I decided to add some scripts because if you decide you don't like Tsup you can go back to the way things were.

I included `publint` a linter for the `package.json` file and `@swc/core`. Tsup gave me an error that told me that it had to be installed in order for the compilation to ES5 to be done.   

## Added Scripts

```
yarn run tsup:watch 
```

This script is the one that will watch for changes in the folder where the entry point is specified.

```
yarn run tsup:build 
```

This is the script that will build the files. 

```
yarn run publint
```

Lints the `package.json` file using `publint`

```
yarn run tsup:test
```

It will run this script `vitest run --coverage && yarn run tsup:watch`

## Configuration

I decided to use a `tsup.config.ts` file that has this config.

```ts
import { defineConfig } from 'tsup';

export default defineConfig(ctx => ({

    entry: ['src/index.ts'],

    format: ['cjs', 'esm'],

    minify: !ctx.watch,

    dts: true,

    splitting: false,

    clean: !ctx.watch,

}));
```

As you can see, the config file has a function that returns an object. it's parameter is called the context, which tsup uses to decide what to do when. The `!ctx.watch` tells tsup to only enable the option when tsup is building. 

`entry: ["src/index.ts"]` Tells tsup to only look for the `src/index.ts` file and bundle all exported modules that are in there.

`format:["cjs", "esm"]` Tells tsup to create a CJS and ESM file. 

`dts:true` Tells tsup to create a `d.mjs` and a `d.ts` file for this project. the `.js` is common JS. 

`splitting: false` Tells tsup not to code split.

`minify: boolean` Tells tsup whether to minify. I decided to only minify in production. 

`clean: boolean` Tells tsup whether to remove all files before build. I decided to only clean in production.


.gitignore
```diff
@@ -8,7 +8,7 @@ out.json
 scripts/def.json
 dist
 readme.html
-
+.vscode/
```
eranhirsch, line 11: The `.vscode` directory is important for project-level settings that we'd like to recommend everyone use (we currently don't do it). It shouldn't be ignored. 
```suggestion

```

package.json
```diff
@@ -17,23 +24,29 @@
...
     "compile:dist": "tsc --project tsconfig.dist.json",
+    "tsup:watch": "tsup --watch",
     "build": "yarn run clean && yarn run build:es && yarn run build:cjs",
     "build:es": "tsc --project tsconfig.build.json --module esnext --outDir \"./dist/es\"",
     "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir \"./dist/commonjs\"",
+    "tsup:build": "tsup",
     "prettier": "prettier . --write",
     "prettier:check": "prettier . --check",
     "lint": "eslint src/**/*.ts --fix",
+    "publint": "publint",
```
eranhirsch, line 49: Unless we run this in some CI pipeline there's no need to define a script for it. It's just mirrors `npx publint` (`npm run publint`)
```suggestion
```

package.json
```diff
@@ -17,23 +24,29 @@
     "lint-staged": "^13.2.0",
     "marked": "^0.4.0",
     "prettier": "^2.8.4",
+    "publint": "^0.2.1",
```
eranhirsch, line 27: And once we drop the script, there's no value in having this as a dependency. Unless this is part of our CI process we can just run it manually when we want to check it
```suggestion
```

eranhirsch (changes requested):
Unless we change the regular build scripts to use this then the changes to the package.json won't align with the our build and would break the package.

I'd leave the build process here to @TkDodo to review as I have less context and i've never used tsup to know enough about.

package.json
louiss0: I could not get this package to work with code sandbox. I will try to use the build scripts that were written to execute tsup instead.

louiss0:
I am glad the code-sandbox example works now. I did only what I was told to do by `publint`.

package.json
```diff
@@ -2,11 +2,19 @@
...
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.mts",
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.js"
+    }
+  },
   "sideEffects": false,
   "devDependencies": {
+    "@swc/core": "^1.3.78",
```
eranhirsch, line 17: Why do we include swc? I thought tsup uses esbuild under the hood.

package.json
```diff
@@ -18,17 +26,20 @@
...
   },
   "scripts": {
     "clean": "rm -rf ./dist && mkdir dist",
     "test": "vitest run --coverage && yarn run compile",
+    "tsup:test": "vitest run --coverage && yarn run tsup:watch",
     "test:ci": "vitest run",
     "compile": "tsc",
     "compile:dist": "tsc --project tsconfig.dist.json",
-    "build": "yarn run clean && yarn run build:es && yarn run build:cjs",
+    "tsup:watch": "tsup --watch",
```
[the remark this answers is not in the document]
louiss0: The watch script is there to replace the **compile** scripts in the codebase. I'll be happy to get rid of it for you but it's worth it to replace one of the compile scripts with it then remove the useless one.

package.json
```diff
@@ -18,17 +26,20 @@
...
     "ts-node": "^8.4.1",
+    "tsup": "^7.2.0",
     "typedoc": "^0.11.1",
     "typescript": "^5.0.4",
     "vitest": "^0.28.4"
   },
   "scripts": {
     "clean": "rm -rf ./dist && mkdir dist",
     "test": "vitest run --coverage && yarn run compile",
+    "tsup:test": "vitest run --coverage && yarn run tsup:watch",
```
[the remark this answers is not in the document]
louiss0: Ok then I guess you want to build then test. The test script is supposed to replace the one above it. You don't have to do that but if you are going to use tsup then you might as well use the command for it as the new one for the test script.

package.json
louiss0: I did this because I got an error saying that I needed it for ES5.

package.json
```diff
@@ -18,17 +26,20 @@
...
     "test:ci": "vitest run",
     "compile": "tsc",
     "compile:dist": "tsc --project tsconfig.dist.json",
-    "build": "yarn run clean && yarn run build:es && yarn run build:cjs",
+    "tsup:watch": "tsup --watch",
+    "build": "tsup",
     "build:es": "tsc --project tsconfig.build.json --module esnext --outDir \"./dist/es\"",
     "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir \"./dist/commonjs\"",
```
TkDodo, line 44: question: are those two build scripts still needed ?

TkDodo:
> error tsup@7.2.0: The engine "node" is incompatible with this module. Expected version ">=16.14". Got "14.21.3"
error Found incompatible module.

seems like tsup only works with node16+. I'm fine dropping node14 support from our pipeline, given that it's end of life was in April 2023

package.json
louiss0: No. I left them there in case you didn't merge the pr. I'll be happy to remove them for you

louiss0: > yeah please remove them 👍

Done!

louiss0:
I removed all the scripts that had to do with tsc and decided to place the ones that were necessary to tsup scripts. 
I hope you are convinced that this is the right path.

TkDodo:
can you please remove node14 from the pipeline checks:

https://github.com/remeda/remeda/blob/0138d52eca9e9cd3d9c78b6e3d965b4f767cf4bc/.github/workflows/ci.yml#L88

@eranhirsch would we be fine with dropping node14 support in a minor? I personally would be fine, because that version is end-of-life already...

louiss0:
> 

Ok I did it

louiss0:
> can you please remove node14 from the pipeline checks:
> 
> https://github.com/remeda/remeda/blob/0138d52eca9e9cd3d9c78b6e3d965b4f767cf4bc/.github/workflows/ci.yml#L88
> 
> @eranhirsch would we be fine with dropping node14 support in a minor? I personally would be fine, because that version is end-of-life already...

Ok I did it

.github/workflows/ci.yml
```diff
@@ -78,14 +78,7 @@ jobs:
           - 20
           - 18
           - 16
-          # Our testing infra (vitest) requires node 14 or higher, this
-          # effectively sets our guaranteed minimum supported version as we
-          # aren't testing on anything lower.
-          # When this value changes in the future it needs to be communicated
-          # to the users. (NOTE: because users can use remeda in client-only
-          # code too, this doesn't define the minimum ECMAScript version we
-          # support. That is defined by the tsconfig file.)
```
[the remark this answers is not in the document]
louiss0: It's done. I hope this is the last change I would need to do.

TkDodo:
I'm really hesitant to make this change in a minor version. Experience from the past has shown that every bundling change, especially one that has something to do with CJS / ESM changes, will break in a weird way for some consumers in one way or the other.

I'm not saying we shouldn't switch to tsup, but doing this in v2 seems more appropriate to me.

louiss0:
> I'm really hesitant to make this change in a minor version. Experience from the past has shown that every bundling change, especially one that has something to do with CJS / ESM changes, will break in a weird way for some consumers in one way or the other.
> 
> I'm not saying we shouldn't switch to tsup, but doing this in v2 seems more appropriate to me.

No pressure. If you want to switch later then you could simply save the code I wrote for you and use it later. 
I don't know what is coming for v2 but if it is radical then switching to Tsup at that time would be a good decision.

TkDodo:
@eranhirsch maybe we should talk about a v2 :)

louiss0:
> @eranhirsch maybe we should talk about a v2 :)

Is this PR still useful? I'm not sure if it needs to exist. I assume going to a v2 means radical change. I know there are some features missing.

TkDodo:
> Is this PR still useful?

absolutely. We haven't talked about v2 yet, but I don't think changes will be massive. It will likely be about dropping functions that we have deprecated lately, and maybe changing the behaviour of some where we've seen limitations.

louiss0:
> > Is this PR still useful?
> 
> absolutely. We haven't talked about v2 yet, but I don't think changes will be massive. It will likely be about dropping functions that we have deprecated lately, and maybe changing the behaviour of some where we've seen limitations.

Ok then. Please TQ v5 we all need that!

eranhirsch:
Thanks for your help guiding us through this. I've incorporated the changes suggested in #592 into the V2 release.

PR #361, opened 2023-08-20
feat(prop): improve the type of prop

hyousi:


eranhirsch:
Can you provide some context on what use cases this is needed that are currently not working?

hyousi:
I think it's redundant to pass the second type parameter if a user wants to have a getter function. 

```typescript 
type Data = {
  name: string;
};
// current version 
const getName = prop<Data, keyof Data>('name'); // $ExpectType (data: Data) => string

// with the default type parameter
const getName2 = prop<Data>('name'); // $ExpectType (data: Data) => string
```

eranhirsch (approved):

PR #364, opened 2023-08-22
docs(gh-pages): pass router base name

eranhirsch:
github pages are always located at `<ORGNAME>.github.io/<REPONAME>`. React router assumes the `/` route is always at the base of the URL. That's why it needs the base name to be defined for it to work if the root isn't at the URL root.

Here we use the fact that parcel inlines enviroment varialbes on build: https://parceljs.org/features/node-emulation/#environment-variables. We also assume that `${{ github.event.repository.name }}` would resolve to the current repo name ("remeda"). Finally, we assume that if the env variable is missing parcel would use undefined for the value, allowing us to fallback to the default root base name for the netlify deployment

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -39,7 +39,10 @@ jobs:
       - name: Build docs
         run: yarn docs:build
         env:
-          ROUTER_BASE_NAME: ${{ github.event.repository.name }}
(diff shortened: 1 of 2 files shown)


eranhirsch:
gh-pages env removed in #365

PR #370, opened 2023-08-29
chore: remove unused scripts

eranhirsch:
* These scripts seem unused.
* Took the opportunity to also cleanup the tsconfig from all the default comments which just make it harder to read. Added a schema that would generate documentation (at least in vscode) instead.


eranhirsch:
Moved to #374

PR #371, opened 2023-08-29
chore(docs): move marked processing to runtime

eranhirsch:
Our main package.json has a dependency on marked just because the docs site generation script uses it. It's a really old version of marked too!

It makes more sense for the site itself to make the conversion from marked to html as it's an implementation detail of how we use markdown, and it's also odd to convert things to html in a pre-processing step. ~~This means that a broken markdown now won't break the whole site build, only the specific component that broke, and we can fallback to the raw markdown so that the user get's the closest thing.~~ EDIT: Broken markdown would continue to break the build as is right now because vite would run the markdown as part of it's build process.

I've also added sanitization so that our HTML is always safe.


eranhirsch:
I'll reimplement it in the gen script

PR #380, opened 2023-09-06
refactor(docs): sass --> tailwind

eranhirsch:
Integrating tailwindcss into the project instead of the sass solution.

Bootstrap still brings a lot of styling changes that happen implicitly and we need to tackle them next, but in the meantime the site looks *almost* identical.


TkDodo:
oh, this PR didn't go to main 🤔

Issue #381, opened 2023-09-09
Compose isNil and isEmpty


norbornen:
Hello!

How to compose functions isNil and isEmpty (like isNilOrEmpty) and its return types?

eranhirsch:
Hey, some more context would really help in understanding what you are trying to achieve and what would be the ideal solution for it.

If you are filtering, you can pipe 2 filters one after the other, the performance cost over a single filter would be negligible in a pipe because filter is lazy-evaluated.

```ts
const result = pipe(
  data,
  filter(isDefined),
  filter(isEmpty),
);

**Comment by norbornen** [comment]
Hi! Thank you for your attention to my question.

I need two universal type-safe functions that take an argument of any type:

function "isNilOrEmpty" returns `true` if the argument is equals `null`, `undefined`, `"`, `[]`, `{}`
behavior is the same as when calling `(!isDefined || isEmpty)`

function+guard "notNilOrEmpty" returns `true` if the argument is not `null`, `undefined`, `"`, `[]`, `{}`
behavior is the same as when calling `(isDefined && !isEmpty)`.

Thanks.

**Comment by TkDodo** [comment]
I had a very similar issue today, trying to get an `isNonEmpty` util:

```
import * as R from 'remeda'
const data: Array<string> = []

const isNonEmpty = R.isNot(R.isEmpty)

if (isNonEmpty(data)) {
  console.log(data[0].toUpperCase())
}
```

this has two issues ([TypeScript playground](https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true#code/JYWwDg9gTgLgBAKjgQwM5wEpwGZQiOAcigFMQSATZQgKBoGMIA7VeKmZALjgEEopkATwA8rKMCYBzAHxwAvHADaAXTqMW8YKgByzAKLgYg+ZgB0W3TAAUGc6gNgjASjrBscKxf2HBV9sicnOABvGjg4dVQIABsSU2iIST9kDkUABmVTGAgAVTAwEigAYTQSK0CaAF8gA)):

1) `isNonEmpty` only accepts. Locally, I'm getting:
```
TS2345: Argument of type 'string[]' is not assignable to parameter of type 'null | undefined'.
```
but the sandbox shows a different error

2) This is more of a feature request, but after the `isNonEmpty` check, accessing [0] should be safe even under `noUncheckedIndexedAccess`.

In the past, I've made my own util for this:

```
type NonEmptyArray<T> = [T, ...Array<T>]
[...]

PR #382, opened 2023-09-14
feat(hasAtLeast): narrow non-empty arrays by length

eranhirsch:
Adds a utility that allows narrowing the type of arrays so that accessing individual indices is defined under strict typescript.

Closes: #381 

---

Make sure that you:


TkDodo (approved):
this is great 🙌

PR #387, opened 2023-09-29
fix(values): Update types to be more permissive

RedHatter:
Updates the types of `R.values` to allow keys that extend `PropertyKey`.

In the current version the follow fails due the the key type being too narrow.
```
interface Example {
  [index: string]: string
}

const a: Example = { b: 'c' }

R.values(a)
```

---

Make sure that you:


src/values.ts
```diff
@@ -15,8 +15,8 @@
  * @category Object
  */
 
-export function values<T>(
-  source: Record<PropertyKey, T> | ArrayLike<T>
+export function values<P extends PropertyKey, T>(
```
eranhirsch, line 18: There's no point in defining a generic if it isn't then used to tie the input and output. Our current typing isn't good because an index property can't be defined as a Record, and we are probably failing because of contravariance, and your fix simply swaps it to covariance so it "falls into place" (i might be mixing up contra- and co- variance here).

eranhirsch (changes requested):
The correct fix here would be to be more lenient on what types we accept for this method, and then define a better return type instead.

See this playground with a proposed implementation:

https://www.typescriptlang.org/play?#code/C4TwDgpgBAaghgGwK4QM4B4AqUIA9gQB2AJqlAPYBGAVhAMbAB8UAvFNngSWQE4RzFyhBCChJCAa0LkA7oQDaAXSgAfKEoBQUKAH4oAQR484ILPMJIAtpQg9FjLVABcBoybMSIIcgDN29gG4NDR9xBgBLISgAN0QUDA58IlIKGnomAApUciQeOggXTABKF3hkNCxmAG9HPmBcwigAeTSGADpY8tQsnLyIIqg4MjL4yqCAX2C6IVRgKFwARlYYuIqa7QB6De1tAD0dR3lwkjwXWZ5jgHNFM+ALwkuJxgyqwZcAIkp3qHGioI1poRZvMAEzLTrxDKhQgRKKEOCWCDEABiYWAkUIGQGVV+QS2Oyg+ymMzmuAAzODVt15AsADRQEH0smKP4afE7IkAknzAAslK6GRp9MZUGZgzIgNmrPZewOXKBpIArPzIa84C47igftLthy5ZLSQA2FUVABK9HIPGI6HOV3pFmstkYzzVLgWIIpuLZutlxIV8wA7CaMK8+AIhCIoKBIB86HBgN81O9BJd3gEoGHBMJRHBLgUoA6bDwfi7vQSidpo-n3nGE7THLn8wH616gA

Anyway, please add tests that would show that the fix now accepts something that was previously broken.

RedHatter:
Updated the types and added tests.

Thanks for the help!

src/values.test.ts
```diff
@@ -8,4 +8,21 @@ describe('Test for values as data first', () => {
...
+
+  it('should correctly type object keys', () => {
+    expectTypeOf(
+      values<{ type: 'cat' | 'dog'; age: number }>({ type: 'cat', age: 7 })
+    ).toEqualTypeOf<Array<'cat' | 'dog' | number>>();
+  });
+
+  it('should correctly type array keys', () => {
+    expectTypeOf(values([1, 2, 3] as const)).toEqualTypeOf<Array<1 | 2 | 3>>();
+  });
```
eranhirsch, line 27: Can you split the type tests to a different "describe" section?

eranhirsch (approved):
Thanks!

I'd add a few more tests (see my playground link) so that our tests do a better job of defining the expected semantics of the function now that it takes any object

src/values.test.ts
RedHatter: done

RedHatter:
I've added more tests.

src/values.test.ts
```diff
@@ -9,3 +9,41 @@ describe('Test for values as data first', () => {
...
+
+  it('should correctly type objects', () => {
+    expectTypeOf(values({ a: true })).toEqualTypeOf<Array<boolean>>();
+  });
+
+  it('should correctly type Records', () => {
+    expectTypeOf(values({ a: true })).toEqualTypeOf<Array<boolean>>();
+  });
```
eranhirsch, line 42: This is the same test, did you mean to use a literal generic value here?
```suggestion
  it('should correctly type Records', () => {
    expectTypeOf(values<Record<string, boolean>>({ a: true })).toEqualTypeOf<Array<boolean>>();
  });
```

eranhirsch (approved):
Thanks for doing this!

src/values.test.ts
RedHatter: I did indeed

eranhirsch (approved):

TkDodo (approved):
I fixed the lining errors

PR #393, opened 2023-10-11
feat: add `splice` function

fneco:
I would like to add  a `splice` function that works almost identically to the `Array.splice` method but does not mutate the data and does not return a deleted array.

Thanks for creating a great library!

---

Make sure that you:

---

### Available Types:

- `feat`: new functions!

existing).

</details>


eranhirsch (changes requested):
So this is just a non-mutating `splice` with a purried data-last impl. I think we should name it `splice`. Except people coming from Ramda, people who are more familiar with vanilla JS wouldn't except it to be named differently.

Once renamed, we should probably also accept the additional "replacement" array (but not as a variadic param as that would break purrying) so that it is complete.

fneco:
Thanks for checking my PR and replying!
I renamed `remove` to `splice` and add `replacement` parameter.  9e51ff7
And I editted PR title and body.

src/splice.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Removes elements from an array and, inserts new elements in their place.
+ * @param items the array to splice.
+ * @param start the zero-based location in the array from which to start removing elements.
+ * @param deleteCount the number of elements to remove.
+ * @param replacement the elements to insert into the items in place of the deleted elements.
+ * @signature
+ *    R.splice(items, start, deleteCount, replacement)
+ * @example
+ *    R.splice(2, 3, [1,2,3,4,5,6,7,8], []); //=> [1,2,6,7,8]
+ *    R.splice(2, 3, [1,2,3,4,5,6,7,8], [9, 10]); //=> [1,2,9,10,6,7,8]
```
eranhirsch, line 13: These examples are wrong and don't match the signature...
```suggestion
 *    R.splice([1,2,3,4,5,6,7,8], 2, 3, []); //=> [1,2,6,7,8]
 *    R.splice([1,2,3,4,5,6,7,8], 2, 3, [9, 10]); //=> [1,2,9,10,6,7,8]
```

src/splice.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Removes elements from an array and, inserts new elements in their place.
+ * @param items the array to splice.
+ * @param start the zero-based location in the array from which to start removing elements.
+ * @param deleteCount the number of elements to remove.
+ * @param replacement the elements to insert into the items in place of the deleted elements.
+ * @signature
+ *    R.splice(start, deleteCount, replacement)(items)
+ * @example
+ *    R.pipe([1,2,3,4,5,6,7,8], R.splice(2, 3)) // => [1,2,6,7,8]
```
eranhirsch, line 33: This example is wrong, the replacement array is required, it should always be provided (empty here)
```suggestion
 *    R.pipe([1,2,3,4,5,6,7,8], R.splice(2, 3, [])) // => [1,2,6,7,8]
```

src/splice.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    expect(splice([1, 2, 3, 4, 5, 6, 7, 8], 2, 3, [9, 10])).toEqual([
+      1, 2, 9, 10, 6, 7, 8,
+    ]);
+  });
+
+  test('immutability', () => {
+    const data = [1, 2, 3];
+    const result = splice(data, 0, 0, []);
+    expect(result).toEqual(data);
+    assert.notStrictEqual(result, data);
```
eranhirsch, line 16: I think it's easier when all assertions are of the same flavor, we usually use expect so you should use it here too
```suggestion
    expect(result).not.toBe(data);
```

src/splice.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+describe('data first', () => {
+  test('splice', () => {
+    expect(splice([1, 2, 3, 4, 5, 6, 7, 8], 2, 3, [])).toEqual([1, 2, 6, 7, 8]);
+    expect(splice([1, 2, 3, 4, 5, 6, 7, 8], 2, 3, [9, 10])).toEqual([
+      1, 2, 9, 10, 6, 7, 8,
+    ]);
+  });
+
+  test('immutability', () => {
```
eranhirsch, line 12: cool test

TkDodo:
@eranhirsch I'm not sure if we should have a function named splice that is always associated with mutating its input 🤔

src/splice.test.ts
```diff
@@ -0,0 +1,29 @@
+import { pipe } from './pipe';
+import { splice } from './splice';
+
+describe('data first', () => {
+  test('splice', () => {
+    expect(splice([1, 2, 3, 4, 5, 6, 7, 8], 2, 3, [])).toEqual([1, 2, 6, 7, 8]);
+    expect(splice([1, 2, 3, 4, 5, 6, 7, 8], 2, 3, [9, 10])).toEqual([
+      1, 2, 9, 10, 6, 7, 8,
+    ]);
```
eranhirsch, line 9: I'd split this into several tests. There's more value in having a unique test for each case (here: with replacement and without).

eranhirsch (changes requested):
This is great, just a few nits. Can you also add a few more tests to cover all the edge cases (index > length, index < 0, deleteCount > length, deleteCount < 0, etc...) so that if anyone changes the impl, they will know if they are changing an output that someone might be expecting.

eranhirsch:
@TkDodo , it's called splice even in the non mutating version. But the toXXX naming convention wouldn't work for us

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSpliced

src/splice.ts
```diff
@@ -0,0 +1,57 @@
+import { purry } from './purry';
+
+/**
+ * Removes elements from an array and, inserts new elements in their place.
+ * @param items the array to splice.
+ * @param start the zero-based location in the array from which to start removing elements.
+ * @param deleteCount the number of elements to remove.
+ * @param replacement the elements to insert into the items in place of the deleted elements.
```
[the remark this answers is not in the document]
fneco: I corrected it in 6083c89 .

src/splice.ts
fneco: Sorry, I corrected it in 6083c89 .

src/splice.test.ts
fneco: I corrected it in this line: https://github.com/remeda/remeda/pull/393/files#diff-0696fb3eb6df2fdeb8e06f4dab9cd586776e32d60d7001dc8eb02cd00ad58061R23

src/splice.test.ts
fneco: I fixed test case and assertion to be one-to-one. 

> This is great, just a few nits. Can you also add a few more tests to cover all the edge cases (index > length, index < 0, deleteCount > length, deleteCount < 0, etc...) so that if anyone changes the impl, they will know if they are changing an output that someone might be expecting.

I added regression testing. I tested with matrix data to include as many edge cases as possible.

cf. https://github.com/remeda/remeda/pull/393/files#diff-0696fb3eb6df2fdeb8e06f4dab9cd586776e32d60d7001dc8eb02cd00ad58061

eranhirsch (approved):
Thanks for this addition to the library, it's great

PR #394, opened 2023-10-11
feat: add `unset` function

fneco:
Add a `unset` function with the nearly same functionality as ramda's [dissoc](https://ramdajs.com/docs/#dissoc).

- what `nearly` means...: this `unset` don't handle array, but ramda's `dissoc` can handle arrays. This is consistent with the fact that with remeda's `set` don't handle array.
- I named it `unset`, because this function do the opposite of `set`. but, I have no particular preference.

Thanks for creating a great library!

---

Make sure that you:

---

### Available Types:

- `feat`: new functions!

existing).

</details>

### DIFF
--- a/mapping.md
+++ b/mapping.md
@@ -90,6 +90,8 @@ documentation when migrating._
 | `uniq`              | `uniq`              | `uniq`              |
 | `uniqBy`            | `uniqBy`            | `uniqBy`            |
 | `uniqWith`          | `uniqWith`          | `uniqWith`          |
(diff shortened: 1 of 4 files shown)


eranhirsch (changes requested):
We already have `omit` that does the same thing, including optimizing the single-prop case. I'm not sure this adds enough to the library.
```
const unset = (o, p) => omit(o, [p]); 
```

fneco:
Thank you for checking my PR and replying.

PR #402, opened 2023-10-24
feat(indexBy): add strict variant of indexBy

### LINKED ISSUE #398: indexBy return type widens object keys to `string` type
I'm currently using `indexBy` to construct an object that has its `id`s typed generically. Ideally, It would be nice if the return type of the function ensured that the object keys share the same type. Instead, the keys are always widened to `string`:

```ts
type Object<T extends string> = {
  id: T;
  property1: string;
}

const keyedObjects = indexBy([] as Object[], (i) => i.id)

typeof keyedObjects: Record<string, {
  id: T;
  property1: string;
}>
```

I think a signature similar to the below should allow for this to work properly. I'd be happy to open a PR, if its something people agree with.

```ts
export function indexBy<K extends PropertyKey, T>(
  array: ReadonlyArray<T>,
  fn: (item: T) => K
): Record<K, T>;
```

neefrehman:
Closes #398

`indexBy` currently widens all keys to a `string` type, ensuring that any extra type information is lost, and that other valid property keys (such numbers or symbols) can't be used.

This PR fixes this by adding a strict variant of the function which uses a new type parameter, `K`, tho maintains this type information and use it as the output record's key type.

---

Make sure that you:

---

### Available Types:

- `feat`: new functions!

existing).

</details>


src/indexBy.ts
```diff
@@ -33,33 +33,35 @@ export function indexBy<T>(
...
+  <K extends PropertyKey, T>(
+    array: Array<T>,
+    fn: PredIndexedOptional<T, K>
+  ) => {
+    return array.reduce<Record<K, T>>((ret, item, index) => {
+      const key = indexed ? fn(item, index, array) : fn(item);
       ret[key] = item;
       return ret;
-    }, {});
+    }, {} as Record<K, T>); // eslint-disable-line @typescript-eslint/prefer-reduce-type-parameter
```
neefrehman, line 54: It seems like this assertion, and disabling of the eslint rule, as a necessary evil, as otherwise the empty object can't be used as the initial value parameter. It's a little redundant to keep the type parameter, but leaving it there for now in anticipation of a better way to handle this issue.

eranhirsch: you'd have to type the object as Partial<Record<K, T>> otherwise typescript is correct that the types don't match. The empty object doesn't have props for each K

neefrehman: Thanks @eranhirsch. Though I think there will have to be some assertions unless I'm mistaken, as if we type the object as `Partial<Record<K, T>>`, then it seems we must also assert the result of the reduce—as it would affect the return type. I've gone with the below for now:

```ts
    const initialvalues: Partial<Record<K, T>> = {}
    return array.reduce((ret, item, index) => {
      const key = indexed ? fn(item, index, array) : fn(item);
      ret[key] = item;
      return ret;
    }, initialvalues) as Record<K, T>;
```

But this is also an option, which I decided against for readability reasons. Though I'm happy to change:

```ts
    return array.reduce<Partial<Record<K, T>>>((ret, item, index) => {
      const key = indexed ? fn(item, index, array) : fn(item);
      ret[key] = item;
      return ret;
    }, {}) as Record<K, T>;
```

eranhirsch: You are right that we'd have to cast again at the end, that's fine. I would split this into two lines so the casting is more obvious. Assign the reduce into a local variable and then return the result casted to the right type

neefrehman: I've just made that change 👍🏾

eranhirsch (changes requested):
TL;DR-
1. Typing this correctly is actually harder and there's a lot of places where it can return the wrong type. (i'll give an example)
2. Regardless, this would change the return type for existing users, I think we need to add a "strict" option to the function for this type, so that existing usages of this function don't break when we release this.
--
The simplest example of where this could go wrong is an empty array:
```ts
const result = indexBy(
  [],
  (item: string): "cat" | "dog => item === 42 ? "cat" : "dog",
);
const { cat } = results;
//      ^?:  string;
```
But cat is actually undefined (because the reducer never ran, we return the initial object which is `{}`.

To solve this we need to consider typescript's `exactOptionalPropertyTypes` which when enabled requires us to treat bounded records (those that have a finite set of values for it's `keyof`, e.g. `"cat" | "dog"`) differently than unbounded records (those that have an infinite one, e.g. `string`); at least returning `Partial<Record<...>>` for the former case.

src/indexBy.test.ts
```diff
@@ -6,17 +6,30 @@ const array = [
   { dir: 'right', code: 100 },
 ] as const;
 const expected = {
-  left: { dir: 'left', code: 97 },
-  right: { dir: 'right', code: 100 },
+  '97': { dir: 'left', code: 97 },
+  '100': { dir: 'right', code: 100 },
```
neefrehman, line 10: I've inverted the expected object here, and have used `.code` as the accessor in tests. This helps us have one object and indexer for all tests

src/indexBy.ts
```diff
@@ -52,15 +67,42 @@ const _indexBy =
...
+  fn: (item: T) => K
+): Partial<Record<K, T>>;
+
+export function indexByStrict<K extends PropertyKey, T>(
+  fn: (item: T) => K
+): (array: ReadonlyArray<T>) => Partial<Record<K, T>>;
+
+export function indexByStrict() {
+  return purry(_indexByStrict(false), arguments);
+}
```
eranhirsch, line 81: You are exporting indexByStrict which would make it accessible directly (and not via the namespace accessor).
```suggestion
function indexByStrict<K extends PropertyKey, T>(
  array: ReadonlyArray<T>,
  fn: (item: T) => K
): Partial<Record<K, T>>;

function indexByStrict<K extends PropertyKey, T>(
  fn: (item: T) => K
): (array: ReadonlyArray<T>) => Partial<Record<K, T>>;

function indexByStrict() {
  return purry(_indexByStrict(false), arguments);
}
```

eranhirsch (changes requested):

src/indexBy.ts
```diff
@@ -52,15 +67,42 @@ const _indexBy =
...
-    fn: PredIndexed<T, any>
-  ): Record<string, T>;
-  export function indexed<T>(
-    fn: PredIndexed<T, any>
-  ): (array: ReadonlyArray<T>) => Record<string, T>;
+    fn: PredIndexed<T, K>
+  ): Record<K, T>;
+  export function indexed<K extends PropertyKey, T>(
+    fn: PredIndexed<T, K>
+  ): (array: ReadonlyArray<T>) => Record<K, T>;
```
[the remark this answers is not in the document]
neefrehman: Thanks @eranhirsch. I've dropped this change for now. Had a quick look at nesting a namespace or using `export namespace indexBy.strict`, but in both cases was met by name clashing errors. I'd be curious to know the correct way to handle this, but I'm not desperate to add an indexed version in this PR.

src/indexBy.ts
```diff
@@ -2,16 +2,22 @@ import { purry } from './purry';
 import { PredIndexedOptional, PredIndexed } from './_types';
 
 /**
- * Converts a list of objects into an object indexing the objects by the given key.
+ * Converts a list of objects into an object indexing the objects by the given key
+ * (casted to a string). Use the strict version to maintain the given key's type, so
+ * long as it is a valid `PropertyKey`.
+ *
```
eranhirsch, line 8: Keep the docs in one line; our docs site would add the word breaks as needed.
```suggestion
 * Converts a list of objects into an object indexing the objects by the given key (casted to a string).
 * Use the strict version to maintain the given key's type, so long as it is a valid `PropertyKey`.
 *
```

src/indexBy.ts
```diff
@@ -52,6 +67,32 @@ const _indexBy =
...
+  <K extends PropertyKey, T>(
+    array: Array<T>,
+    fn: PredIndexedOptional<T, K>
+  ) => {
+    return array.reduce<Partial<Record<K, T>>>((ret, item, index) => {
+      const key = indexed ? fn(item, index, array) : fn(item);
+      ret[key] = item;
+      return ret;
+    }, {});
+  };
```
eranhirsch, line 94: We're not using the indexed version right now, so there's no point in making the code more complex.
```suggestion
  <K extends PropertyKey, T>(
    array: Array<T>,
    fn: (item: T) => K,
  ) => {
    return array.reduce<Partial<Record<K, T>>>((ret, item) => {
      const key = fn(item);
      ret[key] = item;
      return ret;
    }, {});
  };
```

eranhirsch (approved):
Looks great! Thanks for adding this. Check out my nits inline; once you send a fix, I'll merge this into main.

src/indexBy.ts
```diff
@@ -52,6 +66,30 @@ const _indexBy =
...
+function indexByStrict<K extends PropertyKey, T>(
+  fn: (item: T) => K
+): (array: ReadonlyArray<T>) => Partial<Record<K, T>>;
+
+function indexByStrict() {
+  return purry(_indexByStrict, arguments);
+}
+
+const _indexByStrict = <K extends PropertyKey, T>(
+  array: Array<T>,
```
eranhirsch, line 83:
```suggestion
  array: ReadonlyArray<T>,
```

eranhirsch (approved):

PR #405, opened 2023-11-03
build: migrate from yarn@1 to npm

eranhirsch:
# RFC

Remeda uses Yarn v1 as it's dependency tool and runner. Yarn v1 requires some non-trivial setup that might not be easy for newcommers to set up and might deter them from contributing to the project (it requires it to be installed as a global dependency, and requires users to know that they need v1 and not v2, and how to install v1 specifically).

npm comes bundled with node on every platform and is usually the first dependency util that people would know how to use (at least npm install, npm run, npx, etc...).

For a project the size of Remeda the problems that yarn solves are a non-issue as the project is small and doesn't have a lot of dependencies.


TkDodo (approved):

eranhirsch:
🥳

PR #409, opened 2023-11-23
feat: Add tap function

pekam:
This adds the `tap` function, which is useful for performing side effects within a `pipe` sequence, such as logging or persisting the value. The same function exists in Lodash and Ramda.

One annoying thing is that probably the most common use case, `console.log`, breaks the typings when referencing the function directly:
```ts
pipe(
  [-1, 2],
  filter(n => n > 0),
  tap(console.log),
  map(n => n * 2) // 'n' is of type 'unknown'
);
```
This can be avoided by using an arrow function:
```ts
pipe(
  [-1, 2],
  filter(n => n > 0),
  tap(value => console.log(value)),
  map(n => n * 2)
);
```
...or by creating your own log function with a generic type instead of `any`:
```ts
const log = <T>(value: T) => console.log(value);
pipe(
  [-1, 2],
  filter(n => n > 0),
  tap(log),
  map(n => n * 2)
);
```

---

Make sure that you:

---

### Available Types:

- `feat`: new functions!

existing).

</details>


eranhirsch (approved):
This is very cool! Thanks for doing it.

I think this could fix the typing problem with console.log

function tap<T, S extends T>(interceptor: (data: T) => void): (data: S) => S;

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbziAhmOBfOAzKERwBEUApiCQCYqEDcAUHdgK4B2AxjMBC3DGgDwAVADRwAynBIAPGCRYUAznEEA+ABTAWsqGxJgY0AFxw1VPscEBKOAF4VcAG4RgFS8dMpz463fGI6cHCkMExQPGoOKAA2PvYIAYFwmtq6+tAR0Zb0iUEkIWGO0dmYdBgMbNwK8GYoxqQoFNxRAJ5wLEwgAEYkUADaALq2cAP0FSxVcFJDqGBqCQD084EAegD8CTXCCXyzYwoQUSQAdFEQAOaWW1kMQA

eranhirsch:
```ts
function tap<T, F extends (data: T) => any>(interceptor: F): (data: T) => T;
```

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbziAhmANHMwwFNMBmwANjLlHAL5wFQQhwBEUuIuAJiowNwBQvBAK4A7AMYxgEYXBhoAPABVMAMTi4AHmWHsAznAAUnWQC44CgJRwAvAD44KYQE8b+4MLJRRuMDGinl5qaGKCZmlrZmiLxwcCwwglDS+gBuKMThdgjRMXBuHl4+0Clp5nw5sbjxiXCpxGVUvJT8vKJSOvDq1lg4uPrZANoAjOgATOgAzAC66NmoYPqdEZ0A1HCD5jMxAPRbMQB6APzZsvOtwjoQxLgAdMQQAOYbs2j6jtZ2b6vrm3A7+0cxE76ABe7zgwNuuGE9xgAAs4F8nttdnBDtkiKRyPoAO5g3F2b7ZP7lNHmIA

This would do the trick. We covered some excellent cases here, so let's ensure the tests represent all of them so it's easy to refactor this in the future.

pekam:
Great, thanks!

I started thinking, should we just drop the data first version, since the function is only useful in a pipe (as far as I know), or would you prefer to keep both versions for consistency?

src/tap.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    expect(
+      pipe(
+        [-1, 2],
+        filter(n => n > 0),
+        tap(foo),
+        map(n => n * 2)
+      )
+    ).toStrictEqual([4]);
+  });
```
pekam, line 68: Is it okay to use test cases for ensuring TS type inference? Technically, running this test doesn't provide any extra value, because the functionality is already tested. The code just needs to be compiled with `tsc` during validation, so we could also just wrap it in a function that is never called. Are there similar cases in the codebase?

eranhirsch: You can add an `exceptTypeOf` in the middle of the pipe. It'll be better than just having code that compiles against the types, but the current implementation is OK, too.

```suggestion
  it('should infer types after tapping function reference with parameter type any', () => {
    // (same as console.log)
    function foo(x: any) {
      return x;
    }

    expect(
      pipe(
        [-1, 2],
        filter(n => n > 0),
        tap(foo),
        map(n => { 
          expectTypeOf(n).toEqual<number>();
          return n;
        })
      )
    ).toStrictEqual([4]);
  });
```

src/tap.test.ts
```diff
@@ -0,0 +1,69 @@
+import { filter } from './filter';
+import { map } from './map';
+import { pipe } from './pipe';
+import { tap } from './tap';
+
+const value = [1] as const;
```
eranhirsch, line 6: I prefer the name data, regardless, this is a const so have it ALL_CAPS

```suggestion
const DATA = [1] as const;
```

eranhirsch (approved):
This is cool. Thanks so much for doing it.

Let's keep the dataFirst implementation; I can't think of a good use case right now.

Add a line somewhere in the function doc-block to explain that the return value of the interceptor function is ignored. It would add to the clarity of what tap does and how it's used.

src/tap.test.ts
pekam: Cool, done. :+1: 

I tested that running `npm test` doesn't catch these type errors. According to vitest docs, we should add `--typecheck` flag: https://vitest.dev/guide/testing-types#run-typechecking

...but when I tested adding that flag, I get `CACError: Unknown option --typecheck`, which is weird since they say it should work since vitest 1.0 and remeda has version `^1.0.1`.

But I guess `tsc` catches those anyway during build.

src/tap.test.ts
pekam: :+1: Done

eranhirsch (approved):
🥳

pekam:
Thank you @eranhirsch for a pleasant review process and for maintaining the best utility library for TS! :)

PR #410, opened 2023-11-24
type(toPairs): Improve toPairs.strict typings

cloud-walker:
The current `toPairs.strict` implementation returns a weird ObjectEntry thing that loose the tuple binding between each key and value

---

Make sure that you:

---

### Available Types:

- `feat`: new functions!

existing).

</details>


src/toPairs.test.ts
```diff
@@ -25,7 +25,7 @@ test('stricter typing', () => {
 
 test('stricter typing with optional', () => {
   const actual = toPairs.strict({} as { a?: string });
-  assertType<Array<['a', string]>>(actual);
+  assertType<Array<['a', string | undefined]>>(actual);
```
cloud-walker, line 28: I think is correct that the undefined is part of the value type of the tuple

eranhirsch: I don't think so, If `a` is just optional, but not nullable, when will `Object.toEntries()` ever return `[['a', undefined]]`?

src/toPairs.ts
```diff
@@ -16,22 +14,10 @@ export function toPairs<T>(object: Record<string, T>): Array<[string, T]> {
...
-];
-type ObjectEntries<T extends Record<PropertyKey, unknown>> = Array<
-  ObjectEntry<T>
->;
-
 export namespace toPairs {
   export function strict<T extends Record<PropertyKey, unknown>>(
     object: T
-  ): ObjectEntries<T> {
+  ): Array<NonNullable<{ [K in keyof T]: [K, T[K]] }[keyof T]>> {
```
eranhirsch, line 20: What's the purpose of the NonNullable here? It doesn't seem like the tuple would ever be nullable.

eranhirsch (changes requested):
There's something weird with the typing for optional.

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgZVGANsAZgTzgXziyghDgCIYcwBTAWixoGcZyBuAKA6trgAUAhsChMAPABUAfHAC8cAIJQoAnKIByEAHZqArunQCARuhqikAbQDScYJrgBrGjghY44gLoAuOFYA0bq3d3AnNHZ1cPSUkuHho4AHVZfiERMzgBAH5vFihbAHM2OEMsuE0dEEMaKAJJTgB6OrgmgD0MrhoAD0hYOFi4AHlDACsaAGMYSycxcThOmBpNABMmOAhhsZhpOQADABIEAFEO0fQdRdMwlzd-JhwKiHRJfG3OPsGR8YA1AVPmCVmOvMlisAEpjaCLUR8Ei0WA4SY4fw6TT2TQQADumiisg4TVQ4EwuFEYIAjjphDRIVJzO8NgjppJ3LVuNQ4rTxgdNDAoKoZnMFss4GDRhCoTCqlQEUiUWjMdi5OZcQN1uN6RJJL4leyYN9fgzNe5XqzlR8YJzucA-nzAQLQeCoJDoRBYZKnNLURisVsFEoVKIlfiMNhVNrzTz1dFmSzePEAExJUNc3J-JCZbIWzQFIolMoVKo1eqNFoZIA

I think what you'd want to do is first run Required on the object, and then run your logic.

cloud-walker:
> Hey, after thinking about this PR all night, I think the correct typing for the pairs is:
> 
> ```ts
> type Pairs<T> = Array<{ [K in keyof T]-?: [key: K, value: Required<T>[K]] }[keyof T]>
> ```
> 
> We are doing two things differently here; the first is that we treat the props in T as if they are non-optional; this is because, as an array of entries, the optionality comes from the fact that a pair is present or not, and not from the intrinsic value of a pair, the pair itself doesn't know about its optionality.
> 
> The other thing is that we need to remove the optionality (`-?:`) from the mapping itself because when you do `[K in keyof T]: ... ` If K is optional, it would also be optional in the mapping, and that's not what we want here (this is what introduces the `undefined` to the pair.

Nice thought! Pushed the Pairs<T> type, do you think we should expose it? Or leave it private? 🤔

eranhirsch (approved):
Looks great! Thanks for handling this!

You didn't touch any of the tests, but this change would require changes to at least one of the tests ("stricter typing"). Can you change that test to include a stricter check for the exact shape of pairs?

Also, that test uses `assertType`, which does the equivalent of an `expectTypeOf(x).toMatchTypeOf<<TYPE>()`, which is a broader check that `typeof x extends <TYPE>`, but what we want to test here is an equality test (using `expectTypeOf(x).toEqualTypeOf<<TYPE>>()`), this is the reason the existing test works right now although the type it tests is broader than the type we actually produce.

cloud-walker:
Tests reworked with the describe('typing') thing discussed elsewhere, I've also separated the toPairs tests from the toPairs.strict ones.

src/toPairs.ts
```diff
@@ -16,22 +14,14 @@ export function toPairs<T>(object: Record<string, T>): Array<[string, T]> {
...
-];
-type ObjectEntries<T extends Record<PropertyKey, unknown>> = Array<
-  ObjectEntry<T>
+type Pairs<T> = Array<
+  { [K in keyof T]-?: [key: K, value: Required<T>[K]] }[keyof T]
 >;
 
 export namespace toPairs {
-  export function strict<T extends Record<PropertyKey, unknown>>(
+  export function strict<const T extends Record<PropertyKey, unknown>>(
```
eranhirsch, line 22: With the new Pairs implementation, we don't need the narrowing to only `Record<PropertyKey, unknown>>`, it would work on any T. Also, the `const` part would not be supported in previous typescript versions we still want to support. Is it required to make the typing work?
```suggestion
  export function strict<T>(
```

eranhirsch (changes requested):

src/toPairs.ts
cloud-walker: I'll check on that

cloud-walker: Removing the `const` part means that I have to update a test adding `as const` to the object, I dont think is too much of an issue 🤔 

Do we have any part of the docs saying the minimun ts version supported?

eranhirsch: Not the docs because we never made the decision public, but due to existing code and our desire to support as many use cases as possible, our effective maximum typescript version is 4.4 for any public facing api (although internal code can use the latest typescript as long as that isn't exported). We have tests to validate this is the case and we don't ship unsupported code. 

Check out our ci github action for more context.

cloud-walker: Removing the narrowing we have a new ts error to register to ts-expect-error (TS2769) because the original `Object.entries` expect a Record:

image

Do you think it's ok?

eranhirsch: It's just because of the nullabe/undefined cases, use:
```ts
  export function strict<T extends NonNullable<unknown>>(obj: T): Pairs<T> {
```

cloud-walker:
Ok, @eranhirsch I think the PR is finally ready 😄

eranhirsch (approved):
🥳 Thanks!

Issue #408, opened 2023-11-20
point free pick / omit types are not working


cloud-walker:
Using pick or omit functions in a pipe for example, does not autocomplete the object keys coming from the pipe (and does not type check them).

```ts
import {pipe, omit} from 'remeda'

pipe({foo: 'hello', bar: 123}, omit(['asdf'])) // this should give a ts error
```

I've tried various remeda versions from the 1.29 to the 1.20 without luck 😢 

Here a repro: https://codesandbox.io/s/remeda-point-free-pick-omit-types-jx9m4j?file=/index.ts

eranhirsch:
Hey, this might be the case, we did some changes a few months ago to the dataLast implementation of these functions and we observed some issues (which were already fixed for `prop`). We might need to replicate those fixes to other functions.

If you want to take a stab at it i'd be happy to review as my bandwidth is limited recently. If not, i'll try and fix it myself soon(ish)

cloud-walker:
Sure! You can you share some information to start with? (Like the `prop` fix PR)

PR #411, opened 2023-11-24
type(omit|pick): Improve typings

cloud-walker:
Fixes: #408 

---

Make sure that you:

---

### Available Types:

- `feat`: new functions!

existing).

</details>


src/_types.ts
```diff
@@ -6,6 +6,15 @@ export type PredIndexedOptional<T, K> = (
...
 
+/**
+ * This helper type is useful to unwrap complex types exposing the underlying
+ * bare type. For example `Omit<{a: number, b: string}, 'a'>` becomes
+ * `{b: string}` directly.
+ */
+export type Unwrap<T> = {
+  [K in keyof T]: T[K];
+} & NonNullable<unknown>;
+
```
eranhirsch, line 17: It looks like you are implementing type-fest's Simplify. https://github.com/sindresorhus/type-fest/blob/main/source/simplify.d.ts

I'd keep the same name and maybe add a link to that project so we can sync any updates that are done to it from that project in the future.
```suggestion
/**
 * This helper type is useful to unwrap complex types exposing the underlying
 * bare type. For example `Omit<{a: number, b: string}, 'a'>` becomes
 * `{b: string}` directly.
 */
export type Simplify<T> = {
  [K in keyof T]: T[K];
} & NonNullable<unknown>;

```

src/pick.ts
```diff
@@ -26,17 +27,17 @@ export function pick<T extends object, K extends keyof T>(
...
   names: ReadonlyArray<K>
-): <T extends Record<PropertyKey, any>>(object: T) => Pick<T, K>;
+): <T extends Record<K, unknown>>(object: T) => Unwrap<Pick<T, K>>;
 
 export function pick() {
   return purry(_pick, arguments);
 }
 
-function _pick(object: any, names: Array<string>) {
+function _pick(object: Record<string, unknown>, names: Array<string>) {
```
eranhirsch, line 36: shouldn't this be PropertyKey (it doesn't matter much, just for correctness)
```suggestion
function _pick(object: Readonly<Record<PropertyKey, unknown>>, names: ReadonlyArray<PropertyKey>) {
```

src/omit.test.ts
```diff
@@ -12,13 +12,31 @@ describe('data first', () => {
     const result = omit(obj, ['a']);
     expect(result).toEqual({});
   });
+
+  test('non existing prop', () => {
+    // @ts-expect-error -- should not allow non existing props
+    const result = omit({ a: 1, b: 2, c: 3, d: 4 }, ['not', 'in'] as const);
+    expect(result).toEqual({ a: 1, b: 2, c: 3, d: 4 });
+  });
```
eranhirsch, line 20: 1. In most of our new tests i've split type tests from regular tests, that way you can focus on the thing you are changing when running tests.

Can you create a new top-level "describe" category for "typing" and put the tests there?

2. You also don't need to *also* test the runtime behavior, as those are 2 different concerns of the test. Here for example there's no point in testing the runtime behavior because the typing error would prevent this code from running.

3. Add the specific typescript error (tsXXXX) we are suppressing here, so that if typescript make changes and that error stops firing, we can debug what the original error was (similar to how you'd add the error message to a `expect(...).toThrow(<ERROR_MESSAGE>)` instead of just `expect(...).toThrow()`

src/omit.test.ts
```diff
@@ -12,13 +12,31 @@ describe('data first', () => {
...
 });
 
 describe('data last', () => {
   test('omit', () => {
-    const result = pipe({ a: 1, b: 2, c: 3, d: 4 }, omit(['a', 'd'] as const));
+    const result = pipe(
+      { a: 1, b: 2, c: 3, d: 4 },
+      omit(['a', 'd', 'a'] as const)
```
eranhirsch, line 27: Why did you add the additional 'a' here? If this is to test something new can you create a separate test for it?

src/omit.ts
```diff
@@ -12,10 +13,10 @@ import { purry } from './purry';
  * @dataFirst
  * @category Object
  */
-export function omit<T extends object, K extends keyof T>(
+export function omit<T extends Record<PropertyKey, unknown>, K extends keyof T>(
   data: T,
   propNames: ReadonlyArray<K>
-): Omit<T, K>;
+): Unwrap<Omit<T, K>>;
```
eranhirsch, line 19: Was this type broken? I thought we are only encountering issues with the dataLast implementation?

eranhirsch (changes requested):
This great solution works for 99% of the cases, but narrowing the typing to `Record<K, unknown>` takes away slightly from the broadest typing we can offer, and I'm not sure we have to.

The original problem is that nothing in the dataLast implementation is tying the `T` in the curried result to the `K` from the input. Your solution makes `T` derived from `K`. The other way to solve this is to make `T` part of the original signature and have `K` be derived *from it*. The problem with the latter is that you must have all data up-front. Doing this is fine for pipes but not for the pickFactory (or omitFactory) style functions (`const fooPicker = pick(['foo']);`). I don't think there are real-life cases where these functions are needed, and I prefer to **optimize for the pipes**.

In addition, we don't need the Unwrap around these types. Pick and Omit are built-in types that are easy to grasp and short enough to make sense of via tooltips. Adding the unwrap would require typescript to do more processing, which in a big project could add up to make the experience sluggish. Similar concerns have surfaced in projects like Zod and Prisma, where excessive inference causes typescript to stall. It's not relevant to this issue, and it should be reverted.

```ts
export function omit<T, K extends keyof T = keyof T>(
  propNames: ReadonlyArray<K>,
): (data: T) => Omit<T, K>
```

src/omit.test.ts
cloud-walker: I think it is just a typo

src/omit.ts
cloud-walker: I just watched the `object` type and auto updated it 😅

src/_types.ts
cloud-walker: I removed the type altogether

eranhirsch (changes requested):
Thanks for addressing my comments, but you didn't address the most important one, where T should be defined at the top level of the function, and not just on the returned curried function.

cloud-walker:
Sorry! Totally miss the T on the top level function thing, I'm working on it 👍🏻

cloud-walker:
I'm not totally sure that the T on top level function is what we want, adding the proposed override signature I get the following test error:

image

But in this case we know the object we are passing to `omitFoo()`, so I'm not super convinced that we should not support that, giving a more opaque type in this case 🤔 

Also, can you provide an example in which the proposed signature (the one with T on the inner function), will work not as expected? I can't currently see the problem we are trying to solve 🤔

eranhirsch:
I don't think the omitFoo capability has any real-life use-case, so we don't need to make it work.

My concern is the typing of Record<K, unknown>, which would only accept simple/plain records but would not work once the types become more complicated. When T is available upfront, typescript has more in its inference engine to work with.

Here's an example:
https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzmYYCmcC+cBmUIhwBEUaIaAJgIZEDcAUPTAJ7pwCyzAKqxgLyI4VAFxwAzjCjAAdgHMscAD6CqAflESpc2nABGo6QFcQutFCwN6AYwjSJcAB5wBSEXADkACzQAbHxA9sKjEObl5LNAdIWFxDaSsYYFs4AmAYAEYAHgBpOEiYNGkKEIAFfHRYZmy0ZgA+AAp6OGRygDkqcjFRACU0KgpbH2YAQSgoKmYc2voASlFMrjyHAqKQ3psoChyAGjg4gGtpCAB3aVqG6hg3LhnnWrgAeRA0hd3s+4QmuAB6b7gKYwgZhwUBgHy7Y4YQbAqxUUhCXQQQzwGDeOAsdBiL6kGCGKDSOD1S5UW58e7EhiYRiRaLwHBxBJJAmpGAAJlecFy+UKxTg+xqEBwcC4DS+YDaHTQXTgvX60NG40m71moiJVCuohud0ezxgHPeiC+v3+gOBoPBcEhKWkQzgsPhVERyPRaIxUuxaFx+MJxNJ5PVVEpjGstnsSIy6WcKV16XqAG13FR3ABdGb1BwzBg2OzwcPpVlRlDodO7FmxhNJ1OZkM5lLI1mRgQs1nxxMptMZrOh3P1gsCItoEvRtItivtzNAA

src/omit.ts
```diff
@@ -28,15 +28,15 @@ export function omit<T extends object, K extends keyof T>(
  * @dataLast
  * @category Object
  */
-export function omit<K extends PropertyKey>(
+export function omit<T extends Record<PropertyKey, unknown>, K extends keyof T>(
```
eranhirsch, line 31: We don't need to limit T to Records, this is only needed if you want to maintain typing for the omitFoo factory that we don't want to maintain.

```suggestion
export function omit<T, K extends keyof T>(
```

src/omit.ts
```diff
@@ -28,15 +28,15 @@ export function omit<T extends object, K extends keyof T>(
...
   propNames: ReadonlyArray<K>
-): <T extends object>(data: T) => Omit<T, K>;
+): (data: T) => Omit<T, K>;
 
 export function omit() {
   return purry(_omit, arguments);
 }
 
-function _omit<T extends object, K extends keyof T>(
+function _omit<T extends Record<PropertyKey, unknown>, K extends keyof T>(
```
eranhirsch, line 39: Similarly
```suggestion
function _omit<T, K extends keyof T>(
```

src/omit.test.ts
```diff
@@ -26,5 +26,42 @@ test('type for curried form', () => {
 
   const result = omitFoo({ foo: 1, bar: 'potato' });
 
-  expectTypeOf(result).toEqualTypeOf<{ bar: string }>();
+  expectTypeOf(result).toEqualTypeOf<
+    Omit<Record<PropertyKey, unknown>, 'foo'>
+  >();
```
eranhirsch, line 31: Let's drop the whole test, it doesn't test anything that is meaningful for us to maintain.

src/pick.test.ts
```diff
@@ -42,14 +42,45 @@ test('read only', () => {
...
 
-  expectTypeOf(true as any as ReturnType<typeof pickFoo>).toEqualTypeOf<
-    Record<'foo', any>
-  >();
+    const result = pickFoo({ foo: 1, bar: 'potato' });
 
-  const result = pickFoo({ foo: 1, bar: 'potato' });
+    expectTypeOf(result).toEqualTypeOf<{ foo: number }>();
+    expect(result).toEqual({ foo: 1 });
+  });
```
eranhirsch, line 57: Like in the omit case, we don't need to maintain the 'pickFoo' case so we don't need a test for it

eranhirsch (changes requested):
The pick implementation doesn't have the fixes we discussed for the omit implementation.

src/omit.ts
cloud-walker: I need at least an `object` type or the implementation code will yell at me:

image

cloud-walker:
I've left at least the `extends object`, for the reasons described in the comments

eranhirsch (approved):
Thanks for the fixes! This bothered me in one my projects and I never got around to fixing it

🥳

PR #413, opened 2023-11-29
fix(toPairs): Fix `toPairs` to accept interface types

RedHatter:
The current version does not allow interface types.
```
interface Example {
  b: string
}

const a: Example = { b: 'c' }

R.toPairs(a)
```
This PR widens the type to allow any object not just `Record`s.

I was unsure how to type the current strict version however it should be easy after #410 is merged. 

---

Make sure that you:

---

### Available Types:

- `feat`: new functions!

existing).

</details>

### DIFF
--- a/src/toPairs.test.ts
+++ b/src/toPairs.test.ts
@@ -9,35 +9,47 @@ test('should return pairs', () => {
   ]);
 });
 
-test('should return pairs, strict', () => {
-  const actual = toPairs.strict({ a: 1, b: 2, c: 3 });
-  expect(actual).toEqual([
-    ['a', 1],
-    ['b', 2],
-    ['c', 3],
-  ]);
+it('should accept interface types', () => {
+  interface T {
+    a: string;
+  }
+
+  expectTypeOf(toPairs<T>({ a: 'b' })).toEqualTypeOf<Array<[string, string]>>();
 });
 
-test('stricter typing', () => {
-  const actual = toPairs.strict({ a: 1, b: 2, c: 3 } as const);
-  assertType<Array<['a' | 'b' | 'c', 1 | 2 | 3]>>(actual);
-});
+describe('stricter', () => {
+  test('should return pairs', () => {
+    const actual = toPairs.strict({ a: 1, b: 2, c: 3 });
+    expect(actual).toEqual([
+      ['a', 1],
+      ['b', 2],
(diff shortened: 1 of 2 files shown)


eranhirsch:
This looks and works great, but the non-strict version of toPairs should be kept as bare to the core typing as possible. The strict version is there for anyone who needs better typing, but it isn't without a caveat that the typing might be wrong, and you can run into issues. We should change the typing for the non-strict version to make it accept interfaces, but the return value you suggest is incorrect in that context.

The non-strict version of toPairs is basically `Object.entries()` and is redundant and could be replace with it. Looking at the typing of Object.entries, typescript handle 2 cases:
```ts
   /**
     * Returns an array of key/values of the enumerable properties of an object
     * @param o Object that contains the properties and methods. This can be an object that you created or an existing Document Object Model (DOM) object.
     */
    entries<T>(o: { [s: string]: T } | ArrayLike<T>): [string, T][];

    /**
     * Returns an array of key/values of the enumerable properties of an object
     * @param o Object that contains the properties and methods. This can be an object that you created or an existing Document Object Model (DOM) object.
     */
    entries(o: {}): [string, any][];
```

This means that for interfaces, the return value is `Array<[string, any]>`. This kinda makes sense because typescript is duck-typed, so I can build an object:
```ts
const myObject: MyInterace & { notPartOfTheInterface: UniqueTypeThatIsntInMyInterface } = ...
const pairs = Object.entries(myObject as MyInterface);
//    ^? [string, Required<MyInterface>[keyof MyInterface]][]
```
And pairs would contain a pair `['notPartOfTheInterface', UniqueTypeThatIsntInMyInterface]` which wouldn't be typed correctly by `[string, Required<MyInterface>[keyof MyInterface]]`. Because of duck-typing typescript uses `[string, any]`.

(this isn't the only reason, interfaces are prone to https://www.typescriptlang.org/docs/handbook/declaration-merging.html, so typescript can't even say for sure how the type would end up looking).

TL;DR - I think the current type could be updated to mirror `Object.entries` type (accepting interfaces via `{}`), but not the type you suggested here.

eranhirsch:
@RedHatter, are you planning on continuing to work on this?

eranhirsch:
I'm closing this to keep our queue clean. Feel free to re-open once you get back to it, or if anyone else wants to pick this up.

PR #421, opened 2023-12-05
feat(debounce): add debounce function

### LINKED ISSUE #419: `debounce` utils
For example, I'd like to delay state update when input value changes,
have any plans to implement debounce utils?

eranhirsch:
Closes #419

Inspired by the lodash implementation (https://github.com/lodash/lodash/blob/main/src/debounce.ts).

Tests were copied from lodash too (https://github.com/lodash/lodash/blob/main/test/debounce.spec.js), but the latest version of their test has some issues, and I'm not sure their tests actually run (I didn't try).

---

Make sure that you:

---

### Available Types:

- `feat`: new functions!

existing).

</details>

### DIFF
--- a/src/debounce.ts
+++ b/src/debounce.ts
@@ -1,29 +1,41 @@
-type SimplifiedFunction<Params extends ReadonlyArray<any>, Return> = (
-  ...args: Params
-) => Return;
-
-type Debounced<Params extends ReadonlyArray<any>, Return> = {
-  readonly call: SimplifiedFunction<Params, Return | undefined>;
+type Debounced<F extends (...args: any) => unknown> = {
+  readonly call: (...args: Parameters<F>) => ReturnType<F> | undefined;
   readonly cancel: () => void;
-  readonly flush: () => Return | undefined;
+  readonly flush: () => ReturnType<F> | undefined;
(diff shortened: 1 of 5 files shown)


src/debounce.ts
```diff
@@ -0,0 +1,264 @@
...
+ * the exact same signature.
+ * @param options An object allowing further customization of the debouncer:
+ * - `timing?: 'leading' | 'trailing' |'both'`. The default is `'trailing'`.
+ *   `leading` would result in the function being invoked at the start of the
+ *   cool-down period; `trailing` would result in the function being invoked at
+ *   the end of the cool-down period (using the args from the last call to the
+ *   debouncer). When `both` is selected the `trailing` invocation would only
+ *   take place if there were more than one call to the debouncer during the
+ *   cool-down period. **DEFAULT: 'trailing'**
+ * - `waitMs?: number`. The length of the cool-down period in milliseconds. The
```
cjquines, line 63: **suggestion:** put `waitMs` as a second argument, not in options?

it is overwhelmingly the case that the wait is the second argument across implementations, and it'd be strange to change that convention imo

TkDodo (approved):

Issue #414, opened 2023-11-30
`isObject` documentation is slightly off


stwlam:
The current doc for `isObject` is:

> A function that checks if the passed parameter is of type Object and narrows its type accordingly

This isn't quite right:
- `isObject([])` returns `false` but is an `Object`
- `isObject(Object.create(null))` returns `true` but isn't an `Object`

Accounting for those could make the documentation a little verbose, but it is perhaps at least important to call out that an array will not be counted among "objects".

eranhirsch:
Actually going over Lodash and the PRs that introduced this guard, it wasn't clear to me that the exclusion of arrays was intentional. I think it just made sense at the time. Lodash has 2 guards, one for just the type-check (`isObjectLike`) and one for simple Records (`isPlainObject`).

I've sent a PR that takes a similar approach and introduces 2 new guards, and deprecates this one instead.

Thanks for pointing this out!

stwlam:
Thanks for looking into this! I suppose what originally caught my eye was that `Object` is capitalized in the docs, seemingly implying that `isObject(x)` is true if and only if `x instanceof Object`. Instead it guarantees that `typeof x === "object"` and is neither an array nor equal to `null`.

PR #436, opened 2023-12-07
feat(isObject): Replaced with isObjectType and isPlainObject

eranhirsch:
Fixes: #414

---

Make sure that you:

---

### Available Types:

- `feat`: new functions!

existing).

</details>


src/isObjectType.ts
```diff
@@ -0,0 +1,25 @@
+import type { NarrowedTo } from './_types';
+
+/**
+ * Checks if the runtime `typeof` of the param is `object` and not null.
+ *
+ * Notice that a lot of things in JS are objects, pretty much anything that isn't a primitive type This includes Arrays, Classes, RegExps, Maps, Sets, Dates, URLs, Errors, etc...
```
cjquines, line 6: **nit:**

```suggestion
 * Note that a lot of things in JS are objects; pretty much anything that isn't a primitive type. This includes Arrays, Classes, RegExps, Maps, Sets, Dates, URLs, Errors, etc.
```

src/isObjectType.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Use `isPlainObject` for a narrower check that only accepts simple struct/shape/record-like objects.
+ *
+ * @param data the variable to check
+ * @signature
+ *    R.isObjectType(data)
+ * @returns true if the passed input is an Object, Promise, Date or Error, false otherwise
```
cjquines, line 13: **question:** is this correct?

src/isPlainObject.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Checks if the param is a simple struct/shape/record-like object, that is, an object with keys and values. The values themselves can be anything, including other objects, functions, classes, etc.
+ *
+ * Use `isObjectType` for a wider check that accepts anything JS considers an object
+ *
+ * @param data the variable to check
+ * @signature
+ *    R.isPlainObject(data)
+ * @returns true if the passed input is an Object, Promise, Date or Error, false otherwise
```
cjquines, line 12: **todo:** this needs to be changed

src/isObjectType.ts
eranhirsch: All of these are objects in javascript.

mapping.md
```diff
@@ -40,6 +40,8 @@ documentation when migrating._
 | `invert`            | `invert`            | `invertObj`         |
 | `isEmpty`           | `isEmpty`           | `isEmpty`           |
 | `isNil`             | `isNil`             | `isNil`             |
+| `isObjectType`      | `isObjectLike`      |                     |
```
TkDodo, line 43: question: why would we opt for `isObjectType` rather than `isObjectLike` ?

eranhirsch: I would call it `isTypeOfObject` because I want names to be as close to their ECMAScript semantics, but I'm afraid that people would have difficulty finding it because they'd start typing `isObject...` and look for it via autocomplete.

I don't understand the semantics of "Like" in the context of `isObjectLike`, is a `RegExp` object-like? Is a `Class` with only `static` methods object-like? Because `typeof === "object"` is kinda the fallback in JS for anything that isn't something else, I feel that likeness to an object is not the key attribute being checked here...

eranhirsch: We need to add fuzzy search in the documentation site, and then add the mapping/aliases to the index so that people can search for different kinds of names coming from different other libraries and still find what they are looking for.

stwlam:
How does it cope with something like the following?

```ts
const arr: Record<number, string> = ["a", "b", "c"];
if (R.isPlainObject(arr)) {
  console.log("a plain object"); // Probably not this
} else if (Array.isArray(arr)) {
  console.log("an array");  // Probably this
}
```

eranhirsch:
I'm not sure I understand the question, the prototype of `["a", "b", "c"]` is not `Object.prototype` or `null` so it would return false on `isPlainObject`

stwlam:
> I'm not sure I understand the question, the prototype of `["a", "b", "c"]` is not `Object.prototype` or `null` so it would return false on `isPlainObject`

The guard seems like it would exclude the tested variable from being a `Record`, which seems like it would become `never` after that.

eranhirsch:
I don't understand what you mean. Can you provide a specific bug where the typing wasn't correct?

The guard looks to be working as expected: https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcByBDKUIHdgAmAKhADxEA0cAQhgM7AB8cAvHAKIgxQYDGM5KrQbNQMYADsCdOBOAA3YFABQcOAH4a9YKrgAuDlx79BWkQG5ly0JFhwAZgFcJ-AJYQJcV3QAKAGwxXCQB5ACMAK2ATIkYACl0CDBgMAyI4AB84ACUo6AJSHxwUWEQAaWBEKmcAawk8CUZlAEoDROSvGUxsPEISUxzePIKipSRyyrgautwG5gBvXVd7OFikFAhltow4AEIWNgByCAiomAOMzK3WfdlHPz8muAW1NShgGEcoT3sMPwZLNQAXysakGEjo8DAOBgEFYcDCkX4ADoAObvQoQGFrYDBeyxLZNAFwN4fL5wKGY2H7NgSO5+C7k6FUm4I05IilY5DASzA5RgiFwLBQAwDIa0gC2oSUVAhUCCKOYbAA2gAiDAqqgq0IauAq3gqgC6liWK28-kCIRO-FiQqaj2ecH5ED8wCRfggKJt2EJcAA9L64BjQhhQn5ELJMQgABbeXT+l4Jl4APXUykBcGAf1QJtiAEFsBhEEjvPmeIgvVA7U9dE6XW6PRWfX6A0GQ2Ho7G1PHEwmU2mgA

The problem is not with our guard; it's with `Array.isArray` cannot narrow a Record<number, string> to an array; that's outside the scope of this change.

stwlam:
It is true that not all `Record<number, T>`s are `Array<T>`s, but it is also the case that all `Array<T>`s are `Record<number, T>`s. The `isPlainObject` return predicate isn't stating some variable is a "plain object" in some sense: it's reporting that the passed argument is a `Record`. Many non-plain objects are also `Record`s, so the guard is prone to returning false negatives. As far as I'm aware, typescript has no type for a plain object, so any implementation wouldn't be to accurately type-guard.

eranhirsch:
You are looking at it from the wrong side. isPlainObject is not called isRecord on purpose. It doesn't try to detect anything that could implement a Record interface; it narrows down whatever **is** a plain object, and those objects should always extend a `Record<PropertyKey, unknown>`. I couldn't find any example of a plain object not also being some sort of Record.

The runtime definition of a plain object is not new; we copied it as-is from lodash.

I'm still trying to understand what bug you are reporting; it would be helpful if you could draft a test case showing the bug, and if you have suggestions for fixes to the types, that would be great, too.

TkDodo (approved):

PR #437, opened 2023-12-10
impl

eranhirsch:

---

Make sure that you:

---

### Available Types:

- `feat`: new functions!

existing).

</details>

### DIFF
--- a/src/_purryOrderRules.ts
+++ b/src/_purryOrderRules.ts
@@ -1,0 +1,113 @@
+import type { NonEmptyArray } from './_types';
+
+export type OrderRule<T> = OrderProjection<T> | OrderPair<T>;
+
+type OrderPair<T> = readonly [
+  projector: OrderProjection<T>,
+  direction: Direction,
+];
+
+type OrderProjection<T> = (x: T) => Comparable;
+
+type Comparable = ComparablePrimitive | { valueOf(): ComparablePrimitive };
+type ComparablePrimitive = number | string | boolean;
+
+const ALL_DIRECTIONS = ['asc', 'desc'] as const;
+type Direction = (typeof ALL_DIRECTIONS)[number];
+
+const COMPARATOR = {
+  asc: <T>(x: T, y: T) => x > y,
+  desc: <T>(x: T, y: T) => x < y,
+} as const;
+
+/**
+ * Allows functions that want to handle a variadic number of order rules a
+ * a simplified API that hides most of the implementation details. The only
+ * thing users of this function need to do is provide a function that would take
+ * the data, and a compare function that can be used to determine the order
+ * between the items of the array.
+ * This functions takes care of the rest; it will parse rules, built the
+ * comparer, and manage the purrying of the input arguments.
+ */
+export function purryOrderRules<T>(
+  func: (data: ReadonlyArray<T>, compareFn: (a: T, b: T) => number) => unknown,
+  inputArgs: IArguments
+): unknown {
+  // We rely on casting blindly here, but we rely on casting blindly everywhere
+  // else when we call purry so it's fine...
+  const [dataOrRule, ...rules] = Array.from(inputArgs) as
+    | [data: ReadonlyArray<T>, ...rules: Readonly<NonEmptyArray<OrderRule<T>>>]
+    | Readonly<NonEmptyArray<OrderRule<T>>>;
+
+  if (!isOrderRule(dataOrRule)) {
+    // dataFirst!
+
+    // @ts-expect-error [ts2556]: Typescript is failing to infer the type of rules
+    // correctly here after the type refinement above, rules should be non-empty
+    // when we get here.
+    const compareFn = orderRuleComparer(...rules);
+    return func(dataOrRule, compareFn);
+  }
+
+  // dataLast!
+
+  // Important: initialize the comparer outside of the returned function so it
+  // it's constructed and shared everywhere (it's stateless so should be safe
+  // if used multiple times).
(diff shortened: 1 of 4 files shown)


src/_heap.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Heap related utilities.
+ */
+
+/**
+ * Turn an array into a **MAX**-heap using `compareFn`
+ */
+export function heapify<T>(heap: Array<T>, compareFn: (a: T, b: T) => number) {
```
cjquines, line 8: **nit:** explicitly type return? (or not)

src/_purryOrderRules.ts
```diff
@@ -0,0 +1,147 @@
...
+  direction: Direction,
+];
+
+type OrderProjection<T> = (x: T) => Comparable;
+
+type Comparable = ComparablePrimitive | { valueOf(): ComparablePrimitive };
+type ComparablePrimitive = number | string | boolean;
+
+const ALL_DIRECTIONS = ['asc', 'desc'] as const;
+type Direction = (typeof ALL_DIRECTIONS)[number];
```
cjquines, line 17: **question:** is there a reason it's in this order?

i found it a bit confusing to read for the first time, maybe reordering would help (similar to the order in old `sortBy`)

(comments might be overkill, but whatever)

```ts
type ComparablePrimitive = number | string | boolean;

/** A type that we can compare with < or >. */
type Comparable = ComparablePrimitive | { valueOf(): ComparablePrimitive };

type OrderProjection<T> = (x: T) => Comparable;

const ALL_DIRECTIONS = ['asc', 'desc'] as const;
type Direction = (typeof ALL_DIRECTIONS)[number];

type OrderPair<T> = readonly [
  projector: OrderProjection<T>,
  direction: Direction,
];

/** A function, or a function and a direction, used to give a key to order by. */
export type OrderRule<T> = OrderProjection<T> | OrderPair<T>;

export type CompareFunction<T> = (a: T, b: T) => number;
```

src/_purryOrderRules.ts
```diff
@@ -0,0 +1,147 @@
...
+
+const COMPARATOR = {
+  asc: <T>(x: T, y: T) => x > y,
```
cjquines, line 20: **nit:** `T extends Comparable`?

src/_purryOrderRules.ts
```diff
@@ -0,0 +1,147 @@
...
+  const comparator = COMPARATOR[direction];
+
+  const nextComparer =
+    secondaryRule === undefined
+      ? undefined
+      : orderRuleComparer(secondaryRule, ...otherRules);
+
+  return (a, b) => {
+    const projectedA = projector(a);
+    const projectedB = projector(b);
```
cjquines, line 113: **question:** this means we may invoke `projector` more than once per element, right?

the guarantee that `projector` is invoked at most once per element isn't necessary, and it's not even something most libraries give... the only one i can think of that guarantees this is [deno std](https://deno.land/std@0.208.0/collections/mod.ts?s=sortBy). so not necessary

src/_purryOrderRules.ts
eranhirsch: I usually put exported stuff at the top and implementation details below (I also do this with functions). That way, people have to scroll down less to see the primary focus of a file. Anything that isn't exported is implementation details that shouldn't matter to anyone except maintainers.

src/_purryOrderRules.ts
eranhirsch: If typescript is happy, I'm happy. It's best not to limit the implementation if it doesn't need to be limited; that way, a future change that would want to use these comparators with something other than Comparable would be able to do so without needing to change this code.

src/_purryOrderRules.ts
eranhirsch: It's not the job of this utility to optimize the projector function. The user can provide a function that also caches its results if they want to. If we did this automatically without the user's control, we could get to a situation where a simple projection (like `identity`) would result in doubling the array's memory footprint; if you are sorting a huge array, that is just redundant memory.

src/_purryOrderRules.ts
```diff
@@ -0,0 +1,175 @@
...
+    | [data: ReadonlyArray<T>, ...rules: Readonly<NonEmptyArray<OrderRule<T>>>]
+    | Readonly<NonEmptyArray<OrderRule<T>>>;
+
+  if (!isOrderRule<T>(dataOrRule)) {
+    // dataFirst!
+
+    // @ts-expect-error [ts2556]: Typescript is failing to infer the type of rules
+    // correctly here after the type refinement above, rules should be non-empty
+    // when we get here.
+    const compareFn = orderRuleComparer(...rules);
```
TkDodo, line 78: last time we shipped something with `ts-expect-error`, it broke someone's code because they were checking lib code and the comments were stripped. I'd prefer if we'd do `as any` instead ?

TkDodo (approved):
amazing work 👏

src/_purryOrderRules.ts
eranhirsch: That's what our typescript tests are for; they check that the typescript lib files compile without error, too. This code would be compiled to JS and only live as JS code in our distribution, where ts errors don't matter.

TkDodo: just fyi, it was this issue:

- https://github.com/remeda/remeda/issues/163

not sure if it will come back ...

TkDodo: aah, this was about using ts-expect-error on interfaces that result in declaration, which is different

- https://github.com/remeda/remeda/issues/163#issuecomment-1071215533

PR #442, opened 2023-12-15
feat: Add takeLastWhile function

pekam:


src/takeLastWhile.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export function takeLastWhile<T>(
+  fn: (item: T) => boolean
+): (array: ReadonlyArray<T>) => Array<T>;
+
+export function takeLastWhile() {
+  return purry(_takeLastWhile, arguments);
+}
+
+function _takeLastWhile<T>(array: Array<T>, fn: (item: T) => boolean) {
```
pekam, line 39: I think the implementation could support lazy evaluation, but `takeWhile` doesn't support it either, so maybe that should be a separate task to keep these functions consistent.

src/takeLastWhile.ts
```diff
@@ -0,0 +1,46 @@
...
+  for (let i = array.length - 1; i >= 0; i--) {
+    if (!fn(array[i])) {
+      return array.slice(i + 1);
+    }
+  }
+  return array;
```
pekam, line 45: Returning the original array instead of making a copy is more efficient, but not sure if it would be better to return a new instance in all cases.

src/takeLastWhile.ts
eranhirsch, line 39: Prefer Readonly wherever possible, it's extra important in Remeda in params because we never mutate inputs.
```suggestion
function _takeLastWhile<T>(array: ReadonlyArray<T>, fn: (item: T) => boolean): Array<T> {
```

src/takeLastWhile.ts
eranhirsch, line 45: To stay consistent with the case where you don't fall through to the end return statement, you need to provide a new copy. Otherwise, we create different semantics for the other cases. So, if the user wants to mutate the result, they will know it's safe and that they aren't mutating the input array by accident just because the condition didn't return true.

btw, setting the input param to be Readonly would surface this return value as a type error.

```suggestion
  return [...array];
```

src/takeLastWhile.ts
eranhirsch: It can't; when running lazily we go over the items from start to end and need to make a decision based on the items we already saw. Here the first item only knows if it should be returned or dropped once a further item comes along and returns a false value for the predicate.

takeWhile can be lazy because items have all the info they need while processing.

src/takeLastWhile.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @param fn the predicate
+ * @signature
+ *    R.takeLastWhile(fn)(array)
+ * @example
+ *    R.pipe([1, 2, 3, 4, 3, 2, 1], R.takeLastWhile(x => x !== 4))  // => [3, 2, 1]
+ * @dataLast
+ * @category Array
+ */
+export function takeLastWhile<T>(
+  fn: (item: T) => boolean
```
eranhirsch, line 32: Use a more descriptive name: `predicate`
```suggestion
  predicate: (item: T) => boolean
```

src/takeLastWhile.ts
```diff
@@ -0,0 +1,46 @@
...
+): (array: ReadonlyArray<T>) => Array<T>;
```
eranhirsch, line 33: nit: prefer `data` instead of `array`
```suggestion
): (data: ReadonlyArray<T>) => Array<T>;
```

src/takeLastWhile.ts
```diff
@@ -0,0 +1,46 @@
...
+ *    R.takeLastWhile(array, fn)
+ * @example
+ *    R.takeLastWhile([1, 2, 3, 4, 3, 2, 1], x => x !== 4) // => [3, 2, 1]
+ * @dataFirst
+ * @category Array
+ */
+export function takeLastWhile<T>(
+  array: ReadonlyArray<T>,
+  fn: (item: T) => boolean
+): Array<T>;
```
eranhirsch, line 18: Prefer descriptive names
```suggestion
export function takeLastWhile<T>(
  data: ReadonlyArray<T>,
  predicate: (item: T) => boolean
): Array<T>;
```

src/takeLastWhile.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Returns elements from the end of the array until the predicate returns false.
```
eranhirsch, line 21: The description needs to be clearer about what order the returned items are in. It can be read as though they are returned in reverse order (as they are checked by the predicate).

src/takeLastWhile.ts
```diff
@@ -0,0 +1,46 @@
+import { purry } from './purry';
+
+/**
+ * Returns elements from the end of the array until the predicate returns false.
```
eranhirsch, line 4: See my comment about the description below

src/takeLastWhile.test.ts
```diff
@@ -0,0 +1,60 @@
+import { pipe } from './pipe';
+import { takeLastWhile } from './takeLastWhile';
+
+const DATA = [1, 2, 3, 4] as const;
```
eranhirsch, line 4: We don't need the DATA const; it makes tests easier to read when the data is inline.

eranhirsch (changes requested):
My only real concern is that we need to return a clone of the data for the trivial case.

src/takeLastWhile.ts
pekam: Ok :+1: My idea was to keep the naming and docs consistent with `takeWhile` (and most other functions).

Done.

src/takeLastWhile.ts
pekam: Good point. :+1: Done. I also changed the example to a bit more obvious one (the old example repeated the same numbers in different order).

src/takeLastWhile.ts
pekam: Done

src/takeLastWhile.ts
pekam: Done

src/takeLastWhile.ts
pekam: Done

src/takeLastWhile.ts
pekam: Yeah, makes sense (I was unsure about this and added a question about this with the PR).

Done.

src/takeLastWhile.ts
pekam: Done

src/takeLastWhile.test.ts
pekam: Done

src/takeLastWhile.ts
pekam: Ok. Interesting... I would imagine it's possible to do the same thing from end to start. But nevermind, I can look at the implementation if I ever feel like I want to understand how the laziness really works. :)

eranhirsch (approved):

eranhirsch:
Thanks for doing this!

pekam:
Thanks @eranhirsch!

PR #455, opened 2024-01-10
docs: Disable single line breaks in docs site

pekam:
So that JSDocs of functions can be formatted on multiple lines without affecting the docs site. Double line breaks will still be rendered as a new paragraph.

This is done by disabling the GFM (GitHub flavored markdown) style of line breaks when parsing the function descriptions with the 'marked' package.

### DIFF
--- a/docs/scripts/transform.ts
+++ b/docs/scripts/transform.ts
@@ -41,7 +41,6 @@ export type FunctionsData = ReadonlyArray<
 
 const MARKED_OPTIONS = {
   gfm: true,
-  breaks: true,
 } satisfies MarkedOptions;
 
 const PRETTIER_OPTIONS = {


pekam:
@eranhirsch I noticed that the issue we discussed in https://github.com/remeda/remeda/pull/445#discussion_r1445866083 is caused by this single option used in the documentation parsing.

If this is something that we want to commit, I can scan through all the JSDocs and try to figure out where single line breaks are actually intended to create a new paragraph, and change those to double line breaks. Or if you prefer to do it, let me know.

eranhirsch (approved):
amazing!

pekam:
Thanks @eranhirsch! Did you see my comment above? My intention was that if you approve the actual fix, I could adjust all the JSDocs in the same PR, because in some cases single line break is probably meant to create a new line. What do you think? Do you want me to do it in a new PR, or do you prefer to scan through the docs yourself? Or is this step even required in your opinion?

eranhirsch:
I didn't notice the comment. It's ok not to fix the previous blocks. If we encounter any broken docs, we can always send a specific fix for it.

pekam:
Ok :+1:

Issue #456, opened 2024-01-10
Add isSymbol


viveleroi:
We're hoping to replace lodash and ramda entirely with remeda. There are a few functions remeda doesn't have that I believe it could benefit from.

Remeda does not yet have `isSymbol` which we use from lodash. Admittedly `typeof value === 'symbol'` is probably enough to type narrow for most cases but it would fit nicely with the existing group of `is___` methods.

[Lodash docs](https://lodash.com/docs/4.17.15#isSymbol)
[Lodash source](https://github.com/lodash/lodash/blob/main/src/isSymbol.ts)

eranhirsch:
Sounds great! can you send a PR?

TkDodo:
fixed by #460

PR #460, opened 2024-01-12
feat: Adds isSymbol function

viveleroi:
Adds an `isSymbol` type guard function.  Addresses #456 

---

Make sure that you:


viveleroi:
Note: This is PR is incomplete as somethings keeps overwriting isDefined/isNotNull types on commit which causes tsc failures and I'm not immediately sure what.

viveleroi:
I would appreciate some input on what's triggering the automatic writing of the files I need to modify.

eranhirsch (changes requested):
The error you are seeing is expected. You added another type to `TestObj`, so you need to fix these tests to also consider these cases. Because `symbol` is not `undefined`, it needs to be added to the expected output of running `isDefined` on `TestObj` and similarly on `isNonNull`.

Other then that this PR is great

viveleroi:
> The error you are seeing is expected. You added another type to `TestObj`, so you need to fix these tests to also consider these cases. Because `symbol` is not `undefined`, it needs to be added to the expected output of running `isDefined` on `TestObj` and similarly on `isNonNull`.
> 
> Other then that this PR is great

Except I did that, and those test files get overwritten somehow during the pre-commit phase. I made all of the changes, went to commit, and they're rejected for missing the values you mentioned, and I opened the files and saw they had all vanished.

eranhirsch:
I don't know how your dev env is set up, i pushed a commit that fixes these issues for you. I'll approve and merge this PR now...

eranhirsch (approved):
Thanks!

viveleroi:
I cloned the repo, did npm install, editing the files in VSCode, and commit changes. I'm not making it up lol, every time the commit hooks ran these two specific files reset completely.

viveleroi:
Ok I figured it out.

1. Make the changes to isDefined.test.ts
2. Run `git status` to confirm the file shows as modified
3. Make one change to isNonNull.test.ts but not both, leaving a tsc error
4. Try to commit. The lint-staged process runs tsc and will error because a change was missed
5. Notice that isDefined.test.ts has been reset and no longer shows up in `git status`

Now, repeat the process but make all changes so there's no tsc error. Files are not reset.

eranhirsch:
I'm sorry the commit hooks made a mess for you; there's always a fear they'll add edge cases that make development harder.

viveleroi:
I would probably rework things to avoid writing files outside of prettier, something isn't working right. I've used all of those in other repos without issue so either there's an uncaught problem with the configuration or you're intentionally writing files and it's not documented in a developer readme.

eranhirsch:
The part that rewrote your file is lint-staged; it rewrites commits intentionally to automatically fix any fixable issues (eslint, prettier) without manual intervention (like how format-on-save in your IDE works). We run type-checking in lint-staged, so if your commit as a whole results in bad types, it will fail the commit. I don't have the exact flow thought out, but lint-staged can revert a commit, so that it probably what happened.

viveleroi:
Outside of `--fix` for eslint I have used all of these tools and have never seen files get completely reset like this. It's not lint-staged, and prettier would only format the file, not revert changes, even if the syntax was literally broken it would error and not write the file. We use prettier, eslint, tsc, husky + lint-staged in multiple apps and I've never seen this, so I'm not sure what's wrong but something is up. 

I'm brining this up so that someone can investigate - this needs its own I issue I suppose - but this shouldn't be accepted as par for the course.

eranhirsch:
What we have which might be different than regular set ups is typescript in eslint. Most configs would probably not have this, just prettier and eslint.

eranhirsch:
> but whoever setup the repo should probably look as they'd know best.

🤔

That would be me, and I have no idea. I would appreciate any help on this, but know that this is very lo pri so I would only recommend looking into it out of curiosity.

PR #463, opened 2024-01-14
feat(round, ceil, floor): Add round, ceil, floor functions

### LINKED ISSUE #462: Add round

I'm currently using `round` from lodash and would like to see it added. 

[Lodash docs](https://lodash.com/docs/4.17.15#round)
[Lodash source](https://github.com/lodash/lodash/blob/main/src/round.ts)

paolostyle:
Includes implementation of `round`, `ceil` and `floor` functions. I need these (well, mostly `round` but the other two are trivial to implement once you have one) to migrate off of `lodash`, hence this PR. 

The main difference between these and native functions from the `Math` object is the second argument `precision`. I implemented it in the simplest way possible which I believe is correct for most actual use cases. Still, there are certainly one-off rounding issues caused by floating-point nature of `number` type in the language. 

There's one pretty significant difference from `lodash` - `precision` is limited to values from -15 to 15 as higher precision doesn't really make any sense, for example, `Math.PI` has only 15 decimal places. Similar function implemented in C# native library has a similar validation rule (0 to 15, I believe negative precision is pretty useful in some scenarios), [link to their docs for reference](https://learn.microsoft.com/en-us/dotnet/api/system.math.round?view=net-8.0#system-math-round(system-double-system-int32)).
Happy to discuss your concerns regarding this set of functions. 

Also, closes #462

---

Make sure that you:

---

### Available Types:

existing).

</details>


src/_round.ts
```diff
@@ -0,0 +1,21 @@
+export function _round(roundingFn: (value: number) => number) {
```
eranhirsch, line 1: like in Lodash, the right name here is probably something like createRound or as a higher-order function: `withPrecision`
```suggestion
export function _withPrecision(roundingFn: (value: number) => number) {
```

src/_round.ts
```diff
@@ -0,0 +1,21 @@
...
+  return (value: number, precision: number) => {
+    if (!Number.isInteger(precision)) {
```
eranhirsch, line 3: We should special-case the `precision === 0` case to reduce our footprint:

```suggestion
  return (value: number, precision: number) => {
    if (precision === 0) {
      return roundingFn(value);
    }

    if (!Number.isInteger(precision)) {
```

src/floor.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Rounds down a given number to a specific precision.
+ * @param value The number to round down.
+ * @param precision The precision to round down to. Must be an integer between -15 and 15.
+ * @signature
+ *    R.floor(value, precision);
+ * @example
+ *    R.floor(123.4317, 3) // => 123.431
+ *    R.floor(483.22243, 1) // => 483.2
+ *    R.floor(8541, -2) // => 8500
```
eranhirsch, line 13: All these examples need to show the difference between floor, ceil, and round. I would take an example where each one of them would have a different result. Assume some users don't know what `ciel` and `floor` do.

e.g.
```suggestion
 *    R.floor(123.9876, 3) // => 123.987
 *    R.floor(483.22243, 1) // => 483.2
 *    R.floor(8541, -1) // => 8540
 *    R.floor(456789, -3) // => 456000
```

For `round`, these examples would look like this:
```ts
 *    R.round(123.9876, 3) // => 123.988
 *    R.round(483.22243, 1) // => 483.2
 *    R.round(8541, -1) // => 8540
 *    R.round(456789, -3) // => 457000
```

And for `ceil`:
```ts
 *    R.ceil(123.9876, 3) // => 123.987
 *    R.ceil(483.22243, 1) // => 483.3
 *    R.ceil(8541, -1) // => 8550
 *    R.ceil(456789, -3) // => 456000
```

This way users can compare the output and pick the right one for their use-case.

eranhirsch (approved):
Looks great, my only real concern is around documentation: People don't assume that Remeda incurs runtime costs (because of the `purry` function), when `precision === 0`  this function doesn't add anything over the native implementation so it's best to make sure users are aware of that. We can also improve perf slightly for this trivial case (see my comment below)

src/floor.ts
paolostyle: I agree and that was my intention when providing these examples, although I can see how your examples give a clearer view on the differences between these functions than the ones I provided (particularly the negative precision ones).

src/round.ts
```diff
@@ -0,0 +1,36 @@
+import { _round } from './_round';
+import { purry } from './purry';
+
+/**
+ * Rounds a given number to a specific precision.
```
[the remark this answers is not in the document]
paolostyle: I think the return type warning is a bit of an overkill 😅 I could see it as a linter rule or something. In any case I updated the comment.

paolostyle:
Thanks for the review, I addressed your comments.

PR #466, opened 2024-01-20
feat(partialBindTo): Add partialBindTo, partialRightBindTo

### LINKED ISSUE #458: Add partial, partialRight

Lodash and ramda both have `partial`, and `partialRight` methods and I don't see any replacements in remeda yet. We use lodash's `parseInt` which is sort of a partialRight of `Number.parseInt` with the radix of 10 already applied. I was going to write my own but noticed there's no partial/currying methods yet.

[Lodash docs](https://lodash.com/docs/4.17.15#partial)
[Lodash source](https://github.com/lodash/lodash/blob/main/src/partial.ts)

cjquines:
closes #458 

adds `partial`, `partialRight` functions

---

Make sure that you:

---

### Available Types:

existing).

</details>


eranhirsch:
I'm sorry it took me so long to review this. I'm having a hard time finding the unique value of this function over modern arrow functions. I must admit that when I said I was supportive of this, I thought we were discussing a function that handles objects and not functions.

I understand this is part of lodash, but that is due to lodash being a very old library that predates ES6, which added arrow functions to the language. Can you show a use-case where this would provide substantial benefit over using arrow functions?

cc: @viveleroi

cjquines:
:shrug: i personally dont have one

are you thinking about https://ramdajs.com/docs/#partialObject ?

viveleroi:
The problem with arrow functions is that we'd have to either redefine all target function arguments and types or use the rest operator and exclude the param we're overriding in the `Parameters<>` type. Using partial/partialRight simply reduces boilerplate when we need this.

We don't need this very often but here's an example - a grid library provides a cell formatter and we provide a modified version that uses our defaultDateFormat const.

`valueFormatter: partialRight(cellDateFormatter, [defaultDateFormat])`

eranhirsch:
> valueFormatter: partialRight(cellDateFormatter, [defaultDateFormat])

I would assume that `valueFromatter` is already typed, so it would provide the types for typescript to infer into your arrow function, but regardless I get your broader point about typescript not being able to infer "backward" from the function implementation to the function signature makes certain typing situations harder. It's still not a very good utility in my mind because it doesn't solve many other common cases (like functions that take an "options" object as its param). I'm OK with these functions, but I think they should be renamed to include some hint that they are used for functions. One name that comes to mind from the Lodash documentation of `partial` itself is `partialBind.` Would that work for everyone?

eranhirsch:
@cjquines - going over the code, I'm curious to get your thoughts on the API for this function. I feel like the variadic args, although that's how it is done in lodash, is actually less useful than how it's done in Ramda where the args are an array. Making the function take an array and a function, and flipping the order of things so that the args are the "data" component, and the function is the param, e.g. `partial(args, func)` would allow us to implement a dataLast version, e.g. `partial(func)(args)` which could be used **in a pipe**.

This would allow things like:

```ts
const partialFoo = pipe(
  data,
  map(...),
  filter(...),
  groupBy(...),
  values,
  ...,
  partial(foo),
);
const value = partialFoo(...otherArgs);
```

Whereas the current implementation doesn't permit it.

This does make the "partial" name even less suitable, though, as the "data" now isn't the function at all; it's the arguments. Maybe `assign`, `assignArguments`, `augmentFunction`, `useWith`, or some other name...

eranhirsch:
ChatGPT thinks that `partialApply`, `bindArgs`, `preApply`, and `prependArgs` are valid/good names too.

cjquines:
i do like that api better (and it is something that'll be more useful in my code too). let me rewrite the function and we can bikeshed the name—don't have any opinions right now

cjquines:
ok done—as for the name, i am liking `preApply`/`postApply` and `prependArgs`/`appendArgs`

src/partial.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Creates a function that calls `func` with `args` put before the arguments
+ * it receives.
+ *
+ * **Note:** This method doesn't set the `length` property of partially applied
+ * functions.
+ *
+ * @param func the function to wrap
+ * @param args the arguments to put before
```
eranhirsch, line 16: flip order
```suggestion
 * @param args the arguments to put before
 * @param func the function to wrap
```

src/partial.ts
```diff
@@ -0,0 +1,76 @@
...
+  U extends ReadonlyArray<unknown>,
+> = T extends [...U, ...infer V] ? V : never;
+
+/**
+ * Creates a function that calls `func` with `args` put before the arguments
+ * it receives.
+ *
+ * **Note:** This method doesn't set the `length` property of partially applied
+ * functions.
+ *
```
eranhirsch, line 14: I don't think this note is very important or valuable. It's confusing and not necessarily relevant in a post-ES3/5 world where people rarely care for the "length" of a function.
```suggestion
```

src/partial.ts
```diff
@@ -0,0 +1,76 @@
+import { purry } from './purry';
+
+type RemovePrefix<
+  T extends ReadonlyArray<unknown>,
+  U extends ReadonlyArray<unknown>,
+> = T extends [...U, ...infer V] ? V : never;
```
eranhirsch, line 6: You need to add `readonly` here to make this generic enough for all cases because mutable arrays extend readonly arrays, but not vice-versa.

```suggestion
> = T extends readonly [...U, ...infer V] ? V : never;
```

src/partial.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @dataFirst
+ * @category Function
+ * @see partialRight
+ */
+export function partial<
+  A extends ReadonlyArray<unknown>,
+  B extends ReadonlyArray<unknown>,
+  C,
+>(args: [...A], func: (...args: B) => C): (...rest: RemovePrefix<B, A>) => C;
```
eranhirsch, line 35: There's a slight issue with how this is implemented now, in regards to when the "A" type is not a prefix of the arguments for `func`. The invocation of "partial" would be successful and won't result in any error on this line, but the returned function would take arguments as never, e.g. `foo(...args: never)` so only when that function is called will the user see the issue. This could be hard to debug if the function is called much later or further downstream from the partial call when a lot of typescript inference is relied on.

One trick to provide more context is to use a literal string instead of `never` for the `RemovePreifx` type, e.g. 

```ts
type RemovePrefix<
  T extends ReadonlyArray<unknown>,
  U extends ReadonlyArray<unknown>,
> = T extends [...U, ...infer V] ? V : "ERROR(partial): the provided argument types are not a prefix of the function's expected argument types";
```

eranhirsch, line 35: To get a more generic and flexible implementation we can use a function type instead of just the array of it's arguments, this provides typescript more power to infer for itself the types:


```suggestion
export function partial<
  T extends IterableContainer,
  // This is how functions are defined in ReturnType and Paramters by typescript
  F extends (...args: any) => any,
>(data: T, func: F): (...rest: RemovePrefix<Parameters<F>, T>) => ReturnType<F>;
```

eranhirsch (changes requested):
This looks great, let's go over the points I added, and if you agree with them, incorporate them throughout (i only commented on the dataFirst of `partial`.

And let's decide on a name for the functions. I like `partiallyBindTo` so that we keep the `partial` part for people coming from Lodash and Ramda, but make it a little more easier to grok when reading the code for someone who doesn't know the function. the `To` reverses the roles so that the function provided is the "param" and not the "data".

src/partial.ts
cjquines: hm, i like this, but in this case i had to put the literal string in an array to make it a rest argument type. idk if there's a better solution here

cjquines:
i settled with `partialBindTo` and `partialRightBindTo`; note that in this case `partialRightBindTo` is now the longest function name :tada:

src/partial.ts
eranhirsch: This is solved by using our `IterableContainer` helper type (i tried it!), I think it has something to do with how typescript infers tuples generically. 


Inspired by:
https://github.com/microsoft/TypeScript/blob/2a0edf7949ad8427e3ef233a4c549ecbe32953e1/src/lib/es2015.promise.d.ts#L21

src/partialBindTo.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @dataFirst
+ * @category Function
+ * @see partialRightBindTo
+ */
+export function partialBindTo<
+  T extends ReadonlyArray<unknown>,
+  F extends (...args: any) => any,
+>(
+  data: [...T],
```
eranhirsch, line 34: Use our `IterableContainer` utility type. I can't explain why it works differently, but it seems to have something to do with how typescript infers tuples. I got the inspiration from:

https://github.com/microsoft/TypeScript/blob/2a0edf7949ad8427e3ef233a4c549ecbe32953e1/src/lib/es2015.promise.d.ts#L21

```suggestion
  T extends IterableContainer,
  F extends (...args: any) => any,
>(
  data: T,
```

src/partialBindTo.ts
```diff
@@ -0,0 +1,74 @@
...
+export function partialBindTo() {
+  return purry(_partialBindTo, arguments);
+}
+
+function _partialBindTo<
+  A extends ReadonlyArray<unknown>,
+  B extends ReadonlyArray<unknown>,
+  C,
+>(data: A, func: (...args: [...A, ...B]) => C): (...rest: B) => C {
+  return (...rest: B) => func(...data, ...rest);
```
eranhirsch, line 73: To validate that our types are sound, use the same types in the implementation:

```suggestion
function _partialBindTo<
  T extends IterableContainer,
  F extends (...args: any) => any,
>(
  data: T,
  func: F
): (...rest: RemovePrefix<Parameters<F>, T>) => ReturnType<F> {
  return (...rest) => func(...data, ...rest);
```

src/partialBindTo.ts
```diff
@@ -0,0 +1,74 @@
+import { purry } from './purry';
+
+type RemovePrefix<
+  P extends ReadonlyArray<unknown>,
+  T extends ReadonlyArray<unknown>,
+> = P extends readonly [...T, ...infer U]
+  ? U
+  : ["ERROR(partialBindTo): Data doesn't match function signature."];
```
eranhirsch, line 8: I'm afraid that engineers that don't know Remeda well might find the error message confusing, let's add more context:
```suggestion
  : ["RemedaTypeError(partialBindTo): Data doesn't match function signature."];
```

src/partialRightBindTo.ts
```diff
@@ -0,0 +1,74 @@
+import { purry } from './purry';
+
+type RemoveSuffix<
+  P extends ReadonlyArray<unknown>,
+  U extends ReadonlyArray<unknown>,
+> = P extends readonly [...infer T, ...U]
+  ? T
```
eranhirsch, line 7: Similar comment to the above, the naming here is confusing to me, Usually T should be reserved to the "main" actor in the type, and the others could be named more descriptivly.
```suggestion
type RemoveSuffix<
  T extends IterableContainer,
  Suffix extends IterableContainer,
> = T extends readonly [...infer Rest, ...Suffix]
  ? Rest
```

src/partialBindTo.ts
```diff
@@ -0,0 +1,74 @@
+import { purry } from './purry';
+
+type RemovePrefix<
+  P extends ReadonlyArray<unknown>,
+  T extends ReadonlyArray<unknown>,
+> = P extends readonly [...T, ...infer U]
+  ? U
```
eranhirsch, line 7: 1. These should probably also use IterableContainer to cover all cases (I can't think of one, but just in case).
2. Are the letters here reversed? P is for "Prefix"? shouldn't it be the other way around?
```suggestion
type RemovePrefix<
  T extends IterableContainer,
  Prefix extends IterableContainer,
> = T extends readonly [...Prefix, ...infer Rest]
  ? Rest
```

src/partialBindTo.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      expect(partialBindTo(fn)([1])(2, 3)).toBe(fn(1, 2, 3));
+    });
+
+    test('should partially apply all args', () => {
+      expect(partialBindTo(fn)([1, 2, 3])()).toBe(fn(1, 2, 3));
+    });
+  });
+});
+
+describe('typing', () => {
```
eranhirsch, line 35: We are not testing any negative flows: too many args and wrong arg types

src/partialBindTo.test.ts
```diff
@@ -0,0 +1,71 @@
...
+        (y: number, z: number) => string
+      >();
+    });
+
+    test('should correctly type all partial args', () => {
+      expectTypeOf(partialBindTo([1, 2, 3], fn)).toEqualTypeOf<() => string>();
+    });
+  });
+
+  describe('data-last', () => {
```
eranhirsch, line 54: dataLast should also test inference via `pipe`, as those work a bit differently to simple dataLast invocations. Because dataLast is mainly used in pipes those are the most important cases to test.

src/partialBindTo.test.ts
```diff
@@ -0,0 +1,71 @@
+import { partialBindTo } from './partialBindTo';
```
eranhirsch, line 1: I want to make sure that we cover a wide range of tests, specifically I'm curious to see that everything works (both runtime and typing) with functions which take variadic parameters themselves (e.g. `(...strings: Array<string>) => strings.join(",")`) and functions that take optional arguments (e.g. `(a: string, b: isSomething = false) => isSomething ? a : "test"`)

eranhirsch (changes requested):
There's a lot that could go wrong with typing in the unhappy flows, let's add tests to cover as much cases as possible to we can assure users that the function can handle their needs.

src/partialBindTo.ts
```diff
@@ -0,0 +1,74 @@
...
+ *    R.partialBindTo(data, func)
+ *
+ * @example
+ *    const fn = (x, y, z) => `${x}, ${y}, and ${z}`
+ *    const partialFn = R.partialBindTo([1, 2], fn)
+ *    partialFn(3) // => 1, 2, and 3
+ *
+ * @dataFirst
+ * @category Function
+ * @see partialRightBindTo
```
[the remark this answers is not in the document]
cjquines: hm, i think we should parse these then! i'll open a docs ticket

src/partialBindTo.ts
cjquines: the only other place where we have a similar type error message is with `fromPairs` i think, and that one uses ERROR—though maybe it'd be nice to audit the library and check other places where `never` is used and change that

cjquines: anyway, will change

cjquines:
ah, yes, the current types indeed aren't great:

```ts
const fn = (x: number, y: number, z: number) => `${x}, ${y}, and ${z}`;

const data: Array<string> = [];

// this should error:
partialBindTo(data, fn); // (...rest: unknown[]) => string
```

```ts
const fn = (...xs: Array<number>) => xs.join(',');

// this should not error:
partialBindTo([1, 2, 3], fn)(4, 5, 6); // error
```

```ts
const fn = (x: number = 1) => x;

// this should not error:
partialBindTo([2], fn)(); // error
```

any thoughts on this?

eranhirsch:
We'll need to recursively consume the Prefix and Suffix to cover all cases. Check out my changes and the extra tests. I don't have enough time to follow the changes to the `partialRight` function, and I think we need to add more tests, would appreciate your help on this.

cjquines:
this still doesn't handle a similar example to the first one i gave above: [playground link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAksECcCGAjANhAwgewHbCQEtdEoBeKAJQiQBM80QBBBZEAHgFdcBrXbAO64AfFAA+UANoBdANwAoeaEhUIAWwi0kAFXAQAoq2wJ2AWQgBnC0gDm0CAA94uWhagXgCYjfFRcnNRREUQpJeSgocytbeycIFzcPL1wbcIioAH4oAANqDS1dSEMEYwAKMCQEYEIkNAAhYlptbABKAC4oABIAbyjrOwBfADpstIiOvM0dPWLjdmzJfUpKNp6+mIHpKABlbA1gAAtvKG5HSABjeFooA6QwSBJaAEIoMAwkC2gECDBjYChDoQ3MQAQcvuoplAbIRDpwUK8SgArCCXbLCAA08jkimU0Dy2AAbhAAArfABmhAc7DS2igjmcrlg8GQ6CweAIxEQmIipIgFIcdLiCSZiFQGBw+CIJAQmJCUF5-MkAHIMClDkqtvT4oyAAxpLLaNIdbTK1U2dWaoW6saZVT5aZFIwmJXNbBQNRIXAgKCVGxuMAlAmEWiaAFusncS6EPBK4Q242ChluSTEMmkbQACRotHRUCG+dT6eoHmkNoiWQVlMT2uThYQ8vJlKzdFz+aGdYbfMpxeApfS-dtlYczeuWuFmezZYHWXxRKH7G0PdzQ57cYHA4mEIKMyd7CVLBsAXi-2wZNB0AEJRSAL0CMJwdDwHDkeqMbX66gm-thQMu4AjO+-YJmOjKSKUHYTnQLQZK2Ba4Gm9aLpYvZTuWnb8tWwopvBpBDiOsHtjh9YrshfYfmheHZphjKQbQqHTnahIko2VJIR4y4sau9FAXaUw-rMzoHke+BQKe55QJeeA+Lid5BiG1xPlAEa4FGb7cZ+6FViBbhscA6lobp+kaZM26OiUJgAEyARuvGmb+5nsAAzMIChKLexKVNUtSMHU2DcLQ1IRAAYtRbilG2voWB0nogC05CiDF3JQLS2kiiy4rslKXLyHK4X5t8Hhfkx84ecg+yIBY7BBRiyXCHFZCiNQwCcAguA-lVLmKMpqm4K8nk1PUjTNIFyWhWlYpspKnIymkIWpXlQyRdFXr1QlXqyqUWgEB0khtto0i5t1HRBe08r9d5IC+f5VW5toojdGk3zNa1UALQVwCrUpkYLVtSAEe9LQKAMijnHgHhQL9HQsGw7D+IEwTkFI2Kg7g4MoH5LiIxUVQDQ0LjNJtSAELmpQOB0cNBDKUAgOTASU7mABetPwwgn0AERs4DbkqFdmMULiYno-58gAPQi+kAB6GRAA)

cjquines:
ok how about this: [playground link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAksECcCGAjANhAwgewHbCQEtdEoBeKAJQiQBM80QBBBZEAHgFdcBrXbAO64AfFAA+UANoBdANwAoeaEhUIAWwi0kAFXAQAoq2wJ2AWQgBnC0gDm0CAA94uWhagXgCYjfFRcnNRREUQpJeSgocytbeycIFzcPL1wbcIioAH4oAANqDS1dSEMEYwAKMCQEYEIkNAAhYlptbABKAC4oABIAbyjrOwBfADpstIiOvM0dPWLjdmzJfUpKNp6+mIHpKABlbA1gAAtvKG5HSABjeFooA6QwSBJaAEIoMAwkC2gECDBjYChDoQ3MQAQcvuoplAbIRDpwUK8SgArCCXbLCAA08jkimU0CY5CgLDY7H8gWCkgA5BgUocKVtHM5XFAAAyZE68fhCKAdEgAN0QCgA9IL0gA9DI4vSqNTYfkABW+ADNCA52GltFAGfEmXBEKgMDh8EQSAhMREFRBlQ5NXEErB4Mh0Fg8ARiIhMSEoBarZTqTZafTbUzmWksto0h1tL74v6DnSbYy3CH0mzvSqE9q3ESkBwkLgQMIxukw0XxtKpoUDEYTBTmtgoGo8yAoJUbG4wCVeYRaJoAfXFdxLoQ8BTCynIxm7ZJiIrSNoABI0WjoqBDNczufUDzSUuppXprVTjcIL37hyLugrtdDY+ny0qrfAHcplNZNPnpeTpkLpe71-S2UIHfdhtEfFd30fMcX3HcsChmat2ApFgbACeJ-mwRVQWgAQShSAEpQ7WVu17YB+0HaoRyg6COnfL8s1YHMQIvWgqOgtlJB-OgMivNc8kA4DQMsYBwLPYRnzYstJjgooEIARlYssNUPJlJFKW9ONoFpuNXddcFnE9BO3P893va1lLcac9NId9mJ4m8rJPCChPEiS3zPZi6KgDTjP-Pj5TPECwLvK1IJ89IJghaSqxKGtkNQ-AoAwrCoBwvAfFxBEiJ7a5SKgAdcCHSiwpos9PMM4AwoiMNH0q7lYOmGSYvYAAmBTwvqytZhMABmYQFCUKU5UqapakYOpsG4Wg1QiAAxTzSmvVsLA6JsWnIUQmzNLzPN1R0DRdY13XkT0FrXb4PAimV-NM9ghuQfZEAsdgZoxLzhDWshRGoYBOAQXBK2evrFHywrcFeYaanqRpmmm7bzPtPUnUNV0TS2ub4dOoYlpW-MPo2-MPVKLQCA6SRr20aQVxBjoZvaL0IdGkBxsm56V20URujSb4fr+qBMfO4A8bywdMeJpA7IFloFAGRRzjwDwoDFjpsw4UkggQT0ZAUOXcAVlAJpcAkKiqSGGhcZoiaQAgV1KBweQCdWVxAe2yVNKAAC8XfVoWACIfalgaVGZw2KFxJL9cm+RhTFDIgA)


```diff
 type RemovePrefix<
   T extends IterableContainer,
   Prefix extends IterableContainer,
 > = Prefix['length'] extends 0
   ? T
   : T['length'] extends 0
-    ? RemedaTypeError<'Too many args provided to function'>
+    ? Prefix extends Array<any>
+      ? T
+      : RemedaTypeError<'Too many args provided to function'>
     : T extends [infer THead, ...infer TRest]
       ? Prefix extends [infer PrefixHead, ...infer PrefixRest]
         ? PrefixHead extends THead
           ? RemovePrefix<TRest, PrefixRest>
           : RemedaTypeError<'Argument of the wrong type provided to function'>
-        : RemedaTypeError<1>
+        : Prefix extends Array<THead>
+          ? [THead?, ...RemovePrefix<TRest, Prefix>]
+          : RemedaTypeError<1>
       : T extends [(infer THead)?, ...infer TRest]
         ? Prefix extends [infer PrefixHead, ...infer PrefixRest]
           ? PrefixHead extends THead
             ? RemovePrefix<TRest, PrefixRest>
             : RemedaTypeError<'Argument of the wrong type provided to function'>
           : Prefix extends TRest
             ? TRest
             : RemedaTypeError<2>
         : RemedaTypeError<3>;
```

PR #473, opened 2024-01-24
chore(docs): rebuilt and modernized docs with Astro, React and shadcn/ui

xanderbarkhatov:


TkDodo:
@eranhirsch can you take a look please?

xanderbarkhatov:
There's probably still some work to configure deploys. I just don't know how you usually do it, so feel free to point me in a direction, and I'll continue.

docs/src/components/code-block.astro
```diff
@@ -0,0 +1,21 @@
...
+const { code, lang = "js", copy = false } = Astro.props;
+---
+
+<div class="relative">
+  <Code code={code} lang={lang} theme="poimandres" />
+
+  <div class="absolute right-2 top-1/2 -translate-y-1/2">
+    {copy && <CopyButton value={code} client:load />}
+  </div>
+</div>
```
eranhirsch, line 21: This component will wrap **all** our code examples in the main "docs" page with 2 divs unnecessarily for the single use-case where we need the copy button.

Let's remove the whole abstraction.

This cuts out about 15% of the size of the docs.html file.
```suggestion
```

docs/src/components/code-block.astro
```diff
@@ -0,0 +1,21 @@
...
+
+import { CopyButton } from "@/components/copy-button";
+
+interface Props {
+  code: string;
+  lang?: "bash" | "js";
+  copy?: boolean;
+}
+
+const { code, lang = "js", copy = false } = Astro.props;
```
eranhirsch, line 12: We should use "ts" as the default language, we sometimes use typescript code in our examples.
```suggestion
interface Props {
  code: string;
  lang?: "bash" | "ts";
  copy?: boolean;
}

const { code, lang = "ts", copy = false } = Astro.props;
```

docs/src/components/copy-button.tsx
```diff
@@ -0,0 +1,52 @@
+import * as React from "react";
```
eranhirsch, line 1: don't import * from React, import the specific things you need from it.
```suggestion
import { useEffect, useState, type HTMLAttributes } from "react";
```

docs/src/components/copy-button.tsx
```diff
@@ -0,0 +1,52 @@
...
+  className,
+  src,
+  ...props
+}: CopyButtonProps) {
+  const [hasCopied, setHasCopied] = React.useState(false);
+
+  React.useEffect(() => {
+    setTimeout(() => {
+      setHasCopied(false);
+    }, 2000);
```
eranhirsch, line 27: extract the magic number to a const at the top of the file

docs/src/components/copy-button.tsx
```diff
@@ -0,0 +1,52 @@
...
+  src?: string;
+}
+
+async function copyToClipboard(value: string) {
+  navigator.clipboard.writeText(value);
+}
+
+export function CopyButton({
+  value,
+  className,
```
eranhirsch, line 18: You never send a className to this component; why add the additional complexity of supporting it prematurely?

docs/src/components/copy-button.tsx
```diff
@@ -0,0 +1,52 @@
+import * as React from "react";
+import { CheckIcon, CopyIcon } from "@radix-ui/react-icons";
+
+import { cn } from "@/lib/utils";
+import { Button } from "@/components/ui/button";
+
+interface CopyButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
+  value: string;
+  src?: string;
```
eranhirsch, line 9: This prop is never used, why do you define it?

```suggestion
```

docs/src/components/copy-button.tsx
```diff
@@ -0,0 +1,52 @@
...
+      setHasCopied(false);
+    }, 2000);
+  }, [hasCopied]);
+
+  return (
+    <Button
+      size="icon"
+      variant="ghost"
+      className={cn(
+        "relative z-10 h-6 w-6 text-zinc-50 hover:bg-zinc-700 hover:text-zinc-50",
```
eranhirsch, line 35: Shouldn't we use the theme colors here instead of just tailwind-zinc?

docs/src/components/copy-button.tsx
```diff
@@ -0,0 +1,52 @@
...
+  src,
+  ...props
+}: CopyButtonProps) {
+  const [hasCopied, setHasCopied] = React.useState(false);
+
+  React.useEffect(() => {
+    setTimeout(() => {
+      setHasCopied(false);
+    }, 2000);
+  }, [hasCopied]);
```
eranhirsch, line 28: This dependency is fragile because the effect doesn't actually rely on it. A better approach would be to include it in the code for useEffect so that it's part of the effect's life-cycle. you can, for example, check that it is in fact true before setting the timeout.

docs/src/components/copy-button.tsx
```diff
@@ -0,0 +1,52 @@
...
+      size="icon"
+      variant="ghost"
+      className={cn(
+        "relative z-10 h-6 w-6 text-zinc-50 hover:bg-zinc-700 hover:text-zinc-50",
+        className,
+      )}
+      onClick={() => {
+        copyToClipboard(value);
+        setHasCopied(true);
+      }}
```
eranhirsch, line 41: I don't know how astro specifically manages re-renders and state, but when sending arrow functions to react components it's usually recommended to wrap them with a useCallback to avoid any accidental object comparison issues in effects or re-renders.

eranhirsch (changes requested):
This is a huge PR and I didn't get to review all of it and ran out of time right now. I started by looking at the config generally without looking at the specifics of the implementation.

Take a look at my comments and feel free to ask any questions or if you have any push backs.

docs/package.json
```diff
@@ -3,51 +3,45 @@
   "version": "1.0.0",
   "repository": "git@github.com:remeda/docs.git",
   "license": "MIT",
-  "author": "Łukasz Sentkiewicz",
```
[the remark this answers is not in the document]
xanderbarkhatov: By accident

docs/.gitignore
```diff
@@ -1,4 +1,23 @@
...
+yarn-error.log*
+pnpm-debug.log*
+
+# environment variables
+.env
+.env.production
+
+# macOS-specific files
+.DS_Store
+.vercel
```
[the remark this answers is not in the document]
xanderbarkhatov: It is deployed from my repo to Vercel atm, so it was added automatically. I'll remove it.

docs/src/components/copy-button.tsx
xanderbarkhatov: Astro doesn't have anything to do with managing React renders and state afaik

docs/package.json
```diff
@@ -3,51 +3,45 @@
...
+  "type": "module",
   "scripts": {
-    "build": "tsc && vite build",
-    "build:netlify-ci": "npm run typedoc && npm run transform && npm run build && npm run netlify-spa-fix",
-    "clean": "rm -rf dist && rm -rf build",
```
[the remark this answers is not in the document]
xanderbarkhatov: What exactly?
The build script is just different and I extracted the generation part in it's own command.
The only thing missing is the clean command.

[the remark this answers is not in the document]
xanderbarkhatov: We don't need netlyfy-spa-fix here, as it's not and SPA

package.json
```diff
@@ -34,7 +34,7 @@
     "compile": "tsc",
     "compile:dist": "tsc --project tsconfig.dist.json",
     "coverage": "vitest run --coverage",
-    "docs:build": "cd docs && npm ci && npm run build:netlify-ci",
```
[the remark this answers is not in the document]
xanderbarkhatov: We don't need netlify-spa-fix here, as it's not and SPA

docs/tailwind.config.js
```diff
@@ -1,8 +1,81 @@
 /** @type {import('tailwindcss').Config} */
-module.exports = {
-  content: ['./index.html', './src/**/*{js,ts,jsx,tsx}'],
+export default {
+  darkMode: ["class"],
+  content: [
+    "./pages/**/*.{ts,tsx,astro}",
+    "./components/**/*.{ts,tsx,astro}",
+    "./layouts/**/*.{ts,tsx,astro}",
+    "./src/**/*.{ts,tsx,astro}",
```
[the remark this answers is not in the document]
xanderbarkhatov: I just added .astro extension to the config. But can change to be exactly as in the guide.

docs/src/components/copy-button.tsx
xanderbarkhatov: I don't think it really matters here

docs/package.json
```diff
@@ -3,51 +3,45 @@
...
     "tiny-invariant": "^1.3.1",
-    "ts-node": "^10.9.2",
-    "type-fest": "^4.9.0",
-    "typedoc": "^0.25.5",
+    "tsx": "^4.7.0",
+    "typedoc": "^0.25.7",
     "typescript": "^5.3.3",
-    "vite": "^5.0.12"
+    "@types/react": "^18.2.48",
+    "@types/react-dom": "^18.2.18"
```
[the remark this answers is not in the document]
xanderbarkhatov: Why does it matter?

docs/package.json
```diff
@@ -3,51 +3,45 @@
...
-    "@vitejs/plugin-react": "^4.2.1",
-    "autoprefixer": "^10.4.16",
-    "eslint": "^8.56.0",
-    "eslint-config-prettier": "^9.1.0",
-    "eslint-import-resolver-typescript": "^3.6.1",
-    "eslint-plugin-import": "^2.29.1",
-    "eslint-plugin-react": "^7.33.2",
-    "eslint-plugin-react-hooks": "^4.6.0",
-    "eslint-plugin-react-refresh": "^0.4.5",
-    "marked": "^11.1.1",
```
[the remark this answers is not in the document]
xanderbarkhatov: I didn't touch this, it is the same as in the current docs.

docs/package.json
```diff
@@ -3,51 +3,45 @@
...
-    "prettier": "^3.1.1",
-    "prettier-plugin-tailwindcss": "^0.5.10",
-    "tailwindcss": "^3.4.0",
+    "@types/node": "^20.11.5",
+    "prettier": "^3.2.4",
+    "prettier-plugin-astro": "^0.13.0",
+    "prettier-plugin-tailwindcss": "^0.5.11",
     "tiny-invariant": "^1.3.1",
-    "ts-node": "^10.9.2",
-    "type-fest": "^4.9.0",
```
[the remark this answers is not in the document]
xanderbarkhatov: Same as above.

I don't think it really matters.

docs/tsconfig.json
```diff
@@ -1,39 +1,11 @@
 {
```
eranhirsch, line 1: The project doesn't compile for me right now, the tsconfig is missing the compiler option: "allowArbitraryExtensions" which should be `true`

docs/src/components/navbar.tsx
```diff
@@ -0,0 +1,83 @@
+import { useMemo, useState } from "react";
+
+import { buttonVariants } from "@/components/ui/button";
+import { Input } from "@/components/ui/input";
+import { ScrollArea } from "@/components/ui/scroll-area";
+import { FunctionTag } from "@/components/function-tag";
+import { cn } from "@/lib/utils";
+
+import { toPairs } from "../../../src";
```
eranhirsch, line 9: You can't do this, this doesn't compile because the path is outside of the typescript project. Add remeda to the project and use it as an external dependency.
```suggestion
import { toPairs } from "remeda";
```

docs/src/components/mobile-nav.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <Button variant="ghost" size="icon" className="md:hidden">
+          <HamburgerMenuIcon />
+        </Button>
+      </SheetTrigger>
+
+      <SheetContent>
+        <div className="h-[calc(100vh-1.5rem)] pb-10 pt-6">
+          <Navbar entries={props.entries} onSelect={() => setOpen(false)} />
+        </div>
+      </SheetContent>
```
eranhirsch, line 25: You don't need this div, I don't understand what it's purpose is. You can add whatever classNames you need to the SheetContent itself, and from my tests you only need to add the top padding to make room for the close button, all the rest are redundant
```suggestion
      <SheetContent className="pt-12">
          <Navbar entries={props.entries} onSelect={() => setOpen(false)} />
      </SheetContent>
```

docs/src/components/navbar.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+export const Navbar = ({
+  entries,
+  onSelect,
+}: {
+  entries: Record<string, Array<{ name: string; tags: Array<string> }>>;
+  onSelect?: () => void;
+}) => {
```
eranhirsch, line 17: you don't need to take entries as a prop, you can import it directly in this component.
```suggestion
export const Navbar = ({
  onSelect,
}: {
  onSelect?: () => void;
}) => {
```

docs/src/components/navbar.tsx
```diff
@@ -0,0 +1,83 @@
...
+                {functions.map((func) => (
+                  <li key={func.name}>
+                    <a
+                      href={`#${func.name}`}
+                      className={cn([
+                        buttonVariants({ variant: "ghost" }),
+                        "text-muted-foreground",
+                        "flex w-full items-center justify-between gap-1",
+                      ])}
+                      onClick={() => onSelect?.()}
```
eranhirsch, line 62: redundant arrow
```suggestion
                      onClick={onSelect}
```

docs/src/components/mode-switcher.tsx
```diff
@@ -0,0 +1,59 @@
...
+    } else {
+      de.classList.add("light");
+      de.style.colorScheme = "light";
+    }
+  } else {
+    de.classList.add(theme);
+    de.style.colorScheme = theme;
+  }
+
+  localStorage.setItem("theme", theme);
```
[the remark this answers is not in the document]
xanderbarkhatov: I don't get it. We're not locking anything with this. We're saving 'dark', 'light', or 'system' to LS (yes, 'light' is redundant). If the saved value is 'system', we're getting the theme from the media query. Value 'system' and the absence of value mean the same thing here. 'System' is just more explicit.

docs/src/components/code-block.astro
```diff
@@ -0,0 +1,12 @@
...
+
+interface Props {
+  code: string;
+  lang?: "bash" | "ts";
+}
+
+const { code, lang = "ts" } = Astro.props;
+---
+
+<Code code={code} lang={lang} theme="poimandres" />
```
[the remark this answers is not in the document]
xanderbarkhatov: It encapsulates the theme in case we wanted to change it

docs/src/components/mode-switcher.tsx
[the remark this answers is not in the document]
eranhirsch: Yea, I re-read the code and don't see how I got to my conclusion. I don't see the bug I mentioned anymore. I think it was due to the fact that in the example in shadcn/astro/tailwind they do things differently.

docs/src/components/code-block.astro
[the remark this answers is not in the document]
eranhirsch: ok

docs/tsconfig.json
xanderbarkhatov: That is weird. It is building for me, deployed from my repo, and even here in this PR there's a preview deployment on Netlify.

xanderbarkhatov:
> I'm sorry that this probably feels slow to you, but this is a big change that would impact the whole docs site so it has to be done "right".

No, I totally get it.
I'm sorry for some silly mistakes, I rushed it at the end to get the initial feedback.

docs/tsconfig.json
eranhirsch: Specifically the netlify build only runs on the initial PR when it's uploaded, so if further commits change things I don't think it redeploys again...

eranhirsch:
Hey, I just checked a fresh clone, you are right that the **build** completes, but typescript fails (and the build doesn't care it seems). This is my full repro:
```bash
gh repo clone remeda/remeda
cd remeda
gh pr checkout 473
nvm use
npm ci
npm run build
cd docs
npm ci
npm run typedoc
npm run transform
npx tsc
```

xanderbarkhatov:
> Hey, I just checked a fresh clone, you are right that the **build** completes, but typescript fails (and the build doesn't care it seems). This is my full repro:
> 
> ```shell
> gh repo clone remeda/remeda
> cd remeda
> gh pr checkout 473
> nvm use
> npm ci
> npm run build
> cd docs
> npm ci
> npm run typedoc
> npm run transform
> npx tsc
> ```

Yea, agree, checked it in my branch too.
I updated the code to use remeda as a dependency and now it looks fine.

docs/src/styles/globals.css
```diff
@@ -0,0 +1,68 @@
...
+    @apply min-w-0 border-border;
+  }
+
+  body {
+    @apply bg-background text-foreground;
+  }
+
+  code {
+    @apply font-mono text-primary;
+  }
```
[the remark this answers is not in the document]
xanderbarkhatov: Text is not primary by default. These styles are for code blocks inside generated functions data.

docs/src/components/theme-script.astro
```diff
@@ -0,0 +1,18 @@
...
+    document.documentElement.classList.remove('dark');
+  }
+
+  query.addEventListener('change', e => {
+    if (e.matches && !localStorage.getItem("theme")) {
+      document.documentElement.classList.add('dark');
+    } else {
+      document.documentElement.classList.remove('dark');
+    }
+  });
```
eranhirsch, line 17: really cool!

docs/src/components/theme-switcher.tsx
```diff
@@ -0,0 +1,55 @@
...
+      "(prefers-color-scheme: dark)",
+    );
+
+    if (prefersDark) {
+      de.classList.add("dark");
+    } 
+
+    localStorage.removeItem("theme");
+  } else {
+    de.classList.add(theme);
```
eranhirsch, line 27: You'll add the "light" class here in case the user selects "light" as a theme and never remove it, even if the user selects "dark" later on. This wouldn't have any impact, but it's funky. Let's only add the "dark" class.

 ```suggestion
    if (theme === "dark") {
      de.classList.add("dark");
    }
```

File: docs/src/components/theme-switcher.tsx [hunk 28]
```diff
@@ -0,0 +1,55 @@
...
+    localStorage.setItem("theme", theme);
+  }
+};
+
+export function ThemeSwitcher() {
+  return (
+    <DropdownMenu modal={false}>
+      <DropdownMenuTrigger asChild>
+        <Button variant="ghost" size="icon">
+          <SunIcon className="scale-100 dark:invisible" />
```
eranhirsch, line 37: You don't need scale-100 here, it's implied.
```suggestion
          <SunIcon className="dark:invisible" />
```

docs/src/layouts/layout.astro
```diff
@@ -0,0 +1,24 @@
+---
+import "@/styles/globals.css";
+import Header from "@/components/header.astro";
+import ThemeScript from "@/components/theme-script.astro";
+---
+
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="description" content='The first "data-first" and "data-last" utility library designed specifically for TypeScript' />
```
eranhirsch, line 10: ❤️

docs/src/styles/globals.css
```diff
@@ -0,0 +1,64 @@
...
+    @apply border-border;
+  }
+
+  body {
+    @apply bg-background text-foreground;
+  }
+
+  code {
+    @apply text-primary;
+  }
```
eranhirsch, line 63: This could be moved to the same style you added for the Code element.

docs/src/components/function-card.astro
```diff
@@ -0,0 +1,57 @@
...
+    <div class="max-w-prose text-balance text-sm text-muted-foreground">
+      {func.description && <p set:html={func.description} />}
+    </div>
+  </CardHeader>
+
+  <CardContent className="grid gap-6">
+    {
+      func.methods.map(
+        (method) =>
+          !!method.signature && (
```
eranhirsch, line 39: don't use javascriptisms... prefer an explicit check:
```suggestion
          method.signature !== undefined && (
```

docs/src/components/function-card.astro
```diff
@@ -0,0 +1,57 @@
...
+            <div class="flex flex-col gap-2">
+              {!!method.tag && <p>{method.tag}</p>}
```
eranhirsch, line 41: similarly
```suggestion
              {method.tag !== undefined && <p>{method.tag}</p>}
```

docs/src/components/function-card.astro
```diff
@@ -0,0 +1,57 @@
...
+
+              <MethodSignature
+                args={method.args}
+                returns={method.returns}
+                client:load
+              >
+                <CodeBlock code={method.signature} />
+              </MethodSignature>
+
+              {!!method.example && <CodeBlock code={method.example} />}
```
eranhirsch, line 51: and here
```suggestion
              {method.example !== undefined && <CodeBlock code={method.example} />}
```

eranhirsch (approved):
Looks great! Thanks for doing this, I know this must have taken a long time to get done!

eranhirsch:
Also, please run `npx prettier . --write` from the `/docs` directory. Most of the code isn't formatted correctly.

docs/src/styles/globals.css
xanderbarkhatov: The <code> tag is used in generated descriptions, so I styled it and decided to also use it on the Home page for consistency.

xanderbarkhatov:
@eranhirsch submitted a new PR to fix formatting in the examples. Don't know why it changed.

PR #488, opened 2024-02-06
fix(docs): fix examples indentation

xanderbarkhatov:
Fixes indentation and semicolons in the examples on the home page


eranhirsch (approved):
thanks, ideally i'll move all this page to markdown now that we have astro that would support it

PR #489, opened 2024-02-06
docs: change copy button icon color to make it visible in light mode

xanderbarkhatov:


docs/src/pages/index.astro
```diff
@@ -30,7 +30,7 @@ const commands = [
                 <CodeBlock code={command} lang="bash" />
 
                 <div class="absolute right-2 top-1/2 -translate-y-1/2">
-                  <CopyButton value={command} client:load />
+                  <CopyButton value={command} className='text-background dark:text-foreground' client:load />
```
eranhirsch, line 33: This isn't formatted correctly
```suggestion
                  <CopyButton value={command} className="text-background dark:text-foreground" client:load />
```

eranhirsch (approved):

PR #490, opened 2024-02-06
fix(docs): make function card content scrollable on small screens

xanderbarkhatov:
It was working in the original version. Guess something changed with the fixes.

image
image

### DIFF
--- a/docs/src/components/function-card.astro
+++ b/docs/src/components/function-card.astro
@@ -32,7 +32,7 @@ const tags = getTags(func);
     </div>
   </CardHeader>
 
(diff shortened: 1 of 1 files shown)


xanderbarkhatov:
I am sorry 🥲. I think this is the last one.

eranhirsch:
Check out #491, I prefer not having hidden overflows.

xanderbarkhatov:
And it was not hidden in my fix. It was auto, which means it could be scrolled. I just have scrollbars hidden in the OS

PR #491, opened 2024-02-06
docs: fix overflow issue in examples

eranhirsch:
Mobile view is currently unusable because it takes the width of the longest example. This makes the mobile experience better by allowing wrapping, limiting containers to the screen width, and decreasing font size.

### DIFF
--- a/docs/src/components/function-card.astro
+++ b/docs/src/components/function-card.astro
@@ -15,7 +15,7 @@ const { func } = Astro.props;
 const tags = getTags(func);
 ---
 
-<Card id={func.name} className="scroll-mt-20">
+<Card id={func.name} className="scroll-mt-20 overflow-hidden">
   <CardHeader>
     <div class="flex items-center gap-2">
       <CardTitle className="text-2xl">{func.name}</CardTitle>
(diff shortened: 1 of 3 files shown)


eranhirsch:
@xanderbarkhatov, I prefer this impl

xanderbarkhatov:
Looks good, but I think it broke the layout and the sidebar is not sticky anymore

docs/src/layouts/layout.astro
```diff
@@ -20,7 +20,7 @@ import ThemeScript from '@/components/theme-script.astro';
   <body>
     <Header />
 
-    <div class="container">
+    <div class="container overflow-hidden">
```
xanderbarkhatov, line 23: This breaks the layout

docs/src/components/function-card.astro
```diff
@@ -15,7 +15,7 @@ const { func } = Astro.props;
 const tags = getTags(func);
 ---
 
-<Card id={func.name} className="scroll-mt-20">
+<Card id={func.name} className="scroll-mt-20 overflow-hidden">
```
xanderbarkhatov, line 18: This overflow-hidden breaks scroll-mt-20 (scroll offset to compensate for the header height)

xanderbarkhatov:
@eranhirsch I think we can safely get rid of overflow-hidden in these 2 cases. Should I do it or you will?

eranhirsch:
Sorry for messing it up, i worked my way down the stack. If you can send a pr that would be very helpful

PR #505, opened 2024-02-11
docs: link function sources

cjquines:
add a link to the function source on the top-right of the function card

image

### DIFF
--- a/docs/src/components/function-card.astro
+++ b/docs/src/components/function-card.astro
@@ -22,26 +22,25 @@ const { func } = Astro.props;
     <div class="flex items-center justify-between gap-2">
       <CardTitle className="text-2xl">{func.name}</CardTitle>
 
-      <div class="flex gap-2">
-        <Badge className="ml-auto bg-primary text-primary-foreground">
-          {func.category}
-        </Badge>
-        {
-          func.source !== undefined && (
-            <a
-              href={func.source}
-              target="_blank"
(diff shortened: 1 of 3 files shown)


docs/src/lib/transform.ts
```diff
@@ -43,7 +44,9 @@ function transformFunction({
 
   const methods = signaturesWithComments.map(transformSignature);
 
-  return { id, name, description, methods };
+  const source = sources?.[0]?.url;
+
+  return { id, name, description, methods, source };
```
eranhirsch, line 49: A better name would be `sourceUrl`, no?
```suggestion
  const sourceUrl = sources?.[0]?.url;

  return { id, name, description, methods, sourceUrl };
```

docs/src/components/function-card.astro
```diff
@@ -25,6 +27,20 @@ const { func } = Astro.props;
...
+
+      {
+        func.source !== undefined && (
+          <a
+            href={func.source}
+            target="_blank"
+            class={buttonVariants({ variant: "secondary", size: "sm" })}
+            title="View source on Github"
+          >
+            <ExternalLinkIcon />
```
eranhirsch, line 39: Use the Github logo, it's the standard for linking to GitHub
```suggestion
            <GitHubLogoIcon />
```

eranhirsch (approved):
Amazing! I wanted to do this for a while now!

PR #506, opened 2024-02-12
fix(pick|omit): Fix type inference in pipe

### LINKED ISSUE #504: Type autocomplete on pick does not populate with 2+ keys when used in pipe
Hello! I wasn't sure if these functions were meant to be piped, since the documentation on the hosted website did not label them as such but the readme uses pick in a pipe example.

Related: https://github.com/remeda/remeda/issues/408

# Reproduction
```typescript
type Data = { aProp: string; bProp: string };
const obj: Data = {
  aProp: 'p1',
  bProp: 'p2',
};
const result = pipe(obj, pick(['aProp', 'bProp']));
```
Pasted image 20240210163808

# Environment
Typescript Version: 5.3.2
[Sandbox (direct fork of repository)](https://codesandbox.io/p/devbox/3lj622?file=%2Fsrc%2Fpick.test.ts%3A81%2C21)

# Expected
The pick function, and similarly, omit, behave as expected when used alone and when used in pipe with one key provided, with full autocomplete support.
# Received
Autocomplete fields are not populated for any additional keys after the first when used in a pipe. 

# Suggested Fix
- The issue seems to be tangentially related to how the Typescript compiler struggles with contextual argument type inference with generics when overloads or currying are involved due to an inherent design limitation as suggested in the following issues: https://github.com/microsoft/TypeScript/issues/30369#issuecomment-476402214
https://github.com/microsoft/TypeScript/issues/46940
- Inverting the order of the overloads will allow autocompletion with multiple keys in a pipe. I am guessing that being able to [short-circuit](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#ordering) at the first match simplifies the compiler's overload resolution strategy.

Pasted image 20240210163703

Please let me know if I may be missing anything. I am willing to make a PR to make any needed changes or update docs.

tbui17:
closes #504

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/pick.test.ts
+++ b/src/pick.test.ts
@@ -75,4 +75,18 @@ describe('typing', () => {
       >();
     });
   });
+
+  test('multiple type', () => {
+    type Data = { aProp: string; bProp: string };
+
+    const obj: Data = {
+      aProp: 'p1',
+
+      bProp: 'p2',
+    };
+
+    const result = pipe(obj, pick(['aProp', 'bProp']));
+
+    expectTypeOf(result).toEqualTypeOf<Pick<Data, 'aProp' | 'bProp'>>();
+  });
 });
--- a/src/pick.test.ts
+++ b/src/pick.test.ts
@@ -76,7 +76,7 @@ describe('typing', () => {
     });
   });
 
-  test('multiple type', () => {
+  test('multiple keys', () => {
     type Data = { aProp: string; bProp: string };
 
     const obj: Data = {
--- a/src/omit.test.ts
+++ b/src/omit.test.ts
@@ -54,4 +54,18 @@ describe('typing', () => {
       >();
     });
   });
+
+  test('multiple keys', () => {
+    type Data = { aProp: string; bProp: string };
+
+    const obj: Data = {
+      aProp: 'p1',
+
+      bProp: 'p2',
+    };
+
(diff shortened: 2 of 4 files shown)


src/pick.test.ts
```diff
@@ -75,4 +75,18 @@ describe('typing', () => {
       >();
     });
   });
+
+  test('multiple keys', () => {
```
eranhirsch, line 79: This test doesn't test what we want it to because it succeeds without any modifications to the implementation.

The thing is that the bug isn't even exactly a Remeda issue, it's more likely a Typescript issue - You can see that Typescript does eventually pick (no pun intended) the correct overload of the function when you finish typing; it would continue to type-check that the prop names given are actual keys of the data object (you can test this by adding this test to the unmodified version of `pick` and changing any of the props to an invalid name, or adding one in addition).

As far as I can tell, the only bug here seems related specifically to auto-completion...

eranhirsch (approved):
See my comment inline, this bypasses the issue, but as far as I can tell, the real bug here is with vscode/Typescript auto-complete not running the inference correctly.

Regardless, thanks for doing this!

src/pick.test.ts
tbui17: Thanks! I agree that it's certainly not a bug with Remeda but rather Typescript.

PR #510, opened 2024-02-13
feat(only): Add `only`

### LINKED ISSUE #467: Add `findSingle`
A util I've written in several ways already: Return an element if and only if that element is the only one where the predicate is true, and undefined otherwise. `R.findSingle(items, fn)` would be equivalent to:
```ts
R.filter(items, fn).length === 1 ? R.find(items, fn) : undefined
```
but pipeable / with short-circuiting.

Example use cases:
* I have an array of command line options, which should contain exactly one of some list, and I want to show an error if there are none or more than one.
* I have an array that should have a single element I want to get: `R.findSingle(a, () => true)`.

cjquines:
closes #467 

i don't think `only` can be made lazy; here's the argument. when called with a value, we either:

* return it immediately as `next`. later if we get another value we can't "take back" the value we gave before
* don't return it as `next`. but then we don't get called again, so there's no way for the caller to get the value

there are changes that can be made to the architecture to allow it to be lazy. the one that needs least changes, i think, is to change `pipe` to be

```diff
    if ((lastLazySeq as any).single) {
-      ret = acc[0];
+      ret = acc[acc.length - 1];
    } else {
      ret = acc;
    }
```

this lets `only` "take back" a value it returned, and replace it with `undefined` (and, surprisingly to me, all tests pass after this change). i don't like it though

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/only.test.ts
+++ b/src/only.test.ts
@@ -1,0 +1,129 @@
+import { only } from './only';
+
+describe('strict typing', () => {
+  test('simple empty array', () => {
+    const arr: Array<number> = [];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+    expect(result).toEqual(undefined);
+  });
+
+  test('simple array', () => {
+    const arr: Array<number> = [1];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+    expect(result).toEqual(1);
+  });
+
+  test('simple non-empty array', () => {
+    const arr: [number, ...Array<number>] = [1];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+    expect(result).toEqual(1);
+  });
+
+  test('simple tuple', () => {
+    const arr: [number, string] = [1, 'a'];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf<undefined>();
+    expect(result).toEqual(undefined);
+  });
+
+  test('array with more than one item', () => {
+    const arr: [number, number, ...Array<number>] = [1, 2];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf<undefined>();
+    expect(result).toEqual(undefined);
+  });
+
+  test('trivial empty array', () => {
+    const arr: [] = [];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf(undefined);
+    expect(result).toEqual(undefined);
+  });
+
+  test('array with last', () => {
+    const arr: [...Array<number>, number] = [1];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+    expect(result).toEqual(1);
+  });
+
+  test('tuple with last', () => {
+    const arr: [...Array<string>, number] = ['a', 1];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf<string | number | undefined>();
+    expect(result).toEqual(undefined);
+  });
+
+  test('tuple with two last', () => {
+    const arr: [...Array<string>, number, number] = ['a', 1, 2];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf<undefined>();
+    expect(result).toEqual(undefined);
+  });
+
+  test('tuple with first and last', () => {
+    const arr: [number, ...Array<string>, number] = [1, 'a', 2];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf<undefined>();
+    expect(result).toEqual(undefined);
+  });
+
+  test('simple empty readonly array', () => {
+    const arr: ReadonlyArray<number> = [];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+    expect(result).toEqual(undefined);
+  });
+
+  test('simple readonly array', () => {
+    const arr: ReadonlyArray<number> = [1];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+    expect(result).toEqual(1);
+  });
+
+  test('simple non-empty readonly array', () => {
+    const arr: readonly [number, ...Array<number>] = [1];
+    const result = only(arr);
+    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+    expect(result).toEqual(1);
+  });
+
+  test('simple readonly tuple', () => {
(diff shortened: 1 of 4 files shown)


src/only.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type Only<T extends IterableContainer> = T extends
+  | []
+  | readonly [unknown, unknown, ...Array<unknown>]
+  | readonly [unknown, ...Array<unknown>, unknown]
+  | readonly [...Array<unknown>, unknown, unknown]
+  ? undefined
+  : T extends readonly [unknown]
+    ? T[0]
+    : T[0] | undefined;
```
eranhirsch, line 12: Typing is hard! This could either be `T[0]` or `T[number]` and each one of them gets right some cases (`[string, boolean?]`) and gets some cases wrong (`[string?, boolean?]`, `[string?, ...boolean[]`), and vice-versa.

I think the `T[number]` is a little broader and perhaps more "real-life" useful.

```suggestion
    : T[number] | undefined;
```

Here's my playground: https://www.typescriptlang.org/play?#code/PTAEBcE8AcFNQPIDsA2kA8AVUsAe5YkATAZ1ACdYBDIge1UlAFckBrJWgdyQG0BdUAB9Q-AHygAvKGx4CxMpRr00Iluy5I+AKBCg9Afmk8ADNt16AXNJz5CpHWD1OhIs4+cvFdBqrYduADTMfhpBAHQRav68fG4eesJeyow8UaGgEWFp3PxB2ZoO8Z7U3io8mfm5weqB1dFxBsFEsABmAJZIsESFlkZITAC2AEaw5ALCLM3tnUQA3FpaUHCIDFg2cqQUJcl1GvwuYpLWsnYkWgmu58VKPqkhtfnhEQCC5ORUGPmi2hdJt48ZF5vD7oL55e4FX7bW6ZV7vT4Q0TgmpIZH1K6GSatDpdK5WGS2eRbG5lfI-JyGTAmcm9KmmFxY6ZdeYLJbwACyVGgcCIaxORIASrAAMa0ci8kjgcgdADmQT+Kkq41conEUgA3lceABpUAdUCsWCQWgtaR8KzINBYHV8UTzAC+LLZoAAGgBGI6c7lddCavS6AB6+iuwqoJFgbqs-FmoF0jJx3T0ofDACYrAqUnwY3HiNiZiGw7AAMxRyXSpAyrOxsBl2UF8MAFnT0LKtYrVd0bZlWntdtZMHgLpTnq5PN9V0DwaThYArKWpbKgkNaLQUNRNNmwPH89PwwA2ZsklJdpcrtdUDfVpp53ETsCYAAWsHDoG8AHJwKBOGLWHqkKABkYNoBmgFB61gAB2edyzlUBl1Xdd9A7GsFwrBlcyZRNQGTWAAA5D1KY9UNg+DzyQJDN1ALt0KmBMez7RYB1dIsR29Xk-SvIM72kJ8X3fT9v3IX99XAJ82nIPUQLA3dYAATmg2V9FPBCLwoq9qOEUj1xom8sJwt1jAInYeC7JS4LPRDkKo4iXC0i8dMwhYZLdSNrJg6N1JsiYMLo5y02JQi3NlDzOy869HN7J0mJdBtWLHDjJ3At0SxELtlR4OzL1CmDbIsi8kqbALjLS65AoyvKspQnLNIqpK51S4j0syoJ+mGUYrI08LfOwws3QPIrbhKxIWxSZrQFakYxkozrt1xSL+2WF0Zzin0ErALjnKghqYKCE9ATCLt+Cs2a9N6-CBtbYjdqu-bDtiTcTqS+TtsXW7iKqNLKMemSU0Mi6iJ2t73L4a6YOOnz83mxjFr3Fb2O4jaetTVzygiO7Qdez6r2+pHYBTfyMxETJ0aCisMfbL6IdxH6UpMm69tM8HaJ3XGU0Kwm6cBhniLUnNmbmhjnRdCC4fHf11qnK9H2feB+K-H8-wgMSJOA0DwJTerOcUp4wkyo7prCzKHO6nCU36jnTJ1vX7s86rzJU-9vP57p5qAA

src/only.ts
```diff
@@ -0,0 +1,42 @@
+import { IterableContainer } from './_types';
+import { purry } from './purry';
+
+type Only<T extends IterableContainer> = T extends
+  | []
```
eranhirsch, line 5: this should probably be `readonly` too. I think the readonly case works right now but falls down to the last case (where it's handled trivially correctly).

```suggestion
  | readonly []
```

src/only.ts
```diff
@@ -0,0 +1,42 @@
...
+  | readonly [unknown, unknown, ...Array<unknown>]
+  | readonly [unknown, ...Array<unknown>, unknown]
+  | readonly [...Array<unknown>, unknown, unknown]
```
eranhirsch, line 8: brilliant :clap:

src/only.ts
```diff
@@ -0,0 +1,42 @@
...
+ *    R.only(array)
+ * @example
+ *    R.only([]) // => undefined
+ *    R.only([1]) // => 1
+ *    R.only([1, 2]) // => undefined
+ * @pipeable
+ * @category Array
+ */
+export function only<T extends IterableContainer>(array: Readonly<T>): Only<T>;
+export function only<T extends IterableContainer>(): (
```
eranhirsch, line 29: Missing documentation for the dataLast impl

src/only.ts
```diff
@@ -0,0 +1,42 @@
...
+ * inferred type is lost.
+ * @param array the target array
+ * @signature
+ *    R.only(array)
+ * @example
+ *    R.only([]) // => undefined
+ *    R.only([1]) // => 1
+ *    R.only([1, 2]) // => undefined
+ * @pipeable
+ * @category Array
```
eranhirsch, line 26:
```suggestion
 * @category Array
 * @dataFirst
```

src/only.test.ts
```diff
@@ -0,0 +1,158 @@
+import { only } from './only';
+import { pipe } from './pipe';
+
+describe('data first', () => {
+  test('empty array', () => {
+    expect(only([])).toEqual(undefined);
```
eranhirsch, line 6:
```suggestion
    expect(only([])).toBeUndefined();
```

eranhirsch (approved):
Looks good, check out my suggestion for a slight change in the typing for optional tuple elements. I usually don't consider them because they are hard to type and not very common, but the fix here is simple.

Consider adding some type tests for optional items.

src/only.ts
cjquines: good catch—added some tuples with optionals to the tests too

eranhirsch (approved):

PR #511, opened 2024-02-13
docs: improve sidebar opening perf by disabling modal mode

xanderbarkhatov:
Because **react-remove-scroll** causes a reflow and /docs page is huge, sidebar opens really slow in modal mode.

Also changed hrefs in navbar so links work correctly from home page.


eranhirsch (approved):
Good catch! Thanks!

PR #513, opened 2024-02-14
docs: link function names

### LINKED ISSUE #512: docs: parse @see tags
* parse @see tags
* edit existing @see tags [so that they have a description](https://jsdoc.app/tags-see)

would be nice to get this to work with linking

cjquines:
somewhat closes #512 

it doesn't parse the `@see` tags because i wasn't quite sure how to style it, but we now link `@link` tags

instead of mentioning a remeda function like `` `toPairs` `` we can now write `{@link toPairs}` (no surrounding backticks)

Peek 2024-02-13 23-12


docs/src/lib/transform.ts
```diff
@@ -49,6 +48,23 @@ function transformFunction({
...
+  summary: Array<JSONOutput.CommentDisplayPart>,
+): string | undefined {
+  if (summary.length === 0) {
+    return undefined;
+  }
+  return summary
+    .map((part) => {
+      const { kind, text } = part;
+      if (kind === "inline-tag" && part.tag === "@link") {
+        return `[\`${text}\`](#${text})`;
```
eranhirsch, line 61: It's hard to notice the little hash sign before the text for the link; extract a function for building the URI and add some documentation to describe what's going on. Also, we probably don't want to completely block the ability to add links to external sites, perhaps we should check that the text doesn't start with any URL prefix (like / or https://) and only add the hash when it doesn't.

```ts
function linkHref(raw: string): string {
  if (raw.startsWith("http") || raw.startsWith("/")) {
    return raw;
  }
  // We assume that any simple link is a reference to a function...
  return `#${raw}`;
}
```

```suggestion
       return `[\`${text}\`](${linkHref(text)})`;
```

eranhirsch (approved):
Looks good, I'm a little concerned that: 1. people won't know about this or remember to use it, 2. that it would break for other use-cases of @link, and 3. that it would break inline documentation in the IDE (that right now is broken because we strip comments during the build, but we need to fix that). The last point could be tested in the development environment, just hover over a function to se the docs, and see what happens when you click on one of these links.

A different approach here would be to search for `<FUNCTION_NAME>` in the description block and perform the conversion on the spot, and not in the jsdocs themselves...

docs/src/lib/transform.ts
```diff
@@ -49,6 +48,23 @@ function transformFunction({
...
+function transformSummary(
+  summary: Array<JSONOutput.CommentDisplayPart>,
+): string | undefined {
+  if (summary.length === 0) {
+    return undefined;
+  }
+  return summary
+    .map((part) => {
+      const { kind, text } = part;
+      if (kind === "inline-tag" && part.tag === "@link") {
```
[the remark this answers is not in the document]
cjquines: `tag` does not exist CommentDisplayPart until narrowed

docs/src/lib/transform.ts
cjquines: if we wanted this to be a bit safer, we can pass through `target`; i made this change

cjquines:
performing the conversion on the spot sounds okay too—we'd have to avoid doing it in code blocks though, but otherwise it'd work i think

is there a reason we strip comments during build?

eranhirsch:
> is there a reason we strip comments during build?

@cjquines - It was done about a year ago by @TkDodo in response to some feedback, but I don't remember the details

eranhirsch (changes requested):
I prefer we don't introduce the {@link ...} notation to the inline JSDocs at the moment. It looks like that tag is not meant directly for these kinds of links. Maybe there's some other way that is built into JSDoc to refer to another symbol inside the same project?

in the meantime, let's just replace anything that starts with a backtick, and contains just a known function name, and then ends with a backtick with a link, and leave anything else as-is.

docs/src/lib/transform.ts
```diff
@@ -12,19 +12,24 @@ export function transformProject(project: typeof DATA) {
...
+    ({ kind }) => kind === ReflectionKind.Function,
+  );
+
+  const functionNames = new Set(functions.map(({ name }) => name));
 
-  return addCategories(project, functions.filter(isDefined.strict));
+  return addCategories(
+    project,
+    functions
+      .map((function_) => transformFunction(function_, functionNames))
```
eranhirsch, line 24: I prefer an abbreviation over adding the suffix underscore.
```suggestion
      .map((func) => transformFunction(func, functionNames))
```

docs/src/lib/transform.ts
```diff
@@ -49,6 +53,28 @@ function transformFunction({
...
+      if (kind !== "code") {
+        return text;
+      }
+      const codeContent = text.slice(1, -1);
+      // If this is a function name, link to its anchor:
+      if (functionNames.has(codeContent)) {
+        return `[${text}](#${codeContent})`;
+      }
+      return text;
+    })
```
eranhirsch, line 74: Instead of pulling out the whole transformSummary, i'd pull out only this part as it's own utility function transformCommentDisplayPart or something like that, so we can reuse it for other parts if we ever want to.

It's more a pet peeve of mine, but I usually don't like utility functions which just wrap a "map" or "filter" which does most of the logic, in those cases I usually prefer the inner function be pulled out instead.

```suggestion
    .map((part) => transformCommendDisplayPart(part, functionNames))
```


```ts
function transformCommentDisplayPart(
  { kind, text }: JSONOutput.CommentDisplayPart,
  functionNames: ReadonlySet<string>,
): string {
  if (kind !== "code") {
    return text;
  }

  const codeContent = text.slice(1, -1);

  return functionNames.has(codeContent)
    // If this is a function name, link to its anchor:
    ? `[${text}](#${codeContent})`
    : text;
}

eranhirsch (approved):
Looks great! I think this is very useful.

PR #515, opened 2024-02-15
feat: add `hasProp`, `hasPropSatisfying`, and `isEqual`

### LINKED ISSUE #461: Object property predicates like `hasProp`
I often find myself wanting to `filter` an array based on whether or not a property exists or that has a property of a certain value (like when working with discriminated unions).

I'm wondering if there's interest in adding some helpers like `hasNonNilProperty` or `hasPropertyWithValue` to the library. I'm happy to work on a PR if this seems like something reasonable to add.

~~I made a [Typescript Playground](https://www.typescriptlang.org/play?#code/PQkAgUQDwQwWwA4BsCmBnMB7AZmBAnFAEwEsBjGAF3TEoAsqwB3TAVySLACMUwA7cr0qYwDAG686KJAhT4M2TPgBQyygE9ZYACqYiIgLxhlYMAG8waSlVZoAXGADkZTIlTUijgDRgXblB4AgpQOACJUKD7UUCGWlPgkfADmPqwIRBFEAELqAPwOVgnJYAC+JmAAPuZxNvZOfJiUALR+yAHE3rQoMQXxiSlgaRkeOflxRUmlqi58VrR6mBhGANoAumAwGLr6awDcyjNzre7E24tgRsL6aOUAdNgkSNT4ABQvV5gAlA4fYCQYGlkOHm+jAADJzFZag5nK42h5HCULgA+EGYW5Qyi2C4GIyw-wIz4HTCzSi+OEnIjBC7kgmnBZoZYABlWt2O7SplGU4FMYAAerlpiS5h80AB1Ej0ACq6UyAAV8JhZPgNDTRXcHk85G8Pt80X8AZoUMDfhCAEooACOrBIhCIAB45eQANb2wHG3AfHyOIaZHKOZGogyogCEIY+t19I3URMOZKjxByaoZEulso8CqVcg0zNZCey6n2PNMAtUoHAqboG1EmwAciTa49M8qNAAJaTK1QkRBKMkAKg2GDNYGwirgTkIcGIMEc+2U3QQvdoRrAlfrfFr7CQMC4qGb2ZI6Ht2h8AGkwN1qHwiBhnSh1MDtEGdODyhbrbbiPazOVTMtT-efx8GAp6rA466bkg267igx7-veqzIvspglEhagrgAajASAkEQAHqDShT9JU-CsHAPD4CRaDqORmBIHOcY1mgEFNoqLbqAAkm4NIvJgXAAFYODAfDqD4d7qEJImfCi5jlIQWL4MBfH8cs4nrCGuKDNeKAPHwxDghCymqQhYAaUYfBbvsJT7AuS7YKwfBkJQJAkkxLFIPuKrqMeZ4XjEKDXre96PsiLzlMpDgnuU4kOKeXjKHqykGqukp0BBW47nubEHkeJ4gWhtkqiODlOS5wEMMxDasVmXn2uel4BTeYBYTheH3qF0X3rF8V6seflXk1FjGQRiQgasYzCQRqFhaYEU6D1MlJf8KX0OlUGZSgnnOblZ4FVAi5FfZjnOa5FXuVt6gvLc12EFYknqGs0k-qY8msIpYBmrcCBvfgl1nVVHnZV5XHID4t2UJ8VlCqSaLiqlMrDLwlwMvcjzPC8-0btV7EvAARPmOS458kPEjDCbJtclYIxEuaRumiaFtywC8vyuRAA) with a couple contrived examples of the type of predicates I'm hoping to add helpers for.~~ edit: see next comment for playground

airjp73:
Resolves #461 

## Summary

Adds three new helpers:

* `isEqual` does a shallow, `Object.is` equality comparison and acts as a type guard
* `hasProp` checks if a given object has a property and that it's defined
* `hasPropSatisfying` checks if a given object has property, that it's defined, and that the provided predicate/type guard returns true.

### `isEqual`

I originally thought to just do an `===` comparison, but I thought [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) made more sense since it's more robust to values like `NaN`.

### `hasProp`

My original plan was to _only_ check for the presence of a property and allow checks like `hasProp({ a: undefined }, 'a')` to return true.

However, Typescript removes `undefined` from the possible values when removing the optionality of the prop ([related github issue](https://github.com/microsoft/TypeScript/issues/31025)). The linked thread has some workarounds, but I wasn't able to get any of them to work in a general enough way for this helper.

This means that the narrowed type resulting from the type guard couldn't be accurate if we returned true for present keys with `undefined` as the value.

I think returning false when the value is `undefined` is a reasonable trade-off to make to keep the types completely accurate.

### `hasPropSatisifying`

Same considerations as `hasProp`, but the type accuracy issue also extends to the type of the argument passed to the predicate.

## Potential future improvements

Lodash and Ramda both have the ability to check for properties at a given path.

* Lodash's [`has`](https://lodash.com/docs/4.17.15#has) supports it directly
* Ramda has a separate [`hasPath`](https://ramdajs.com/docs/#hasPath) helper

As a future improvement, we could conceivably add a couple `hasPropAtPath` and `hasPropAtPathSatisfying` helpers.

---

Make sure that you:

---

### Available Types:

existing).

</details>


src/isEqual.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @returns a function that returns true if the the two values are equal.
+ * @signature
+ *    R.isEqual(specificValue)(wideValue);
+ * @example
+ *   [1, 2, 3].filter(R.isEqual(2)); // => [2]
+ * @dataLast
+ * @category Guard
+ */
+export function isEqual<WideValue, SpecificValue extends WideValue>(
+  specificValue: Narrow<SpecificValue>
```
eranhirsch, line 18: Are there specific cases where the Narrow utility is needed? If not, let's keep things simple here and not use this utility. I'm not sure there are real-life scenarios where it's useful.
```suggestion
  specificValue: SpecificValue
```

src/isEqual.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param specificValue the second value to compare. This type should be narrower than the first value.
+ * @returns true if the the two values are equal, false otherwise
+ * @signature
+ *    R.isEqual(wideValue, specificValue);
+ * @dataFirst
+ * @category Guard
+ */
+export function isEqual<WideValue, SpecificValue extends WideValue>(
+  wideValue: WideValue,
+  specificValue: Narrow<SpecificValue>
```
eranhirsch, line 35: similarly,
```suggestion
  specificValue: SpecificValue
```

src/isEqual.ts
```diff
@@ -0,0 +1,47 @@
+import { Narrow } from './_narrow';
+import { purry } from './purry';
+
+/**
+ * Returns a function that determines if a value is _shallowly_ equal to the specified value using Object.is equality,
```
eranhirsch, line 5: No need to mention that it returns a function; that is what the dataLast designation means.
```suggestion
 * Determines if a value is _shallowly_ equal to the specified value using Object.is equality,
```

src/isEqual.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  // @ts-expect-error these types don't overlap
+  expect(isEqual(symbol, Symbol('hello'))).toBe(false);
+
+  const obj = {};
+  expect(isEqual(obj, obj)).toBe(true);
+  expect(isEqual(obj, {})).toBe(false);
+
+  expect(isEqual(2, 3)).toBe(false);
+  // @ts-expect-error these types don't overlap
+  expect(isEqual(2 as 2 | 4, 3)).toBe(false);
```
eranhirsch, line 21: Split these into separate tests under a `describe` block so that each test passes or fails on it's own, and has a name that describes it's semantics.

src/isEqual.test.ts
```diff
@@ -0,0 +1,47 @@
+import { isEqual } from './isEqual';
```
eranhirsch, line 1: Add tests that test the more interesting parts of your implementation too, specifically I'd like to see some more tests for arrays and objects and for narrowing of types.

eranhirsch (changes requested):
1. I think it's unlikely that people would consider an array as having a prop at an index. We already offer `hasAtLeast` to narrow arrays so that accessing a specific prop would result in a non-undefined value. Handling arrays in these functions complicates their type considerably and might introduce edge-cases and bugs that would be hard to find and fix. Let's keep `hasProp` and `hasPropSatisfying` for proper objects just as pick, omit, and prop are.

2. Handling unions specifically is also beyond what the utility functions should handle. We don't handle these issues with `prop`, `pick`, or `omit` either. If this is to handle discriminated unions, it's usually recommended to have a pivot prop that would allow typescript to align the types. I personally encountered issues in the past with trying to rely on prop existence in order to discriminate a union. Let's keep this simple for now, and we can always go back to this and expand it if we see that the simple solution isn't working for people.

src/_objectProps.ts
```diff
@@ -0,0 +1,37 @@
+import { Simplify } from './type-fest/simplify';
+
+export type AllUnionKeys<Union> = Union extends Union ? keyof Union : never;
+
+type MakeKeyRequired<Obj, Prop extends keyof Obj> = Obj & {
```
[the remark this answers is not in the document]
airjp73: This was my original approach but ran into issues with the guard types. Typescript was thinking that this type was no longer assignable to the original.

src/_objectProps.ts
```diff
@@ -0,0 +1,37 @@
+import { Simplify } from './type-fest/simplify';
+
+export type AllUnionKeys<Union> = Union extends Union ? keyof Union : never;
```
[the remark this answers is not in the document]
airjp73: Yeah, this is one of those weird TS types that looks like it doesn't do anything but actually does. There's a little bit more info [here](https://stackoverflow.com/a/49402091).

Example:
```ts
type SomeUnion = { a: string, b: string } | { b: string, c: string };

type RegularKeys = keyof SomeUnion; // 'b'

type AllKeys = AllUnionKeys<SomeUnion>; // 'a' | 'b' | 'c'
```

src/isEqual.ts
airjp73: This is specifically for the discriminated union case. Something like `hasPropSatisfying(union, 'type', isEqual('foo'))` is meaningless without a narrow type here.

airjp73:
> 1. I think it's unlikely that people would consider an array as having a prop at an index. We already offer `hasAtLeast` to narrow arrays so that accessing a specific prop would result in a non-undefined value. Handling arrays in these functions complicates their type considerably and might introduce edge-cases and bugs that would be hard to find and fix. Let's keep `hasProp` and `hasPropSatisfying` for proper objects just as pick, omit, and prop are.

Good callout. I didn't realize that was there! I only added the type handling because it the implementation technically supports it without extra effort.

> 2. Handling unions specifically is also beyond what the utility functions should handle. We don't handle these issues with `prop`, `pick`, or `omit` either. If this is to handle discriminated unions, it's usually recommended to have a pivot prop that would allow typescript to align the types. I personally encountered issues in the past with trying to rely on prop existence in order to discriminate a union. Let's keep this simple for now, and we can always go back to this and expand it if we see that the simple solution isn't working for people.

I'm all for implementing less complete feature sets and filling it out over time, but I'm not sure I understand the perspective that this is beyond the scope of these helpers.

I agree that `prop`, `pick`, and `omit` shouldn't have this handling. For the same reason that writing `obj.someKey` will error if a key might not exist, `prop` should too.

But since these helpers are type guards, the whole point is to take a wider type and narrow it.

eranhirsch:
Those are good points about it being a type-guard requiring extra care about the types. I think the best thing would be to ensure that there are tests covering these concerns so that we can't accidentally break the code for these cases without even knowing. 

None of the tests break at the moment when I remove the `Narrow` utility from the types in `isEqual`, for example.

src/_objectProps.ts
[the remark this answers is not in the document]
airjp73: Spent a little time trying to add this in and ran into the same issue. I'd need to do some more research to figure out exactly why, but TS gives this error in the `hasProp` implementation file:

```
A type predicate's type must be assignable to its parameter's type.
```

src/isEqual.ts
airjp73: Added test cases that actually fail when this is missing.

src/hasPropSatisfying.test.ts
```diff
@@ -0,0 +1,249 @@
...
+      const obj = { type: 'success', data: { foo: 'bar' } } as
+        | { type: 'success'; data: { foo: string } }
+        | { type: 'error'; error: string };
+
+      if (
+        hasPropSatisfying(obj, 'data', hasPropSatisfying('foo', isEqual('bar')))
+      ) {
+        expectTypeOf(obj).toEqualTypeOf<{
+          type: 'success';
+          data: { foo: string } & { foo: 'bar' };
```
airjp73, line 121: When nesting the helper like this, the type doesn't simplify quite as well. Using `SimplifyDeep` from the typefest files fixes it, but I'm not so sure we should use that for this helper.

I imagine using `Omit` to remove the old version of the value would fix this, but that causes issues in other places as I mentioned in other comments.

But I suspect this is probably an uncommon use-case, so it seems like a reasonable compromise.

airjp73:
In the issue, there was some discussion about the behavior of the `isEqual` function. I kept moving forward with the referential equality implementation and the `isEqual` name for now, since it seemed like shallow and deep equals would have more specific names. Happy to change it you want though.

src/hasPropSatisfying.test.ts
airjp73: After doing some more research, I figured out how to properly address this. So now we're able to get the correct, fully resolved type.

src/_objectProps.ts
```diff
@@ -0,0 +1,42 @@
+import { Simplify } from './type-fest/simplify';
+
+export type AnyValidObjectKey = string | number | symbol;
```
cjquines, line 3: **nit:** is this not `PropertyKey`?

src/_objectProps.ts
```diff
@@ -0,0 +1,42 @@
...
+export type GenericObject = Record<AnyValidObjectKey, unknown>;
```
cjquines, line 4: **suggestion:** we have the same type in `isPlainObject`; so maybe this should be called `PlainObject`?

src/_objectProps.ts
```diff
@@ -0,0 +1,42 @@
...
+type MakeKeyRequired<Obj, Prop extends keyof Obj> = Omit<Obj, Prop> & {
+  // We need to both remove the optional modifier and exclude undefined.
+  // This is because the `-?` modifier will only exclude undefined from the type if the property is optional,
+  // which makes the behavior of the type inconsistent.
+  // More info about this can be found here: https://github.com/microsoft/TypeScript/issues/31025
+  // So excluding `undefined` explicitly also does this for required properties that include undefined.
+  [key in Prop]-?: Exclude<Obj[key], undefined>;
+};
+
+type UnionMemebersWithProp<Obj, Prop extends keyof Obj> = Extract<
```
cjquines, line 20: **nit:** `Memebers` -> `Members`

src/_objectProps.ts
airjp73: TIL

I've always made a custom type for this xD

src/_objectProps.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+type UnionMembersWithProp<Obj, Prop extends keyof Obj> = Extract<
+  Obj,
+  { [key in Prop]?: unknown }
+>;
+
+export type WithRequiredProp<Obj, Prop extends AllUnionKeys<Obj>> = Simplify<
+  MakeKeyRequired<UnionMembersWithProp<Obj, Prop>, Prop>
+>;
```
eranhirsch, line 26: Let's pull this out into the type-fest folder and copy:

https://github.com/sindresorhus/type-fest/blob/main/source/set-required.d.ts

src/hasProp.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    // @ts-expect-error this property does not exist on the object
+    expect(hasProp(obj, 1)).toBe(false);
+    // @ts-expect-error this property does not exist on the object
+    expect(hasProp(obj, fooSymbol)).toBe(false);
+  });
+
+  it("should return false when a prop is present on the object but is undefined", () => {
+    const obj = { b: undefined };
+    expect(hasProp(obj, "b")).toBe(false);
+  });
```
eranhirsch, line 27: This is incorrect and relies on a more permissive type definition. When the typescript project uses `exactOptionalPropertyTypes` an `undefined` value is strictly different than an optional prop not being present. This is also how it works during runtime.

```ts
console.log('a' in { a: undefined }); // "true"
console.log('a' in {}); // "false"
```

eranhirsch (changes requested):
`undefined` is not special when it comes to prop values, it should be treated as a valid value and that should be represented both in the types and the runtime behaviour of hasProp.

See my comment inline in the test file.

src/hasProp.test.ts
airjp73: This behavior was an intentional trade-off. I can play around a bit more since it's been awhile, but I remember something at the type level couldn't reliably distinguish between the two.

eranhirsch: Check out type-fest types for some of your utility types under _objectProps, they most likely can solve most of those issues.

hasProp should work as closely as possible to `"prop" in <obj>` otherwise some users would either misunderstand it's semantics and run into problems in runtime, or need to build their own function for the behaviour they need.

`hasPropSatisfying(obj, prop, isDefined.strict)` can provide the functionality you define for `hashProp` here, but the opposite is not possible.

airjp73: I'll check it out and see what I can do. This is the issue I reference elsewhere in the PR that lead me to this decision. https://github.com/microsoft/TypeScript/issues/31025

cjquines:
hey @airjp73 — now that we released v2, we should point this PR to `main` instead of `master`. do you still intend to work on this?

airjp73:
Hi! Sorry for the radio silence. I'll likely come back and reduce the scope of this, to at least get the new equality helpers through. With predicate inference coming in TS 5.5, I'm finding it hard to motivate myself to finish up the prop helpers.

airjp73:
For sure. I almost wonder if it's reasonable to simplify the types to just use the interference from 5.5? It would still work with older TS versions, but wouldn't have the type narrowing you'd get with 5.5.

eranhirsch:

PR #517, opened 2024-02-15
feat(mapWithFeedback): Add `mapWithFeedback`

### LINKED ISSUE #468: Add `mapWithFeedback` (aka scan, accumulate, partial_sum, runningReduce)
This would be the equivalent of, say, [`scan` from Ramda](https://ramdajs.com/docs/#scan), [`itertools.accumulate` from Python](https://docs.python.org/3.11/library/itertools.html#itertools.accumulate), [`std::partial_sum` from C++](https://en.cppreference.com/w/cpp/algorithm/partial_sum), [`runningReduce` from Kotlin](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/running-reduce.html).

It would have a similar signature to `R.reduce`, but it would also be pipeable, like `R.map`.

Lots of room for bikeshedding the name.

tbui17:
closes #468 

This PR should implement the functionality as specified but there are some issues that may require changes outside the scope of this PR.
 
# Changes
- `mapWithFeedback` implementation
- lazy, lazyIndexed, indexed implementations

# Comments
- Referenced this PR for code style: #445 
- The current implementation uses the same accumulator reference with every iteration. Ramda's scan implementation also has this same behavior so I'm not sure if this is unintended behavior, but it does seem like it'd be quite easy for someone to be caught off guard. Here's an example below.
```typescript
 it("should use the same accumulator on every iteration if it's mutable, therefore returning an array containing {array length} references to the accumulator.", () => {
      const results = mapWithFeedback(
        [1, 2, 3, 4, 5] as const,
        (acc, x) => {
          acc[x] = x;
          return acc;
        },
        {} as Record<string,any>
      );
      const expectedEquality = new Array(5).fill({ '1': 1, '2': 2, '3': 3, '4': 4, '5': 5}) // as opposed to [{ '1': 1 },{ '1': 1, '2': 2 },{ '1': 1, '2': 2, '3': 3 },{ '1': 1, '2': 2, '3': 3, '4': 4 },{ '1': 1, '2': 2, '3': 3, '4': 4, '5': 5 }]
      expect(results).toEqual(expectedEquality);
      const item = results[0];
      expect(results.every(result => result === item)).toBe(true)
    });
```
- Using mapWithFeedback.indexed (or reduce.indexed) with other transformer functions will throw an error. A fix may require changes to the `purry` implementation which would be outside of the scope of this PR.
```typescript
    it("should be compatible with transformer functions", () => {
      expect(() => {
        const result = mapWithFeedback(
          [1,2,3,4,5],
          add,
          100,
        )
        expect(result).toEqual([101, 103, 106, 110, 115])
      }).not.toThrowError("Wrong number of arguments")
    })

    it("indexed should not be compatible with transformer functions", () => {
      expect(() => {
        const result = mapWithFeedback.indexed(
          [1,2,3,4,5],
          add,
          100,
        )
        expect(result).toEqual([101, 103, 106, 110, 115])
      }).toThrowError("Wrong number of arguments")
    })
```
---

Make sure that you:

---

### Available Types:

existing).

</details>


src/index.ts
```diff
@@ -130,3 +130,4 @@ export * from './values';
 export * from './zip';
 export * from './zipObj';
 export * from './zipWith';
+export * from './mapWithFeedback';
```
cjquines, line 133: **nit:** keep this sorted

src/mapWithFeedback.ts
```diff
@@ -0,0 +1,147 @@
...
+import { PredIndexedOptional } from './_types';
+import { purry } from './purry';
+
+/**
+ * A callback function receiving the first two parameters of a reduce operation.
+ */
+type Reducer<T, Accumulator> = (
+  accumulator: Accumulator,
+  currentValue: T
+) => Accumulator;
```
cjquines, line 12: **note:** we inline the same types in `reduce.ts`, and use similar types in `map.ts`... seems like the letters are `(acc: K, item: T) => K`

src/mapWithFeedback.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    return (
+      value: T,
+      index?: number,
+      items?: Array<T>
+    ): LazyResult<Accumulator> => ({
+      done: false,
+      hasNext: true,
+      next: modifiedReducer(value, index, items),
+    });
```
cjquines, line 122: **question:** why introduce an additional closure here?

```suggestion
    let accumulator = initialValue;

    return (
      value: T,
      index?: number,
      items?: Array<T>
    ): LazyResult<Accumulator> => ({
      done: false,
      hasNext: true,
      next: indexed
        ? reducer(accumulator, value, index, items)
        : reducer(accumulator, value),
    });
```

src/mapWithFeedback.ts
```diff
@@ -0,0 +1,147 @@
...
+
+export function mapWithFeedback() {
+  return purry(
+    mapWithFeedbackImplementation(false),
+    arguments,
+    mapWithFeedback.lazy
+  );
+}
+
+const mapWithFeedbackImplementation =
```
cjquines, line 82: **note:** convention here is to use `_mapWithFeedback`

src/mapWithFeedback.ts
```diff
@@ -0,0 +1,147 @@
...
+    initialValue: Accumulator
+  ) => {
+    const implementation = indexed
+      ? mapWithFeedback.lazyIndexed
+      : mapWithFeedback.lazy;
+
+    return _reduceLazy(items, implementation(reducer, initialValue), indexed);
+  };
+
+const lazyImplementation =
```
cjquines, line 96: **note:** convention here is to use `_lazy`

src/mapWithFeedback.ts
tbui17: Thanks! I was looking at some of the more recent PRs and noted they tended to use full type names. I'll stick with the old convention.

src/mapWithFeedback.ts
tbui17: That was probably something I forgot to factor out after I made changes to other parts of the code. There is no reason for an additional closure there in this case.

src/mapWithFeedback.ts
eranhirsch: I usually prefer T for the main "actor" of the type, and full names for all the rest, but regardless we don't have strict conventions on generic variable names, as long as the type is easy to read and "feels" consistent in the codebase

src/mapWithFeedback.ts
eranhirsch: :) I actually prefer the `Implementation` naming convention and use it in my functions. the _ prefix is meaningless and could be cleaned off in the future when we have better lint rules to enforce naming conventions. To summarize, use whatever name you feel like; I prefer names having explicit semantics, so call this "Implementation" because it is... an implementation... of the function.

src/mapWithFeedback.ts
eranhirsch: Similarly

tbui17:
Thanks everyone for the reviews!

This should be ready to merge after I readjust the variable/type names.

src/mapWithFeedback.ts
```diff
@@ -0,0 +1,130 @@
+import { _reduceLazy } from "./_reduceLazy";
+import { _toLazyIndexed } from "./_toLazyIndexed";
+import type { LazyEvaluator } from "./pipe";
+import { purry } from "./purry";
+
+/**
+ * Applies a reducer function on each element of the array, accumulating the results, and returns an array of the successively reduced values.
```
eranhirsch, line 7: The term accumulator is not accurate here. The function could be used to accumulate results, but it could also be used to compute any other form of computation with memory, like sum.

Even the typescript library itself doesn't call the accumulator value in `reduce` accumulator, opting for the more accurate `previousValue`.

https://github.com/microsoft/TypeScript/blob/main/src/lib/es5.d.ts#L1463-L1470

eranhirsch (changes requested):
We should make the return type array's shape depend on the input, like `map.strict` does so that running this function on a non-empty array, for example, would return a non-empty array, and not a simple array.

Also, I think the name accumulator makes the function more confusing, especially when it isn't called reducer: previousValue would communicate much more clearly what's going on here...

src/mapWithFeedback.ts
tbui17: I agree. I think it's best to keep the parameter names consistent with the typescript team's conventions as well.
I'm opting to also change the reducer parameter name to callbackFn even though it's less descriptive, so that people can "pattern match" the two signatures and infer that the usage will be similar without learning any new API pattern.

If this sounds good, I'll introduce that into the PR as well.

tbui17: Added requested changes.

eranhirsch: I totally agree about `callbackfn`, notice that it's all lowercase!

src/mapWithFeedback.ts
```diff
@@ -0,0 +1,135 @@
...
+    );
+  }
+
+  export const lazy = lazyImplementation(false);
+  export const lazyIndexed = _toLazyIndexed(lazyImplementation(true));
+}
+
+type StrictOut<T extends IterableContainer, K> = {
+  -readonly [P in keyof T]: K;
+};
```
eranhirsch, line 135: Pull this up to the top of the file and rename it to Mapped or something, you can even pull this to the _types.ts file and re-use it in both map and here...

The name StrictOut was deliberate to make it stand out against the non-strict version. In this function we don't have a non-strict version so the name doesn't make sense. Also in the past I called all returned types "Out" because i copied these utilities from my own project, now I prefer calling the type by what it does (or more accurately, what it **is**, e.g. First, Grouped, Zipped, etc...)

src/mapWithFeedback.test.ts
```diff
@@ -0,0 +1,209 @@
...
+      const expectedEquality = Array.from({ length: 5 }).fill({
+        "1": 1,
+        "2": 2,
+        "3": 3,
+        "4": 4,
+        "5": 5,
+      });
+      expect(results).toEqual(expectedEquality);
+      const [item] = results;
+      expect(results.every((result) => result === item)).toBe(true);
```
eranhirsch, line 37: This is redundant and over-complex:
```suggestion
      const [item] = results;
      expect(item).toEqual({ "1": 1, "2": 2, "3": 3, "4": 4, "5": 5 });
      for (const result of reslts) {
        expect(result).toBe(item);
      }
```

src/mapWithFeedback.test.ts
```diff
@@ -0,0 +1,209 @@
...
+    });
+
+    it("if an empty array is provided, it should never iterate, returning an empty array.", () => {
+      const result = mapWithFeedback([], (acc) => acc, "value");
+      expect(result).toEqual([]);
```
eranhirsch, line 42: Also add a test against identity so nobody breaks this assumption
```suggestion
      const data = [];
      const result = mapWithFeedback(data, (acc) => acc, "value");
      expect(result).toEqual(data);
      expect(result).not.toBe(data);
```

src/mapWithFeedback.test.ts
```diff
@@ -0,0 +1,209 @@
...
+        const data: Array<number> = [1, 2, 3, 4, 5];
+        const result = mapWithFeedback(data, (acc, x) => acc + x, 100);
+        expectTypeOf(result).toEqualTypeOf<Array<number>>();
+      });
+    });
+  });
+
+  describe("indexed", () => {
+    it("should track index and provide entire items array", () => {
+      let i = 0;
```
eranhirsch, line 92: consider using vitest's function mocking to test these sort of things instead of building your own "state":

https://vitest.dev/api/vi.html#mocking-functions-and-objects

eranhirsch (approved):
Really nice work and great tests! See my comments inline for some fixes I think are worthwhile

src/mapWithFeedback.test.ts
tbui17: The primary reason for the redundancy is that Vitest throws on the first assertion failure, which could mask some details. I made it into one assertion to have the full error details when the test does fail.

However, I do get your point about the complexity, I wouldn't want to maintain a complex test.

tbui17:
Thanks for the feedback! Made requested changes.

eranhirsch (approved):
Looks great! thanks for adding this

PR #518, opened 2024-02-16
feat(hasSubObject): Add `hasSubObject`

### LINKED ISSUE #386: [ Feature Request ] Lodash "matches" function (Also typing "equals")
I'm looking at using a more TypeScript-oriented utility library after using Lodash. This library seems very nice. However, there does not seem to be a way to check if one object is a subset of another (?). This is pretty common functionality in Lodash like`_.find(data, {a: "b"})`. Right now, one could do `R.find(data, ({a}) => a === "b")`; however, when there are more properties it becomes more burdensome to write out the callback. I realize that you could also do something like `R.find(data, R.createPipe(R.pick(['a']), R.equals({a: "b"})))`, but this is also pretty verbose.

To be clear, I'm not looking for the "iteratees" notation, which I don't think makes sense with typings, but rather just the "matches" function: https://lodash.com/docs/4.17.15#matches. This would be similar to the R.prop function in terms of usefulness. So the example above would become `_.find(data, R.matches({a: "b"}))`.

I also took a look at the R.equals implementation to see if maybe that could be used (doesn't seem so), and I noticed the typing permitted comparing any type to any type. I can see why this makes sense, but I do think it is somewhat of a logical error to compare two things of different types and I would want the compiler to warn about it. If it is intended by the programmer, then there are ways to coerce the types to get around the type check. Maybe it would make sense to have an "R.equals.typed" function with the same implementation but a different typing?

The typing I would like for a "matches" function:
```ts
export function matches<U extends string | number | symbol, V, T extends Record<U, V>>(
  subset: T
) {
  return <S extends T>(object: S) => {
    ...
    return true
  }
}
``` 
This allows for the compiler to error when doing something like `matches({ b: 1 })({ a: 3, b: "a" })` or `matches({ c: 1 })({ a: 3, b: "a" })`.

cjquines:
closes #386 (at least the function request part)

---

Make sure that you:

---

### Available Types:

existing).

</details>


src/hasSubobject.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @signature
+ *    R.hasSubobject(object, subObject)
+ * @example
+ *    R.hasSubobject({ a: 1, b: 2, c: 3 }, { a: 1, c: 3 }) //=> true
+ *    R.hasSubobject({ a: 1, b: 2, c: 3 }, { b: 4 }) //=> false
+ *    R.hasSubobject({ a: 1, b: 2, c: 3 }, {}) //=> true
+ * @dataFirst
+ * @category Object
+ */
+export function hasSubobject<
```
eranhirsch, line 19: The O in object needs to be capitalized. `subobject` is not a common term so dictionaries like CSpell will warn about this to users.

src/hasSubobject.ts
```diff
@@ -0,0 +1,57 @@
...
+  S extends Record<PropertyKey, unknown>,
+  T extends S,
+>(object: T, subObject: S): boolean;
```
eranhirsch, line 22: 1. Just thinking out loud here looking at #565 I just sent: this function is a partial deep equality check, e.g. it's the same exact logic as isDeepEquality, with a single relaxation of the `keys.length !== Object.keys(other).length` check; we can enforce that the "subObject" overlaps with the type of the data, and have the function narrow the result instead of returning `boolean`, this will provide more type-safety when a prop name changes between the checked object and the data, and would provide a better type downstream.

2. More generally, i prefer using `object` as the type instead of `Record<PropertyKey, unknown>` as it's more broad and accepts weirder stuff too, and not just really plain simple records.

```suggestion
export function hasSubobject<
T /* extends object? --- we might need it to make the typing work for the impl? */,
S extends T = T,
>(object: T, subObject: S): object is T & S;
```

eranhirsch (changes requested):
I think we should rename it hasSubObject to prevent CSpell from flagging it.

Also, maybe we can convert this to a proper type-guard, it will be more useful.

src/hasSubobject.ts
cjquines: i assume you mean `S extends object` and `T extends S = S`

i'm not sure i see the point to have the type guard... `object` has to have type `T extends S` anyway, so it must be `T & S` right?

eranhirsch: Maybe I don't understand the type correctly, this is what I'm thinking:
```ts
type A = {
  a?: number;
  b?: string;
};

declare a: A = {};
if (hasSubObject(a, { b: "hello" })) {
  console.log(a);
  //          ^?: { a?: number, b: string }
}

if (hasSubObject(a, { b: "hello" } as const)) {
  console.log(a);
  //          ^?: { a?: number, b: "hello" }
}

// @ts-expect-error
hasSubObject(a, { c: true });
```

src/hasSubObject.ts
```diff
@@ -0,0 +1,58 @@
...
+ *    R.hasSubObject({ a: 1, b: 2, c: 3 }, { a: 1, c: 3 }) //=> true
+ *    R.hasSubObject({ a: 1, b: 2, c: 3 }, { b: 4 }) //=> false
+ *    R.hasSubObject({ a: 1, b: 2, c: 3 }, {}) //=> true
+ * @dataFirst
+ * @category Object
+ */
+export function hasSubObject<S extends object, T extends S = S>(
+  object: T,
+  subObject: S,
+): object is S & T;
```
eranhirsch, line 22: The typing here is wrong; it won't be able to narrow T properly, and the "roles" should be reversed. I couldn't get the type to work without using Simplify, probably because typescript doesn't try to "flatten" the object, but it works with it. Finally, just a nit of mine, but avoid using `object` as a name, as it's also a type, which could lead to confusion and even bugs.

```suggestion
export function hasSubObject<T, S extends Partial<T>>(
  data: T,
  subObject: S,
): data is Simplify<S & T>;
```

This is the test I used to play around with the type:
```ts
  test("narrowing", () => {
    const obj = {} as { a?: string; b?: number };

    if (hasSubObject(obj, { a: "a" })) {
      expectTypeOf(obj).toEqualTypeOf<{ a: string; b?: number }>();
    }

    if (hasSubObject(obj, { a: "a" } as const)) {
      expectTypeOf(obj).toEqualTypeOf<{ a: "a"; b?: number }>();
    }

    // @ts-expect-error [ts2353]
    hasSubObject(obj, { c: true });
  });
```

eranhirsch (changes requested):

src/hasSubObject.ts
cjquines: if we use this type, then how would we type the data-last version? 
```ts
export function hasSubObject<T, S extends Partial<T>>(
  subObject: S,
): (data: T) => data is Simplify<S & T>;

// no error:
hasSubObject({ a: 1 })({ b: 2 });
```

should the data-last type remain the same?

also, i think the partial here should be a deep partial, if we want

```ts
hasSubObject({ a: { b: 1, c: 2 } }, { a: { b: 1 } });
```

to be ok

src/hasSubObject.ts
```diff
@@ -0,0 +1,61 @@
...
+  return purry(_hasSubObject, arguments);
+}
+
+function _hasSubObject<T, S extends DeepPartial<T>>(
+  data: T,
+  subObject: S,
+): data is S & T {
+  for (const key of Object.keys(subObject)) {
+    // @ts-expect-error [ts7053] - This indexing is safe:
+    if (!equals(subObject[key], data[key])) {
```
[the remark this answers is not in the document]
cjquines: oh, good point, and i don't think we want to include that; let me change it

src/hasSubObject.test.ts
```diff
@@ -0,0 +1,158 @@
...
+        expectTypeOf(obj).toEqualTypeOf<{ a: string; b?: number }>();
+      }
+
+      if (hasSubObject(obj, { a: "a" } as const)) {
+        expectTypeOf(obj).toEqualTypeOf<{ a: "a"; b?: number }>();
+      }
+    });
+  });
+
+  describe("data-last", () => {
```
cjquines, line 110: @eranhirsch not sure how i feel about these types; i'm kinda leaning on making them both `unknown` or `extends object`?

src/hasSubObject.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @example
+ *    R.hasSubObject({ a: 1, c: 3 })({ a: 1, b: 2, c: 3 }) //=> true
+ *    R.hasSubObject({ b: 4 })({ a: 1, b: 2, c: 3 }) //=> false
+ *    R.hasSubObject({})({ a: 1, b: 2, c: 3 }) //=> true
+ * @dataLast
+ * @category Object
+ */
+export function hasSubObject<S>(
+  subObject: S,
+): <T extends S>(data: T) => data is Simplify<S & T>;
```
eranhirsch, line 44: Why aren't we using the same typing for the dataLast impl? wouldn't the semantics be different now that T needs to extend S whereas in the dataFirst impl the S extends T?

src/hasSubObject.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export function hasSubObject(): unknown {
+  return purry(_hasSubObject, arguments);
+}
+
+function _hasSubObject<T, S extends Partial<T>>(
+  data: T,
+  subObject: S,
+): data is S & T {
```
eranhirsch, line 53: Should we use the same signature here as the dataFirst, adding the Simplify utility?
```suggestion
): data is Simplify<S & T> {
```

src/hasSubObject.ts
```diff
@@ -0,0 +1,61 @@
...
+  for (const key of Object.keys(subObject)) {
+    // @ts-expect-error [ts7053] - This indexing is safe:
```
eranhirsch, line 55: But don't we need to check that the key is also a key of `data`? Assuming we use the typescript option that distinguishes between optional and undefined, the following would come up true, but it shouldn't.

```ts
hasSubObject({} as { a?: string | undefined }, { a: undefined })
```

src/hasSubObject.test.ts
eranhirsch: I don't understand why the dataLast impl has different types

eranhirsch, line 110: Generally speaking about testing dataLast, I prefer testing everything via a pipe because the inference works a little differently, and it represents real-life usage a little better. Can you add a pipe to all the dataLast invocations?

src/hasSubObject.ts
cjquines: oops, for some reason i thought it needed manual typing? but i tried and it actually worked

src/hasSubObject.ts
cjquines: good point, added check for `hasOwnProperty`

src/hasSubObject.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+    // Error has no enumerable properties, so this is true:
+    expect(hasSubObject(new Error("a"), new Error("b"))).toBe(true);
+
+    expect(hasSubObject(["a", "b"], ["a"])).toBe(true);
+    expect(hasSubObject(["a"], ["a", "b"])).toBe(false);
+
+    expect(hasSubObject("ab", "a")).toBe(true);
+    expect(hasSubObject("a", "ab")).toBe(false);
+  });
```
cjquines, line 42: **question:** should we have these as tests?

i think the Error tests are reasonable, less so the array and string ones

notably, there are no type errors with calling something like `hasSubObject(1, 2)` (which returns `true`)

eranhirsch (changes requested):
* I just shipped `isDeepEqual`; we are deprecating `equals,` so we'll need to use the former instead (and refer to it in the docs too).
* I recently encountered issues with the `S & T` type (see my PR about `merge`). Should we use the type-fest `Merge` type for the result so we don't err on types that include index accessors (`{ [key: string]: T }`)?

Sorry for the repeated request changes. It's just complex to get right, and I recently worked on related stuff, so I notice these edge cases more prominently right now.

cjquines:
not sure if we can use typefest merge; it looks like it errors: [playground link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgWQKZQOargXzgMyghDgCIYBPMVAWn1QGcZSBuAKDYBNUBjAGwCGUbPgCuAOx4xgEcXAAWAhgGVRAIwDyagFa8YAHgAqAGjjK4qAB4xU4zgzgAFIdIF8jAPg8AKNnDicAjACAFxwJn5wDOpaulJhysZsAJRhgcFwwA5omKj6ieEeLEA)

eranhirsch:
I'm inclined to say that we can suppress the error message; typescript can't infer that `Merge` is similar enough to `S & T`; it's expected that the type doesn't match because `Merge` is a sort of "fix" for `S & T` _because_ typescript doesn't infer merges cleanly. The reason it worked with `merge` is that we just return a merge, and don't need to narrow anything.

src/hasSubObject.ts
```diff
@@ -0,0 +1,68 @@
...
+ *    R.hasSubObject({ a: 1, b: 2, c: 3 }, { a: 1, c: 3 }) //=> true
+ *    R.hasSubObject({ a: 1, b: 2, c: 3 }, { b: 4 }) //=> false
+ *    R.hasSubObject({ a: 1, b: 2, c: 3 }, {}) //=> true
+ * @dataFirst
+ * @category Object
+ */
+export function hasSubObject<T, S extends Partial<T>>(
+  data: T,
+  subObject: S,
+  // @ts-expect-error: typescript doesn't infer merges correctly
```
eranhirsch, line 24: Can you add the typescript error too? that way when a change in typescript causes the error to not be relevant, or when someone tries to fix the ts error they know what was suppressed here. (I wish typescript would support saying what error is expected and not this generic disabler, like `eslint-disable` supports)

```ts
// @ts-expect-error [tsXXXX]: ...
```

src/hasSubObject.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param data - The object to test.
+ * @param subObject - The sub-object to test against.
+ * @signature
+ *    R.hasSubObject(data, subObject)
+ * @example
+ *    R.hasSubObject({ a: 1, b: 2, c: 3 }, { a: 1, c: 3 }) //=> true
+ *    R.hasSubObject({ a: 1, b: 2, c: 3 }, { b: 4 }) //=> false
+ *    R.hasSubObject({ a: 1, b: 2, c: 3 }, {}) //=> true
+ * @dataFirst
+ * @category Object
```
eranhirsch, line 19: Category is Guard
```suggestion
 * @category Guard
```

src/hasSubObject.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param subObject - The sub-object to test against.
+ * @param object - The object to test.
+ * @signature
+ *    R.hasSubObject(subObject)(data)
+ * @example
+ *    R.hasSubObject({ a: 1, c: 3 })({ a: 1, b: 2, c: 3 }) //=> true
+ *    R.hasSubObject({ b: 4 })({ a: 1, b: 2, c: 3 }) //=> false
+ *    R.hasSubObject({})({ a: 1, b: 2, c: 3 }) //=> true
+ * @dataLast
+ * @category Object
```
eranhirsch, line 41:
```suggestion
 * @category Guard
```

src/hasSubObject.test.ts
eranhirsch: I don't think we should officially test anything except proper objects. I don't mind the type being loose enough to allow people to use it with weird edge-cases, but by testing it it means we support these cases, and can't break them in the future.

eranhirsch (approved):
This looks done! just a few more nits and i'll ship it

eranhirsch:
Because we strip comments on our ts builds the suppression comment is disabled and we get issues on our output types. 

For now, let's use `S & T` but replace the suppression comment with a TODO because we might be able to use it once we move to building with tsup.

eranhirsch (approved):
Finally! Let's ship this!

PR #520, opened 2024-02-16
feat(iterPipe): Add `iterPipe`

### LINKED ISSUE #74: Possiblity to directly use iterators ?
I would want to do something like this:
```ts
const myMap = new Map<string, T>(...);
const iterator: IterableIterator<T> = myMap.values();

const first = R.pipe(
  iterator,
  R.first
);
```

Problem is `R.first` accepts a `T[]` and not an `IterableIterator<T>`, so my `first` var is of type `unknown` instead of being of type `T`;

cjquines:
kinda closes #74 

TODO:
- [ ] add the other overloads
- [ ] more tests

NOTE: after v2, we won't have .lazy exposed any more, but we can still have a "best effort iterpipe", where functions are called lazily if possible, and otherwise called normally and cast to lazy format

---

Make sure that you:

---

### Available Types:

existing).

</details>


src/iterPipe.ts
```diff
@@ -0,0 +1,70 @@
...
+ *      )
+ *    ).join("; ") // => "1 -> 2; 2 -> 4; 3 -> 6; 4 -> 8"
+ *
+ *
+ * @dataFirst
+ * @category Function
+ */
+export function iterPipe<A, B>(
+  values: IterableIterator<A>,
+  op1: (value: A) => LazyResult<B>
```
cjquines, line 50: **question:** we can also do

```ts
op1: (value: A) => LazyResult<B> | IterableIterator<B>
```

so we can write the example, where we include a `function *`. i think it should be possible to tell this apart on runtime, it has `hasNext` if it's a LazyResult, otherwise it's an IterableIterator

tsconfig.json
```diff
@@ -22,6 +22,6 @@
     // the vast majority of browsers are probably above that too already. It's
     // most likely safe to up this value, and it would probably result in a
     // smaller package size.
-    "target": "ES5",
+    "target": "ES6",
```
cjquines, line 25: **note:** we can make this ES5 but it would be a pain to rewrite the generators. however, this would be a breaking change

eranhirsch:
Why did you close the PR @cjquines ?

cjquines:
@eranhirsch i couldn't find the convert to draft button oops. i'll reopen as a draft with some more fixes

eranhirsch:
Should be on the right-side panel, somewhere near the reviewers box

eranhirsch:

Issue #509, opened 2024-02-12
Add string case utils: `toLowerCase`, `capitalize`, etc


lensbart:
Hi,

Thanks for this library, it works well. I recently moved away from using Ramda because I didn’t like that the keys used by `omit` weren’t narrowly typed.

As part of that migration, I had to move a couple of functions that are not implemented by Remeda into my own repository. I think these could be well-suited for inclusion:

```ts
/**
 * similar to https://ramdajs.com/docs/#has, but uncurried and with the
 * opposite order of arguments
 */
export const has = <T extends number | string | symbol>(
  obj: unknown,
  key: T,
): obj is Record<T, unknown> =>
  typeof obj === 'object' && obj != null && key in obj

/** see https://ramdajs.com/docs/#sum */
export const sum = (list: number[]) => list.reduce((a, b) => a + b, 0)

/** see https://ramdajs.com/docs/#toLower */
const toLower = <T extends string>(string: T) =>
  string.toLowerCase() as Lowercase<T>
```

In addition, I find the following utility function useful, because their return value is typed as a string literal, if called with a string literal argument:

```ts
export const capitalize = <T extends string>(key: T) =>
  `${key[0]?.toUpperCase() ?? ''}${key.slice(1)}` as Capitalize<T>

export const uncapitalize = <T extends string>(key: T) =>
  `${key[0]?.toLowerCase() ?? ''}${key.slice(1)}` as Uncapitalize<T>
```

Finally, `unreachable` can be a handy function to enforce exhaustiveness in a `switch`/`case`, either when you don’t use `typescript-eslint` to do this check for you, or if you want to have absolute certainty that there will be no fallthrough (e.g. in case of a stale frontend receiving data from a freshly updated backend). You can use this function in the `default` case, to ensure via the type system that you handled all cases in the `case`s above:

```ts
/** Workaround for https://github.com/typescript-eslint/typescript-eslint/issues/2841 */
export const unreachable = (never: never) => {
  throw new Error('Unreachable', { cause: never })
}
```

eranhirsch:
* `has` was suggested in #461 as `hasProp` and would be added soon as part of that issue
* `sum` can be implemented easily: `R.sumBy(R.identity)`, there's not much value in including it as a standalone function.
* `toLowerCase`, `toUpperCase`, `capitalize`, and `uncapitalize` are all great additions, especially as a dataLast function that could be used in pipes. Send a PR, and I'll approve them.
[...]

PR #525, opened 2024-02-18
feat(capitalize,toLowerCase,toUpperCase,uncapitalize): add `capitalize`, `toLowerCase`, `toUpperCase` and `uncapitalize`

lensbart:
Fixes https://github.com/remeda/remeda/issues/509


lensbart:
@eranhirsch any chance that this can get a review? This PR has conflicts now, due to other PR’s that have been opened and merged in the meantime.

eranhirsch:
Hey, I'm sorry for the delays. The other PRs being merged are mine, and I'm sorry for the extra work they are incurring. I'm getting them done as quickly as possible to minimize these conflicts. The gist is that I'm reinforcing typescript and eslint (and prettier) in the project so that PRs get better feedback quicker, requiring fewer cycles and allowing me to review more PRs faster. Unfortunately, this delays reviewing the current PRs.

By the end of this week, those changes should be done, and then a simple merge from master should bring your environment up to speed You can follow the lint fixes and suggestions if any issues are raised. If you haven't already, you can always run `npx tsc`, `npm run lint` and `npx prettier . --write` to make sure you are getting the right signals from the CI.

eranhirsch (changes requested):
Hey, I merged the index file for you, can you pull this PR locally, run:
`npx prettier . --write` to bring styling up-to-date, and run `npm run lint` and fix any issues (I think it's just adding return types).

Also, please add dataLast implementations for all functions so they could be used in mappers and pipes, e.g.
```ts
pipe(
  stringData,
  map(toLowerCase()),
  ...
);
```

lensbart:
I rebased, re-formatted and added return types.

> Also, please add dataLast implementations for all functions so they could be used in mappers and pipes, e.g.
> 
> ```ts
> pipe(
>   stringData,
>   map(toLowerCase()),
>   ...
> );
> ```

Does this make sense? These functions all take just 1 argument, so they could safely be used in mappers and pipes like so:

```ts
pipe(
  stringData,
  map(toLowerCase),
  ...
);
```

eranhirsch:
We're moving away from 'headless' dataLast invocations for functions that don't take any parameters. They don't *always* work (like `first`), and that creates a confusing and inconsistent API where people might either add parenthesis to a headless function or not add parenthesis to a non-headless one and not get any meaningful errors to fix the issue.

See: https://github.com/remeda/remeda/pull/531

Unfortunately, we can't switch to a consistent format without breaking changes, so we are waiting to release Remeda V2 for this.

src/capitalize.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * A function equivalent to TypeScript’s built-in `Capitalize` utility type.
```
eranhirsch, line 2: We don't need to mention "a function" because everything in Remeda is a function. Assuming people don't know Typescript's Capitalize util, it would be easier for people to get the full description for this function right here...

Suggestion (via chatgpt):
> Converts the first character of a given string to uppercase while maintaining the rest of the string as is.

src/capitalize.ts
```diff
@@ -0,0 +1,14 @@
...
+ * @param string The string to capitalize.
+ * @signature
+ *    R.capitalize(string)
+ * @example
+ *    R.capitalize(`hello`) // => `Hello`
+ * @dataLast
+ * @category String
+ * @strict
+ */
+export const capitalize = <T extends string>(string: T): Capitalize<T> =>
```
eranhirsch, line 13: To prevent any accidental confusion, let's avoid calling the variable `string` as it's also the type name. 
```suggestion
export const capitalize = <T extends string>(str: T): Capitalize<T> =>
```

src/toLowerCase.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  expect(
+    toLowerCase(`
+😎
+/|\__👍
+/'\ `),
+  ).toEqual(`
+😎
+/|\__👍
+/'\ `);
+});
```
eranhirsch, line 30: 😎

src/toLowerCase.test.ts
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable no-useless-escape */
```
eranhirsch, line 1: put this above the line that needs it. If you need it for a block prefer wrapping that area with:

```ts
/* eslint-disable no-useless-escape */
...
...
...
/* eslint-enable no-useless-escape */
```

eranhirsch (changes requested):
Let's add native dataLast implementations

Brian-Pob:
Apologies for my ignorance, but what is the purpose of implementing functions like `toLoweCase` and `toUpperCase` when they already exist in native JavaScript? The only difference I can see is how you call them:

```js
R.toLowerCase("MyString");

"MyString".toLowerCase();

// alternatively
const mystr = "MyString";
mystr.toLowerCase();
```

Thanks.

lensbart:
@Brian-Pob

If some `variable` is typed as a string literal, the result of `toLowerCase(variable)` will be typed as the corresponding string literal, whereas `variable.toLowerCase()` will be typed as `string`.

Brian-Pob:
> @Brian-Pob
> 
> If some `variable` is typed as a string literal, the result of `toLowerCase(variable)` will be typed as the corresponding string literal, whereas `variable.toLowerCase()` will be typed as `string`.

I see. I get it now after reading your explanation and looking at the implementation. Thanks.

eranhirsch:
Also you get the added value of being able to use the dataLast version in pipes and mappers so you can write:
```ts
R.map(data, R.toLowerCase());
```

instead of:
```ts
data.map((tempName) => tempName.toLowerCase());
```

which itself becomes more useful once you start using pipes:
```ts
const result = R.pipe(
  someArrayOfNonStringData,
  R.filter(...),
  R.map(...a mapper that returns a string...),
  R.take(5),
  R.join("_"),
  R.toLowerCase(),
);
```

In the example above you will only run the filter and mapper on the first 5 objects that fulfil the mapper, you will only join 5 items, and then run lowercase just once on the output object.

Brodzko:
I'm also not sure what's the utility of `uncapitalize`. Are there many use-cases where you only want the first letter to lower-case?

lensbart:
@Brodzko `uncapitalize` can be used to convert PascalCase to camelCase. IMO the utility of `uncapitalize` is like the utility of [`Uncapitalize`](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html#uncapitalizestringtype). Since this utility type is part of the TypeScript API, in my view it makes sense to include the corresponding function in a TypeScript-specific utility library.

cjquines:
hey @lensbart — now that we released v2, we should point this PR to `main` instead of `master`. do you still intend to work on this?

cenobitedk:
Hi @lensbart would love to get these features in remeda and ditch lodash. Do you have time to update the PR?

lensbart:
Apologies for the late response!

I might have time in a couple of weeks when work slows down, but (after having fixed a couple of rounds of feedback here) I just copied the code to my repo to be done with it, so it’s not a high priority for me unfortunately.

Feel free to push to this branch (or inline the code yourself) if you want something sooner. Thanks

eranhirsch:

PR #531, opened 2024-02-23
feat: Add native dataLast implementations

eranhirsch:
It's confusing that some functions have native dataLast impls because of typing, like `first` but some don't. It create weird pipes that don't look consistent.

It also means that our pipe function needs to handle more cases, specifically for lazy invocations where the lazyArgs isn't initialized because it doesn't work via the `purry` function.

### DIFF
--- a/src/debounce.test.ts
+++ b/src/debounce.test.ts
@@ -3,7 +3,7 @@ import { identity } from './identity';
 
 describe('Main functionality', () => {
   it('should debounce a function', async () => {
-    const mockFn = vi.fn(identity());
+    const mockFn = vi.fn(identity);
 
     const debouncer = debounce(mockFn, { waitMs: 32 });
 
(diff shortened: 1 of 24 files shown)


cjquines:
i'm confused about which functions this should cover... should this also cover `clone`, `flatten`, `flattenDeep`, `last`, `mergeAll`, `uniq`? should it cover the type guards? should we edit the docs for functions like `only`, `shuffle`, `sort` to have a pipe example for data last?

eranhirsch:
It should cover all 1-param functions. Thanks for surfacing others that need to be done.

TkDodo (commented):
is this a breaking change? will the old syntax that was in the docs before still work?

eranhirsch:
This is not a breaking change and the old syntax should still work, there's no way to block it from working, but the old syntax is not well tested (beyond a single test in `first`)

eranhirsch:
* Added tests for the "headless" versions
* Went over all functions mentioned here and made sure they all have proper docs.
* Skipping `clone` and `mergeAll` for now, i'm not especially concerned about them and they require a little more care to make sure we don't break anything, specifically mergeAll typing for a dataLast impl might be hard and non-trivial.

PR #532, opened 2024-02-25
docs: add ability to filter by category in sidebar

lesha1201:
This PR adds ability to filter in the sidebar (navbar) by a category like on Lodash docs or Ramda docs.

https://github.com/remeda/remeda/assets/10157660/7c7b8f00-2ba1-4910-b258-52c9ad296632

### DIFF
--- a/docs/src/components/navbar.tsx
+++ b/docs/src/components/navbar.tsx
@@ -31,9 +31,11 @@ export function Navbar({
         ([category, entries]) =>
           [
             category,
-            entries.filter(({ name }) =>
-              name.toLowerCase().includes(lowerCaseQuery),
-            ),
+            category.toLowerCase().includes(lowerCaseQuery)
+              ? entries
+              : entries.filter(({ name }) =>
+                  name.toLowerCase().includes(lowerCaseQuery),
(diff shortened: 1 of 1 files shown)


docs/src/components/navbar.tsx
```diff
@@ -31,9 +31,11 @@ export function Navbar({
         ([category, entries]) =>
           [
             category,
-            entries.filter(({ name }) =>
-              name.toLowerCase().includes(lowerCaseQuery),
-            ),
+            category.toLowerCase().includes(lowerCaseQuery)
```
lesha1201, line 34: Not sure but `startsWith` might be better for this case.

eranhirsch: I agree with this observation, I think startsWith will make more sense here.

eranhirsch (changes requested):
Let's use startsWith

docs/src/components/navbar.tsx
lesha1201: Done

eranhirsch (approved):
Thanks for helping us!

PR #538, opened 2024-02-28
fix: opinionated lint rules (auto-fixers)

eranhirsch:
The more eslint rules we have turned on, the easier it is for contributors to get feedback fast and get PRs reviewed and shipped.

This PR adds a bunch of lint rules from both typescript and eslint which aim to prevent JSisms, and standardize the codebase.


TkDodo (approved):
love it ❤️

PR #558, opened 2024-03-06
fix: add jsdoc lint

### LINKED ISSUE #556: Add linting to JSDoc
there's an [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc/)

correctness:
* reports badly formatted jsdoc blocks
* reports bad jsdoc tags (like `@params` in `dropFirstBy` and `takeFirstBy`, which should be `@param`)
* checks that params match. jsdoc does not process params that don't exist. mostly useful for the data-last functions we have, some of which document the data param. you can see this [in the generated docs](https://remedajs.com/docs/#add): note that the data last signatures don't have params for the callback arg, even though they're in the jsdoc
* can enforce that one of `@dataFirst` or `@dataLast` always exists (for example, `flattenDeep` does not have `@dataFirst`....)
    * right now we have functions like `clone` which are neither. i think these should be `@dataFirst`. but we can also introduce a `@dataNeither` option, or put an ignore...

style:
* can enforce that jsdoc exists for exported functions
* can enforce that all params are documented
* can enforce that an example exists
* can enforce that returns are documented

consistency:
* we mix `@return` and `@returns`
* we have inconsistent lines between tags
* we have inconsistent order of tags
* we sometimes put hyphens in `@param` and `@returns`, and sometimes don't

cjquines:
closes #556. someone can make a follow pr to turn on the TODOs here (or i can too)


cjquines:
i think the type errors here also exist on master...???

eslint.config.mjs
```diff
@@ -157,6 +160,93 @@ export default config(
...
+      "jsdoc/no-multi-asterisks": ["warn", { allowWhitespace: true }],
+      "jsdoc/require-description-complete-sentence": [
+        "warn",
+        { abbreviations: ["etc.", "e.g.", "i.e."] },
+      ],
+      "jsdoc/require-hyphen-before-param-description": [
+        "error",
+        "always",
+        { tags: { "*": "never" } },
+      ],
```
cjquines, line 225: these two are the culprits behind most of the changes in this pr. we can turn them off

eslint.config.mjs
```diff
@@ -157,6 +160,93 @@ export default config(
       "no-useless-concat": "warn",
       "no-useless-rename": "error",
 
+      // === JSDoc =============================================================
+      // (We are assuming that the config is extended by JSDoc's: recommended
```
eranhirsch, line 164: we are extending the 'recommended-typescript' extension...

eslint.config.mjs
eranhirsch: I think this is fine... I'm assuming this PR only contains auto-fixers and no manual fixes, right?

eranhirsch (changes requested):
Are all changes to function files due to auto-fixers or are there manual fixes in this PR too?

I think we should turn off whatever rule is removing blank lines from the blocks - It makes the comments easier to maintain and read when there's an empty line between the description and the rest of the JSDoc, if something like this could be enforced by the linter, the better.

Also, I've noticed a few rules have exemptions for deprecated functions, I don't think that's necessary, as all deprecated functions would be ones which were previously part of the project, so they will be up to the lint rules once we ship this PR. Similarly for exampleRaw which is only used in `purry` and should just be removed (maybe i'll send a PR for that).

I saw there's a rule called `no-missing-syntax`, could we use it to force that all docs are marked either as dataFirst or dataLast?

Finally, i have some questions about optional rules that I see you didn't chose, i'm just curious to get your opinion on each one:

* check-examples
* check-indentation
* check-line-alignment
* informative-docs
* no-blank-block-descriptions
* require-throws
* require-yields

eslint.config.mjs
cjquines: yep!

eranhirsch (approved):
Looks great, you didn't address two of my questions from the previous review:

> I saw there's a rule called no-missing-syntax, could we use it to force that all docs are marked either as dataFirst or dataLast?

> [...] about optional rules that I see you didn't choose to include, I'm just curious to get your opinion on each one:
check-examples
check-indentation
check-line-alignment
informative-docs
no-blank-block-descriptions
require-throws
require-yields

eranhirsch:
My bad, this shouldn't have been named fix, as it doesn't change any typing or runtime behavior :(

PR #567, opened 2024-03-10
feat!: deprecate `indexed` variant

eranhirsch:
The `indexed` variant isn't required by the runtime as Typescript considers functions with partial signatures (those that take less params then the expected type) as extending there full counterpart. This can be seen througout the typescript library since ES5 where:
`Array.prototype.map((item) => ...)` is just as valid as `Array.prototype.map(item, index, array) => ...)` without the function needing any overloads to do so.

The current implementation of pipe (and reduce lazy) had explicit checks for `indexed` just for the sake of calling the function with less params, but that was just preventing this optimization.

In this PR I only changed the implementation of `map` from the user's perspective, although all functions are now different.

We can't remove `indexed` altogether because that would be a major breaking change and would need to be done as part of Remeda v2.

I want to release this change as a canary to see if anyone reports any breakages. After a few days we can followup and deprecate all indexed variants.

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/map.ts
+++ b/src/map.ts
@@ -1,11 +1,5 @@
 import { _reduceLazy } from "./_reduceLazy";
-import { _toLazyIndexed } from "./_toLazyIndexed";
-import type {
-  IterableContainer,
-  Pred,
-  PredIndexed,
-  PredIndexedOptional,
-} from "./_types";
+import type { IterableContainer, PredIndexed } from "./_types";
 import type { LazyEvaluator } from "./pipe";
 import { purry } from "./purry";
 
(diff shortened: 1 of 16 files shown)


TkDodo:
I think the idea of indexed was that you can call something like `.map(parseInt)` and not be surprised by the second argument

eranhirsch:
It's a whole lot of infra to protect the user against a langue feature that isn't related to the library, with typescript preventing some cases and eslint capable of protecting other cases I don't think there's room or need for it.

We can even keep the signature but drop the runtime considerations.

I think that if the ecmascript standard functions are built like this then it's more confusing to diverge from the practice.

eranhirsch:
Also, it doesn't exist in lodash or ramda.

eranhirsch:
But I now understand that we'll be breaking people's code if they already have `map(parseInt)` in their code, so we can postpone this to the V2 release instead of shipping it as part of v1.

src/map.ts
```diff
@@ -98,19 +97,24 @@ type StrictOut<T extends IterableContainer, K> = {
...
   ): (array: ReadonlyArray<T>) => Array<K>;
+
   export function indexed(): unknown {
-    return purry(_map(true), arguments, map.lazyIndexed);
+    return purry(mapImplementation, arguments, _lazy);
   }
 
-  export const lazy = _lazy(false);
-  export const lazyIndexed = _toLazyIndexed(_lazy(true));
-
```
cjquines, line 114: `lazy` is used for #520 , but in a purported v2 we can rewrite to properly use iterators

TkDodo:
yes we can do this for v2

eranhirsch:
Will re-do this on top of the v2 branch

PR #368, opened 2023-08-28
feat(multisets): add difference, union, intersect, isContains, symmetric difference (xor), and isEqual

eranhirsch:
Adding multi-set functions:
- [x] Difference
- [x] Union
- [x] Intersect
- [x] IsContains
- [x] Symmetric Difference ("XOR")
- [x] isEqual 
- ~~isDisjoint~~ (this doesn't have any multiset specific impl concerns)

Closes #363

---

Make sure that you:

- [x] Typedoc added for new methods and updated for changed
- [x] Tests added for new methods and updated for changed
- [x] New methods added to `src/index.ts`

PR #568, opened 2024-03-11
feat: isIncludedIn

eranhirsch:
A type-guard that checks existence in a set of values. Useful for narrowing possible enum (literal union) values via a global const.

Optimizes dataFirst and dataLast implementations offering better performance for each case.

Deprecated intersection and difference in favour of using filter in combination with isIncludedIn so we can take advantage of the performance gains and optimizations, and to make room for future interestion and difference functions which will work more closely to the mathematical definitions of multi-sets (which take into account duplicate items, as could be present in arrays). See #368 

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -67,6 +67,7 @@ export default config(
       "unicorn/prefer-at": "off",
       "unicorn/prefer-number-properties": "off",
       "unicorn/prefer-spread": "off",
+      "unicorn/prefer-includes": "off",
 
       // TODO: These rules allow us to really standardize our codebase, but they
       // also do sweeping changes to the whole codebase which is very noisy. We
--- a/src/difference.ts
+++ b/src/difference.ts
@@ -6,8 +6,7 @@ import { purry } from "./purry";
  * Excludes the values from `other` array.
  *
  * **DEPRECATED: equivalent to `R.filter(array, R.isNot(R.isIncludedIn(other)))`
- * and will be removed in v2.**
- *
+ * and will be removed in v2.**.
  *
  * @param array - The source array.
  * @param other - The values to exclude.
@@ -18,8 +17,7 @@ import { purry } from "./purry";
  * @dataFirst
  * @pipeable
  * @category Array
- * @pipeable
- * @deprecated equivalent to `R.filter(array, R.isNot(R.isIncludedIn(other)))` and will be removed in v2.
+ * @deprecated Equivalent to `R.filter(array, R.isNot(R.isIncludedIn(other)))` and will be removed in v2.
  */
 export function difference<T>(
   array: ReadonlyArray<T>,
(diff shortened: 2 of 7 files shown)


cjquines:
hm, why deprecate? `R.filter(array, R.isNot(R.isIncludedIn(other)))` vs `R.difference(array, other)` are quite different reading-wise

i suppose i'm still confused with the philosophy of which functions are included; we have `R.add` and `R.prop` which are ts one-liners (`(x, y) => x + y` and `({ [prop]: value }) => value`), but we don't have things like `R.sum`

eranhirsch:
1. For a while, we opposed having stuff like `add` and `subtract` for exactly the reasons you mentioned. The mapping file still has a section for "simple one-liners," but from my experience in real-life projects, the code ends up being less readable requiring (`map((item) => item + 3)` vs `map(add(3))`, where we don't need to come up with a name for `item` just to add 3 to it). Without a utility, less experienced engineers might think that what they are doing is wrong or discouraged by the system ("if it's not built into the project, it might be incorrect to use it").

2. Conversely, I prefer we provide good building blocks that are **very reusable** over packaging several semantic "tools" into a single "solution". This is more apparent with `reject`, for example; we don't need to provide both a `filter` function and a "filter but reversed" function when we can provide a utility to flip a predicate (`isNot`). This allows us to optimize the implementation and typing of `filter` without needing to backport these optimizations to each sub-case separately; And lets people use `isNot` for *other solutions* without needing to re-package those tools in a new utility "solution".

3. Finally, specifically for `difference` and `intersects`, I feel they have incorrect semantics: they operate on arrays and not sets but act like a hybrid of a set and a multi-set: `difference([1,1,1,1], [1]); // []` but `difference([1,1,2,2], [1]) // [2, 2]`. Forcing people to use `filter` to get the same results makes it *clearer* what these functions do (`filter([1,1,1,1], isNot(isIncludedIn([1])))`). For the person writing the code that uses `intersection`, the semantics might be clear as they read the docs, but for someone reading the code, the fact that the semantics aren't clear is actually confusing and might lead to bad assumptions. My plan is to add new implementations for these functions that have the correct multi-set semantics; having those in addition to the `filter` "recipe" provided here would allow users to pick which logic they need for their use-case. See #368

Bessonov:
I find it very disappointing that `difference` and `intersection` are deprecated. They are very convenient to use and MUCH clearer and cleaner than `filter(array, isIncludedIn(other)`. Maybe they could be reintroduced as shortcuts?

eranhirsch:
After we release v2 of Remeda they will be introduced with a fixed implementation. Because the new implementation would have different semantics, we had to deprecate the existing functions so people would stop using them and the names would be "free" for a new implementation. We didn't want to change the implementation for people without them knowing of the change because it had the potential of breaking people's code in production in ways that would be hard to detect otherwise.

PR #570, opened 2024-03-13
feat(v2): bumping typescript target and library to es2022

eranhirsch:
Just doing this change and following up on any errors and warnings

### DIFF
--- a/tsconfig.dist.json
+++ b/tsconfig.dist.json
@@ -8,9 +8,6 @@
 
     "noEmit": true,
 
-    "target": "ES5",
-    "lib": ["ES6", "ES2017.Object", "DOM", "DOM.Iterable"],
-
     "skipDefaultLibCheck": false,
     "skipLibCheck": false
   }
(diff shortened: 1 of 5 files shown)


tsconfig.dist.json
```diff
@@ -8,9 +8,6 @@
 
     "noEmit": true,
 
-    "target": "ES5",
-    "lib": ["ES6", "ES2017.Object", "DOM", "DOM.Iterable"],
-
```
cjquines, line 13: **question:** why remove this? my understanding is that we use this to typecheck the built output in ci, and we need the lib definitions to do so

eranhirsch: Looking at the CI there are issues with this, I might revert it. It seemed redundant when I looked at it but I might have just gotten something wrong... I thought this file inherits/extends from the main one and these are just duplications.

cjquines: oh, no, it's an entirely separate ts file...

PR #571, opened 2024-03-15
feat(v2): deprecate and rename functions ahead of v2 release.

eranhirsch:
I am marking functions as deprecated in v1 ahead of removal in v2. This includes functions that were sunset and functions that were just renamed. For renamed toPairs and fromPairs this also includes adding the dataLast impl back from pr #531 as the rename allows people to migrate cleanly.

Notice that this PR is for v1, not v2!


src/createPipe.ts
```diff
@@ -3,14 +3,17 @@ import { pipe } from "./pipe";
 /**
  * Creates a data-last pipe function. First function must be always annotated. Other functions are automatically inferred.
  *
+ * ! **DEPRECATED**: Use `R.piped(op1, op2, op3)`. Will be removed in V2!
```
cjquines, line 6: **todo:** if we're renaming `createPipe`, we should change the mentions in the README and in the docs index page

cjquines (commented):
there should probably be a follow-up pr where the deprecated functions are removed in other places in the codebase (like how we use `isObject` in `isEmpty`, etc)

src/createPipe.ts
eranhirsch: Thanks for catching that! We're thinking of rewriting the readme as part of v2 so we will probably take a closer look at it soon.

eranhirsch:
The main goal for this PR is to give as much heads up as possible to Remeda users about some of the breaking changes that are coming with V2. Once we actually delete the files we'll fix any residuals as we find the issues.

dartess:
> ! DEPRECATED: If your type accepts null use R.isNullish(data), otherwise prefer R.isDefined.strict(data). The non-strict version will be removed in V2!

It's strange that `isDefined` doesn't mention `isNonNullish`, since it's literally the same code and obvious way to quickly migration.

cjquines:
we are recreating `difference` and `intersection`, see https://github.com/remeda/remeda/pull/568#issuecomment-2049419579

it is strange that `isDefined` doesn't mention `isNonNullish`; i agree, we should change the docs

eranhirsch:
I'll make sure difference and intersection would be included in v2 as this came up several times.

PR #580, opened 2024-03-19
chore(lint): turn on more jsdoc lint rules

cjquines:
follow-up #558

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -189,9 +189,8 @@ export default config(
       "jsdoc/require-asterisk-prefix": "error",
 
       // Completeness
-      // TODO: Requires manual fixes, enable in a separate PR.
       "jsdoc/no-restricted-syntax": [
-        "off",
+        "error",
         {
           contexts: [
             {
@@ -208,9 +207,8 @@ export default config(
           ],
         },
       ],
-      // TODO: Requires manual fixes, enable in a separate PR.
       "jsdoc/require-jsdoc": [
-        "off",
+        "error",
         {
           enableFixer: false,
           // We only require JSDoc for top-level function exports. Assuming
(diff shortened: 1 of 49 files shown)


eranhirsch:
Let's wait with this PR for after the v2 release. We haven't decided yet if we are adding dataLast impls to guards, and if not it might mean we need to relax the rule to allow some functions not to be tagged with dataFirst and dataLast.

cjquines:
are the non-datafirst and non-datalast rules fine?

i do think require-jsdoc and require-throws are useful to have on (they caught that the `groupBy` jsdoc was misplaced, for example) and that having `signature` be required is also useful (it caught `times` not having a signature, which causes it not to appear in the docs i think?)

eranhirsch:
> are the non-datafirst and non-datalast rules fine?

We'll have to circle back on this once v2 is out so we can see what the state of the library is. I tend to say I prefer them, but I've also noticed that we could simply drop the tags and the duplication of the docblocks and have one documentation block for both... It's a pain to update docs right now as it's just copy-pasting between the 2 versions to fix the example and signature slightly. It was useful when purrying was done manually, but those exceptions are so rare (I think only zipWith) that we're better off just having less optimal docs for that one case.

cjquines:
bringing back the discussion of having two docs, one advantage is that tsserver can serve different docs for each one, so we can get the descriptions for each parameter while we're typing it. but i can only get this to work in vscode:

image

image

in neovim with tsserver it's a bit different, though maybe it's with how i configured it:

image

image

with only one jsdoc, we don't get the full documentation for the other overloads while we're typing, though we do get it on mouseover

image

image

image

i could do a more careful write-up if we want to investigate this further?

eranhirsch:
Those are good points to keep the duplication, at least for now. We can consider writing a "linting" script that would check that the dataFirst and dataLast description have the same content and fail the build if they don't, but it also makes sense in some cases to have 2 different descriptions. e.g. we can remove the "pipable" tag (that should be renamed to "lazy" in v2) from the dataFirst impls.

eranhirsch:

PR #581, opened 2024-03-20
docs: move mapping to jsdoc

### LINKED ISSUE #519: docs: put mapping in jsdoc
it'd be nice to have the mapping on the docs website; right now it's not mentioned

options:
1. link it on `index.astro` (aside: this has diverged from the repo `README.md`; should that be changed?)
2. parse `mapping.md` to a file that we can in `docs/src/data`, then read it for `mapping.astro`
3. replace `mapping.md` with a data file in `docs/src/data`, then read it for `mapping.astro`
4. make new jsdoc tags, say, `@lodash` and `@ramda`, then read existing jsdoc output for `mapping.astro`

(and other options are possible too)

i personally like 4

cjquines:
Partly resolves #519.

This PR moves most of the info from [mapping.md](https://github.com/remeda/remeda/blob/master/mapping.md) to the jsdoc for each function, and parses the info for the docs site. Format is:

```ts
/**
 * @similarTo <library> <name> [- additional context/differences/notes]
 *
 * (add) Exact copy:
 * @similarTo lodash add
 *
 * (conditional) Changes:
 * @similarTo ramda cond - The Remeda function throws when no condition matches.
 *
 * (zip) Other libraries:
 * @similarTo python zip - The Remeda function returns an array, not an iterator.
 * 
 * (indexBy) Similar remeda functions:
 * @similarTo remeda fromKeys - Creates from an array of keys.
 * 
 * (filter) One-liners:
 * @similarTo lodash intersection - Use `R.filter(items, R.isIncludedIn(other))`.
 */
```

I don't know how/where to show this, but wanted to start this PR so we can get rid of the mapping file.


docs/src/lib/transform.ts
```diff
@@ -70,6 +69,36 @@ function transformCommentDisplayPart(
   return functionNames.has(codeContent) ? `[${text}](#${codeContent})` : text;
 }
 
+function getMapping({ blockTags }: JSONOutput.Comment) {
+  const result = new Array<{
+    library: string;
+    name: string;
+    notes: string | undefined;
+  }>();
```
eranhirsch, line 77: 🧐 
```suggestion
  const result = Array<{
    library: string;
    name: string;
    notes: string | undefined;
  }> = [];
```

docs/src/lib/transform.ts
```diff
@@ -70,6 +69,36 @@ function transformCommentDisplayPart(
...
+  const mappingTags = blockTags.filter(({ tag }) => tag === `@mapping`);
+
+  for (const { content } of mappingTags) {
+    const textContent = content.map(({ text }) => text).join("");
+
+    // Matches "<library> <name> [- notes]":
+    const match = textContent.match(/(\w+) (\w+)(?: - (.+))?/);
+    const [, library, name, notes] = match ?? [];
+
+    if (!library || !name) {
```
eranhirsch, line 92: This should probably throw so we fail the build on bad doc strings.
also, don't use implicit boolean checks, prefer !== undefined, or !== "", or whatever library and name are here (I don't know the type for regex matches).

eranhirsch:
Following your question about discoverability and following recent additions to the "not a mapping but still useful one-liners" section of the mapping file - I think we can expand this concept a bit also to support linking to internal functions and to provide snippets to call this @similarTo or something like that.

So we can do something like:

```ts
* @similarTo remeda fromKeys - To create an object from an array of keys, instead of an array of values.
* @similarTo ramda indexBy
```

cjquines:
like that idea. still a bit torn about how we should use the notes here, and how we can write it so that it's unambiguous

```ts
/**
 * @similarTo python zip - Returns an array, not an iterator.
 */
```

does this mean the python `zip` returns an array, not an iterator? or the remeda `zip` returns an array, not an iterator?

```ts
/**
 * @similarTo remeda fromKeys - To create an object from an array of keys, instead of an array of values.
 */
```

does this mean `fromKeys` creates from an array of keys, or `indexBy`?

eranhirsch:
It all depends on how we expose this in the doc site UI and how it looks like in the IDE when you hover over a function. I don't have an definitive answer for this. See my comment on the other PR, i think we should also invest in checking out if we can write a single docblock for every function instead of 2, that would make it much easier to maintain these mappings inline too.

eranhirsch:

PR #584, opened 2024-03-21
docs(product): wrong return value in example

Masyaka:
Very minor typo fix in jsdoc for datalast `product` function.


eranhirsch (approved):
I think it was copilot who wrote this 😎

PR #587, opened 2024-03-23
feat: add indexed versions

eranhirsch:
Adding the params needed to make all predicate/mapper functions conform to the standard where we include the index and the full array as the last 2 params.


cjquines:
including the full array as one of the params makes it incompatible with laziness / iterators, right?

PR #592, opened 2024-03-24
feat(build): use tsup for building

eranhirsch:
! I'm not sure that tsup is doing what we want it to do: we are outputing a single file that contains all utility functions.

As far as I understand it, that means that the project wouldn't be tree-shakable and thus would cause users to have to import everything.

Based on #360


eranhirsch:
@TkDodo, from your understanding of tree-shaking, is this build process going to break tree-shaking?

eranhirsch:
So this definitely breaks tree-shaking. I'll play around with it and see if there's a way to fix it. @TkDodo, do you have any ideas? Is there anyone that can help us with this?

image

cjquines:
why did we want to use tsup again? isnt the existing build like, fine

eranhirsch:
When running publint on v1 you can see that the build isn't perfect. It's both a little complex and requires us to run typescript 3 times with different settings which is brittle. I think the hope at least was that tsup would take the minutiae of building from us so we don't need to be concerned with esm and cjs shenanigans.

eranhirsch:
See #347

PR #595, opened 2024-03-25
fix(build): Enable tree-shaking

eranhirsch:
Making sure we build the library in a way which would enable bundlers and builders to tree-shake it to just the utilities used.


eranhirsch:
This did it!

image

Issue #593, opened 2024-03-24
`mapKeys`, `mapValues`, `entries`, etc. drop symbol keys without excluding them in the return type


russelldavis:
Example:

```ts
const bar: Record<symbol, number> = {[Symbol()]: 42}
const baz = R.mapKeys(bar, (x) => x)
// bar and baz have the same type, but at runtime baz is empty
```

Note: in v1.x, the return type of `mapValues` excludes symbol keys, but I noticed in the beta branch it no longer does this.

eranhirsch:
Thanks, this is a good observation. I'll take a look. We might just opt to drop symbols from our types...

eranhirsch:
@russelldavis - Can you go over #600  (at least the tests for the relevant functions) and see that we now handle symbols as you'd expect?

cjquines:
i'm gonna close this issue as it's released on beta—i don't think we can backport this change

PR #600, opened 2024-03-25
feat(v2): Fix typing

eranhirsch:
A bunch of fixes to typing as a result of modernizing the codebase, introducing type-fest, and feedback we got in issue #593.

Fixes: #593


src/doNothing.ts
```diff
@@ -20,15 +20,15 @@
...
  */
-export function doNothing(): (
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  ...args: any
-) => void {
-  // Because the function does nothing we safely return the same function
-  // (pointer) and don't need to create a new one on every invocation.
-  return doesNothing;
-}
+export const doNothing = (): typeof doesNothing =>
```
cjquines, line 23: **question** if we do this, then the hover for `doNothing` will show `(): typeof doesNothing` rather than `(): <Args extends ReadonlyArray<unknown>>(...args: Args) => void` right?

think this is less useful, so i think it's better to repeat the type here...

src/product.ts
```diff
@@ -13,9 +14,8 @@ import { purry } from "./purry";
  * @dataFirst
  * @category Number
  */
-export function product<
-  T extends ReadonlyArray<bigint> | ReadonlyArray<number>,
->(data: T): T[number];
+export function product(data: ReadonlyArray<number>): number;
+export function product(data: Readonly<NonEmptyArray<bigint>>): bigint;
```
cjquines, line 18: **question:** why is one `ReadonlyArray` and one `Readonly<NonEmptyArray<...>>`?

src/sum.ts
```diff
@@ -12,9 +13,8 @@ import { purry } from "./purry";
  * @dataFirst
  * @category Number
  */
-export function sum<T extends ReadonlyArray<bigint> | ReadonlyArray<number>>(
-  data: T,
-): T[number];
+export function sum(data: Readonly<NonEmptyArray<bigint>>): bigint;
+export function sum(data: ReadonlyArray<number>): number;
```
cjquines, line 17: **nit:** can we swap the order of these overloads (to be consistent)

src/doNothing.ts
eranhirsch: It looks like VSCode is doing a good job of dereferencing for these sort of things. It most likely not even vscode doing it, but typescript itself, so it's probably the same in IntelliJ

image

cjquines: ah, cool. yes, it it typescript itself (it's tsserver), so it's the same across editors

eranhirsch:
This PR is now ready for review. I'll wait a few days for review before shipping it even if it isn't reviewed so we can start testing an "RC" for the beta release

src/omitBy.ts
cjquines, line 1: **question:** right now, if i call

```ts
const result = omitBy(
  { a: 0 } as { a: number },
  (value): value is 1 => value === 1,
);
```

we get that `result` has type `{ a: number }`. shouldn't this be `{ a?: number }`?

cjquines: there's the symmetric problem with `pickBy`:

```ts
const result = pickBy(
  { a: 0 } as { a: number },
  (value): value is 1 => value === 1,
);
```

gives `result` as type `{}`, which i am again expecting to be `{ a?: number }`

src/randomString.ts
```diff
@@ -1,5 +1,4 @@
 import { purry } from "./purry";
-import { times } from "./times";
```
cjquines, line 2: **praise:** always good to keep our bundle sizes small!

src/setPath.ts
```diff
@@ -1,11 +1,31 @@
-import type { Narrow } from "./_narrow";
-import type { Path, SupportsValueAtPath, ValueAtPath } from "./_paths";
 import { purry } from "./purry";
 
+type Path<T, Prefix extends ReadonlyArray<unknown> = readonly []> =
```
cjquines, line 3: **question:** maybe we should share these types with `pathOr`?

russelldavis:
@eranhirsch looks good, thanks for the ping at https://github.com/remeda/remeda/issues/593#issuecomment-2027530019

src/_types.ts
```diff
@@ -21,20 +21,15 @@ export type IterableContainer<T = unknown> = ReadonlyArray<T> | readonly [];
...
-> = number extends Length
-  ? // Because `Length extends number` and `number extends Length`, then `Length` is not a specific finite number.
-    ReadonlyArray<Element> // It's not fixed length.
-  : BuildTupleHelper<Element, Length, []>; // Otherwise it is a fixed length tuple.
+// Object.values does only return the values paired with non-symbol keys.
+export type EnumeratedValueOf<T> = {
+  [K in keyof T]: K extends symbol
+    ? // Ignore any values that are coming from a symbol key
+      never
+    : Required<T>[K];
```
russelldavis, line 31: Just curious, what is the purpose of `Required<T>[K]` here (instead of just `T[K]`)? I tried changing it to the latter and all tests still pass.

src/_types.ts
```diff
@@ -21,20 +21,15 @@ export type IterableContainer<T = unknown> = ReadonlyArray<T> | readonly [];
 
 // Inspired and largely copied from `sindresorhus/ts-extras`:
 // @see https://github.com/sindresorhus/ts-extras/blob/44f57392c5f027268330771996c4fdf9260b22d6/source/object-keys.ts
-export type ObjectKeys<T extends object> = `${Exclude<keyof T, symbol>}`;
+export type EnumeratedKeyOf<T> = `${Exclude<keyof T, symbol>}`;
```
russelldavis, line 24: This is a nit, but you might consider renaming `EnumeratedKeyOf` (and `EnumeratedValueOf` below) to something like `NonSymbolKeyOf`. The current name makes it sound like it has to do with enumerable vs non-enumerable keys, which is orthogonal to string vs symbol keys.

src/omitBy.ts
```diff
@@ -1,21 +1,103 @@
...
+    // The partial case, props here might be part of the output object, but
+    // might not be, hence they are optional.
+    -readonly [P in keyof T as PropIsPartially<T, P, S>]?: Exclude<T[P], S>;
+  }
+>;
+
+// If the input object's value type extends itself when the type-guard is
+// excluded from it we can safely assume that the predicate would always return
+// `false` for any value of that property.
+type PropIsExact<T, P extends keyof T, S> =
```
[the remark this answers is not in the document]
eranhirsch: We don't... I use them interchangeably myself. Usually, if the type refers to keys, I'd use K; if it refers to props, I'd use P; if it doesn't mention something specific, I prefer P, but I don't always notice. It's good that it's consistent, and I wouldn't want contributors to start using odd choices like A or X, but K and P are similarly readable. I think the typescript docs used to use P, as they now have `Property` as the name in all their examples (https://www.typescriptlang.org/docs/handbook/2/mapped-types.html).

As for prop vs key in general, I don't like the duality, but as you said, the difference is that *property* usually refers to both the key and it's value, but then why is `Object.entries()` not called `Object.properties()` 😭

src/omitBy.ts
eranhirsch: good catch! I'll see what I can come up with

src/setPath.ts
eranhirsch: There is a back-and-forth between overgeneralizing and repeating ourselves. These types were previously generic, widened to fit "the generic case," but only used in one place, and `pathOr` (and I assume other functions with path semantics) built their versions. This results in bad typing for the overgeneralized type and code deterioration for the latter almost identical type whenever a bug is fixed.

My approach to these kinds of things is to keep the types close to the function that introduced them until someone actively wants to re-use them for another function; when that happens, they can pull them into  _types.ts or somewhere else that makes sense and generalize them so they fit the 2 cases (but only those 2).

I'm not a huge fan of the path-related functions so I don't want to overwork trying to generalize the types here. I got here because I was cleaning up and wanted to remove any unused abstractions.

src/omitBy.ts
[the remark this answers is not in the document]
cjquines: alas... we can keep this in mind for docs work

src/_types.ts
eranhirsch: Renamed to EnumerableStringKeyOf, following the definition provided in: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#description

src/_types.ts
eranhirsch: It allows us to remove any `| undefined` from types of optional properties. The type of `T[K]` for an optional K would be `Required<T>[K] | undefined` because that is the type you'd get when you access that property in runtime (e.g. `tInstance.k`). To work around this, we need to tell typescript to treat this "access" as though the prop is not optional (by making all props in T required). If the property naturally is typed with `| undefined,` it would remain because it is not the result of optionality, this is why using `NonNullable<T[K]>` would be wrong here.

https://www.typescriptlang.org/play?exactOptionalPropertyTypes=true#code/C4TwDgpgBAglC8UDeBYAUFKBDAXFAzsAE4CWAdgOYDc6mARnoaZVAD5QCuZAJhAGbkI3GhigBjAPyNi5arSjcpBGS3ZdeAskJEBfEelCRYARgSwA2gCIslgLoiA9A8xQAehPQHw0GACYzMFZ0do7OmO6eaIY+AMwBVmIh6E4uEWheRjAALPGW3EloKeEe6VHeUABCpogAShAAjhwkREIAPDAAfFY29slhbiUZ0BX+tQ1NLdztXZbBvYX9aUOVcWONzW2dCQVFA5HRlTlrE5sz+fO7S2hAA

Notice that this acts differently when `exactOptionalPropertyTypes` is turned off in typescript. We assume the option is turned on for our typing.

src/_types.ts
russelldavis: Still a nit, but I think `Enumerable` in the name is still misleading. It could make sense to name a *function* `enumerableStringKeysOf` (because it describes what the function does as runtime), but this is a type definition, and the type system doesn't distinguish between enumerable and non-enumerable keys.

For example, the builtin type definition for `Error` includes the `message` property. If you call `EnumerableStringKeyOf<Error>`, you'll get back a type that still includes `message`. But if you call `Object.keys(new Error())`, `message` won't be included, because it isn't enumerable.

src/_types.ts
russelldavis: Right, but isn't the optionality being immediately reintroduced when you index by `[keyof T]` on the line below? AFAICT, `EnumeratedValueOf` would behave exactly the same if you replaced `Required<T>[K]` here with just `T[K]`. Is there an example that shows otherwise?

src/pickBy.ts
```diff
@@ -1,53 +1,186 @@
...
+// which value would never come up true for the predicate, AND we can also
+// assume that properties that match the predicate perfectly would **always**
+// show up in the output object. Hence to build the output object we need to
+// build and merge 2 output objects: One for the properties which have a value
+// of at type that would always yield a `true` result from the predicate, these
+// are the "matches", which would not change the "optionality" of the input
+// object's props, and one for partial matches which would also make the props
+// optional (as they could have a value that would be filtered out).
+type EnumeratedPartialNarrowed<T, S> = Simplify<
+  ExactProps<T, S> & PartialProps<T, S>
```
cjquines, line 31: **nit:** `AlwaysPickedProps` and `OptionalPickedProps` maybe?

then we can shorten the comment a bit...


```ts
// When the predicate is a type guard (`value is S`), each property will either:
// - never be S, which we always remove
// - always match S, which we always include (`AlwaysPickedProps`)
// - sometimes match S, which we make optional (`OptionalPickedProps`)
```

cjquines: (and something similar for `omitBy`, though i'm not sure what exactly)

cjquines (approved):
i think i'm happy with these types now, though i'll have to write more tests to be happy...

src/_types.ts
eranhirsch: `Required` here is, in fact, redundant—that's a great observation! I don't remember exactly what my thought process was when building this type. I think the intent of the `Required` was to remove `| undefined` from the types when it is the result of a prop being optional (when using `exactOptionalPropertyTypes`). That is solved in this type by removing the optionality from the keys via the mapper (`-?:`), so this is probably just an earlier attempt that I didn't remove.

src/_types.ts
eranhirsch: Handling the JS built-ins this way is outside the scope of what we try to do with typing in Remeda. In general, we don't test classes in most of our utils; the issue with classes (and interface declaration merging, getter/setter functions, etc...) is that they complicate the typing system considerably, and the main reason that Typescript doesn't provide the sort of types we aim to provide for the built-in functions.
Most people avoid using these features, which allows us to simplify our types. If you already have non-enumerable string keys in an object, it's an advanced use case for which something other than general-purpose helper utils might be a better fit.

Even considering symbols in our types is more advanced than what the vast majority of our users need, and that played into how I named this type; adding "symbols" to the name brings in an advanced concept to a type that would be used by users who have no dealing with symbols in their code, and would force them to consider what this means. Referring to enumerability in the name is a hint to users who do understand what that means that this function has caveats when dealing with these concepts, without negatively impacting the experience for those who don't.

TL;DR - Naming is hard...

src/pickBy.ts
eranhirsch: Sorry, I missed this when merging the PR, i'll take a look and maybe send a PR to fix

Issue #572, opened 2024-03-16
Typescript error "Excessive stack depth" for `hasAtLeast`


zavod008:
![image](https://github.com/remeda/remeda/assets/20419214/57d659cc-f19d-4ead-a594-9ba834d1067f)

 "remeda": "^1.50.1",
 "typescript": "~5.4.2"

eranhirsch:
Can you provide more context on what code is causing this, preferably with a typescript playground link or a code sandbox? I don't know of any specific issue right now with `hasAtLeast` and can't reproduce a trivial one myself:

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgCwIYGcCCMAyBTDeAXzgDMoIQ4AiKPEPAE1WoG4AodgYwgDt14zGKjgBeOAG0AunAxxeAVxAAjPFGkd2wUnAAUaLLgIDdQ1ABo4AZgCUNxOzhwe-CABs8AOjcQA5qdRhGw4nAHpQp0iouAA9AH52InYgA

There is currently an issue with Typescript 5.4 that was recently fixed; I would wait to update Typescript to 5.4 until version 5.4.3 comes out.

eranhirsch:
See #559 (although I can't say if they are related yet)

zavod008:
Sorry, I don't have time to pay much attention to this issue, for myself I just rolled back

cjquines:
i don't think it's related to #559 , since that's been fixed in the nightly and [the type is still broken](https://www.typescriptlang.org/play?ts=5.5.0-dev.20240319#code/C4TwDgpgBAShCGATA9gOwDYgCoFczogB4sAaKAOSggA9gJVEBnKVHAWwCMIAnMmK2vSZQsAbQC6UALxQJAPmkAoAJAxRAcgKoA5sAAW6yTToNm5RVEtQA-FG4IUGELIB0bmOItWAXLAdpMXHwiUgoyUVC3Fw85AG5FRVBIKABBbm54EABZAEtUcmIySmMhZlZOHgUZcq5uARNhc0tbNIyQYjkvX2LBUygABi9m1PTMjqGoX1EouCQA7DwCQoo5MijWsaw5cXjFRAgAY3R4eygAMxxUA+ActCg9eEYU4AAZBEZgZZ6GsvZauQAFF5EPBgPBfLNHJgNu0tiQvGw8jk2OxuvCAJS+EFgqA5Zgw3L5L5xIA)

[...]

PR #604, opened 2024-03-29
feat: Support TypeScript 5.4

eranhirsch:
The main thing is a fix to typing for hasAtLeast.

The fixes here are mostly a patch so that users can upgrade to typescript 5.4, proper and better fixes are planned for v2.

closes: #572


src/hasAtLeast.ts
```diff
@@ -1,14 +1,41 @@
...
 
-type ArrayMinN<T, N extends number> = number extends N
-  ? // We can only compute the type for a literal number!
-    Array<T>
-  : // I don't know why we need to special-case the 0 case, but otherwise
-    // typescript complains we have a deep recursion.
-    N extends 0
-    ? Array<T>
-    : [...ReadonlyTuple<T, N>, ...Array<T>];
+type ArraySetRequired<
```
cjquines, line 4: **question:** should we file an issue with type-fest?

cjquines, line 4: **question:** why the rename? i think i understood `ArrayMinN` better than this

src/hasAtLeast.ts
```diff
@@ -1,14 +1,41 @@
...
+  T extends IterableContainer,
+  N extends number,
+  Iteration extends ReadonlyArray<unknown> = [],
```
cjquines, line 7: **nit:** maybe `Result`?

cjquines (approved):

src/hasAtLeast.ts
eranhirsch: We used a copy-pasted version from a few months ago; I didn't check back to see if they changed their implementation for ts5.4. Re-reading our code, I thought that building our own version made sense here.

Generally, my feelings towards type-fest shifted with the work for v2, as many type-fest types don't make sense in real-world use cases, feel more "academic"/"challenge-like"  than practical, and aren't always helpful when typing our functions. I'd rather reduce our surface area to where we use their types as-is when they fit perfectly (or we adjust the runtime behavior to match their type).

eranhirsch: ArrayMinN took the **item's type** as T, just like Array does (Array<T> vs ArrayMinN<T, N>), but now the function takes the **input array** as T. It works more similarly to type-fests [SetRequired](https://github.com/sindresorhus/type-fest/blob/main/source/set-required.d.ts) where T is the object that is being transformed, and the type name describes the transformation (and not the output).

src/hasAtLeast.ts
eranhirsch: Oops, I was supposed to add `undefined` to this type, not the Head. Notice that we never return this type, we return T from within the recursion. This is just the regular typescript trick to stop recursions after N iterations.

Issue #603, opened 2024-03-29
isIncludedIn narrows the negated cases too aggressively 


jinyongp:
```ts
import * as R from 'remeda';

const a1 = [1, 2, 3, 4, 5];
const a2 = [3, 4, 5, 6, 7];

// inferred type is number[]
const r1 = R.difference(a1, a2); // why deprecated?

// inferred type is never[] but should be number[]
const r2 = R.filter(a1, R.isNot(R.isIncludedIn(a2))); // r2 type is never[] but should be number[]

// inferred type is (data: unknown) => data is unknown but should be (data: number) => data is number
const n = R.isNot(R.isIncludedIn(a2));

// inferred type is unknown[] but should be number[]
const r4 = R.filter(a1, n);
```

I think the easiest way to solve the problem is to use `difference` instead of deprecated.

eranhirsch:
Good catch, thanks for reporting, `isIncludedIn` is new and there were obvious edge-cases I missed when adding it. I'll take a look and send a fix for this.

Difference and intersect are deprecated because their semantics are inconsistent with set theory(!) We want to add replacements for these functions using the same name, but to do that, we need to "free up the namespace"; otherwise, we will introduce breaking changes that people will not notice. We added the comment about deprecation so that people can migrate off the existing functions but still have the functionality in the interim and then migrate back to difference and intersect using the new logic once they assess that it makes sense for them (for most people, it will).

PR #607, opened 2024-03-31
fix(isIncludedIn): Only narrow on exact cases

eranhirsch:
Because of limitations of how typescript handles literal unions (and the lack of ability to detect them in the typing system) certain cases narrow the negated case too aggressively.

e.g. If our data is of type `1 | 2 | 3` and our container is `Array<1 | 2>` we can say that the data could be narrowed to `1 | 2`, but we can't say that the negated case could be narrowed to `3` because the container array could still be empty.

I couldn't find a way to tell Typescript to narrow only the successful result without impacting the negative one, so we can only reduce the "surface" for the narrowing variant of the function to those that make the most sense. Even with this, there are erroneous cases, but at least they are reduced to a very small subset.

Fixes: #603

### DIFF
--- a/src/isIncludedIn.ts
+++ b/src/isIncludedIn.ts
@@ -1,38 +1,73 @@
 import type { IterableContainer } from "./_types";
+import type { IsNotFalse } from "./type-fest/internal";
 import type { IsLiteral } from "./type-fest/is-literal";
 
-type IsStrictTuple<T extends IterableContainer> = T extends readonly []
+/**
+ * A "pure" tuple is one that doesn't contain any optional or variadic parts,
+ * i.e. it has a finite and constant number of elements in it.
+ */
+type IsPureTuple<T extends IterableContainer> = T extends readonly []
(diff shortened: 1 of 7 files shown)


src/isIncludedIn.ts
```diff
@@ -1,11 +1,68 @@
...
+      IsNotFalse<IsLiteral<S[number]>>;
 
 /**
  * Checks if the item is included in the container. This is a wrapper around
  * `Array.prototype.includes` and `Set.prototype.has` and thus relies on the
  * same equality checks that those functions do (which is reference equality,
- * e.g. `===`). The input's type is narrowed to the container's type if
- * possible.
+ * e.g. `===`). In some cases the input's type is narrowed to the container's
+ * type if possible.
```
cjquines, line 65: **nit:** i would remove the "if possible", this is a bit too much hedging :shrug:

cjquines (approved):

PR #621, opened 2024-04-08
fix: "no-dangle-underscore", and other nits.

eranhirsch:
And follow up with going over new rules and old rules and making sure our config is top-notch.


eranhirsch:
This will have to wait for typescript-eslint to also update to v9, in the meantime this PR can't build in CI.

eslint.config.js
```diff
@@ -135,12 +122,50 @@ export default config(
...
+      "prefer-spread": "error",
+      "prefer-template": "error",
+      radix: "error",
+      "require-unicode-regexp": "warn",
+      strict: "error",
+      "symbol-description": "warn",
+      "unicode-bom": ["error", "never"],
+      "vars-on-top": "warn",
+
+      // Code Quality "Params"
```
cjquines, line 155: **thought:** i am not usually a fan of these kinds of rules, but these are loose enough that i'm fine with this

src/pipe.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable complexity, max-depth, max-lines, max-statements -- TODO: We should strive to simplify the pipe implementation! */
```
cjquines, line 1: **thought:** see, if we used iterators to deal with laziness, i think we can cut a lot of the complexity here

cjquines: but of course, it would require a somewhat-substantial rewriting of things

src/sample.ts
```diff
@@ -102,6 +102,7 @@ export function sample(...args: ReadonlyArray<unknown>): unknown {
   return purry(sampleImplementation, args);
 }
 
+// eslint-disable-next-line max-statements -- TODO: Can we simplify or split this?
```
cjquines, line 105: **question:** i see this function has a lot of guards, which we don't really have in other functions. for example, for `chunk`, we don't check that `size` is a positive integer. if we eliminate this, that'd bring us under the limit right?

cjquines (approved):

src/sample.ts
eranhirsch: I think you are right and it also looks like how the built-in functions chose to handle this: for example `[1,2,3].slice(1.5)`.

I'll remove them

eranhirsch:
It's taking the community longer than I thought to migrate to eslint v9, and the resulting changes in this PR don't need v9 to work - so we can ship the changes to the config itself later.

This mainly removes the underscore-based private functions in favor of proper names and directories.

Also removes runtime protections from sample which was the only function we had with runtime protections.

PR #623, opened 2024-04-08
fix(last): Improve `last` typings for tuples

Brodzko:
An attempt to consolidate typings and some naming conventions between `first` and `last`. I tried catching all cases with tests but let me know if you spot a problem that might be solvable.

Would it be beneficial to have a look at typings for `take` and `drop` in a similar manner?

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/first.test.ts
+++ b/src/first.test.ts
@@ -9,7 +9,7 @@ function defaultTo<T>(d: T) {
   };
 }
 
-test("should return last", () => {
+test("should return first", () => {
   expect(first([1, 2, 3] as const)).toEqual(1);
 });
 
--- a/src/last.test.ts
+++ b/src/last.test.ts
@@ -47,25 +47,16 @@ describe("last", () => {
       assertType<undefined>(data);
     });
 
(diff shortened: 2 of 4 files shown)


cjquines:
i think this is a good idea, but i would rather we merge a commit into the beta branch, and use the type-fest types `LastArrayElement` for `last` (and possibly `ArraySlice` for `take` and `drop`). wdyt @eranhirsch ?

cjquines:
but also... we can copy the type-fest types (like we currently do) and use those, then forward-port this to the beta branch. i'd support having better typing in v1

Brodzko:
> but also... we can copy the type-fest types (like we currently do) and use those, then forward-port this to the beta branch. i'd support having better typing in v1

I think I originally tried using `LastArrayElement` but it didn't seem to work well with `NonEmptyArray`s (noc including `undefined` in those cases). I'll have another look soon

Brodzko:
@cjquines Updated with `LastArrayElement` from type-fest, type tests seem OK

cjquines (approved):

PR #629, opened 2024-04-10
docs(v2): integrated migration guide

eranhirsch:
Adding a new section to the docs with a function-by-function migration guide.

Using the Astro content API for a lightweight CMS that is synced with our functions.

Fun!

localhost_4321_v1_


cjquines (approved):
is this how docs websites over versions are usually managed? the step of copying things from v1 feels especially fragile, though i suppose if we don't expect to update it as much after releasing v2, it might be fine

eranhirsch:
> is this how docs websites over versions are usually managed? the step of copying things from v1 feels especially fragile.

I don't know how others handle this; I just looked for the easiest way to provide migration hints "inline" with the docs.

We aren't planning on back-porting changes to v1 once v2 ships, so this would remain static.

PR #630, opened 2024-04-10
fix(stringToPath): Fix broken types for `stringToPath`

Brodzko:
Tests only covered paths with single-character keys. While the implementation works correctly even for longer keys, types were actually inferred wrong:

```ts
const path = R.stringToPath("foo.bar");
          // ^?  would be ["f", "o", "o", "b", "a", "r"]
```
Reworked the `StringToPath` type and added tests covering more broad edge-cases.

I am not entirely clear why the "handles dot paths inside bracket access" is necessary, or even why it makes sense, since e.g. `"foo[bar.baz]"` is not a valid object path. 

I'm not sure if `pathOr` resolves an array access (or how), but even ignoring this was created to support `pathOr`, shouldn't we only allow integers between `[` and `]`, and potentially parse them out as numbers too? 🤔 

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/stringToPath.test.ts
+++ b/src/stringToPath.test.ts
@@ -47,15 +47,6 @@ describe("types", () => {
     expectTypeOf(data).toEqualTypeOf<["foo", "bar.baz", "qui"]>();
   });
 
-  test("string constructed from const strings are inferred", () => {
-    const bar = "bar";
-    const baz = "baz";
-    const qui = "qui";
-    const input = `foo[${bar}.${baz}].${qui}.che`;
-    const data = stringToPath(input);
-    expectTypeOf(data).toEqualTypeOf<["foo", "bar.baz", "qui", "che"]>();
-  });
-
   test("dynamic strings cannot be inferred", () => {
     const bar = "bar" as string;
     const input = `foo.${bar}[baz]`;
--- a/src/stringToPath.test.ts
+++ b/src/stringToPath.test.ts
@@ -20,24 +20,6 @@ describe("runtime", () => {
     const res = stringToPath("foo.bar[3]");
     expect(res).toEqual(["foo", "bar", "3"]);
   });
-
-  test("should handle relatively long paths without performance issues", () => {
-    const res = stringToPath(
-      "lorem.ipsum[dolor.sit].amet.con.sec.tetur[adi.pisc.ing].elit.42",
-    );
-    expect(res).toEqual([
-      "lorem",
-      "ipsum",
-      "dolor.sit",
-      "amet",
-      "con",
-      "sec",
-      "tetur",
-      "adi.pisc.ing",
-      "elit",
-      "42",
-    ]);
-  });
 });
 
 describe("types", () => {
@@ -47,10 +29,30 @@ describe("types", () => {
     expectTypeOf(data).toEqualTypeOf<["foo", "bar.baz", "qui"]>();
   });
 
+  test("should handle relatively long paths without performance issues", () => {
+    const res = stringToPath(
+      "lorem.ipsum[dolor.sit].amet.con.sec.tetur[adi.pisc.ing].elit.42",
+    );
+    expectTypeOf(res).toEqualTypeOf<
+      [
+        "lorem",
+        "ipsum",
+        "dolor.sit",
+        "amet",
+        "con",
+        "sec",
+        "tetur",
+        "adi.pisc.ing",
+        "elit",
+        "42",
+      ]
+    >();
+  });
+
   test("dynamic strings cannot be inferred", () => {
     const bar = "bar" as string;
     const input = `foo.${bar}[baz]`;
     const data = stringToPath(input);
-    expectTypeOf(data).toEqualTypeOf<Array<string>>();
+    expectTypeOf(data).toEqualTypeOf<never>();
   });
 });
--- a/src/stringToPath.test.ts
+++ b/src/stringToPath.test.ts
@@ -1,57 +1,68 @@
 import { stringToPath } from "./stringToPath";
 
-test("should convert a string to a deeply nested path", () => {
-  const res = stringToPath("foo.bar[3].baz");
-  expect<["foo", "bar", "3", "baz"]>(res).toEqual(["foo", "bar", "3", "baz"]);
-});
+describe("stringToPath", () => {
+  describe("implementation", () => {
+    test("should convert a string to a deeply nested path", () => {
+      const res = stringToPath("foo.bar[3].baz");
+      expect(res).toEqual(["foo", "bar", "3", "baz"]);
+    });
 
-test("should handle nested dot paths", () => {
-  const res = stringToPath("foo[bar.baz].qui.che");
-  expect<["foo", "bar.baz", "qui", "che"]>(res).toEqual([
-    "foo",
-    "bar.baz",
-    "qui",
-    "che",
-  ]);
-});
+    test("should handle nested dot paths", () => {
+      const res = stringToPath("foo[bar.baz].qui.che");
+      expect(res).toEqual(["foo", "bar.baz", "qui", "che"]);
+    });
 
-test("should handle short path with only bracket access", () => {
-  const res = stringToPath("foo[bar]");
-  expect<["foo", "bar"]>(res).toEqual(["foo", "bar"]);
-});
+    test("should handle short path with only bracket access", () => {
+      const res = stringToPath("foo[bar]");
+      expect(res).toEqual(["foo", "bar"]);
+    });
 
-test("should handle bracket access at the end", () => {
-  const res = stringToPath("foo.bar[3]");
-  expect<["foo", "bar", "3"]>(res).toEqual(["foo", "bar", "3"]);
-});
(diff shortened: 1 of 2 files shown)


src/stringToPath.test.ts
cjquines, line 1: **question:** can we add explicit type tests? in particular, what happens when the string passed in is constructed dynamically, like

```ts
stringToPath(`foo[${bar.baz}].${qui}`);
```

cjquines (changes requested):
> I am not entirely clear why the "handles dot paths inside bracket access" is necessary, or even why it makes sense, since e.g. `"foo[bar.baz]"` is not a valid object path.
> 
> I'm not sure if `pathOr` resolves an array access (or how), but even ignoring this was created to support `pathOr`, shouldn't we only allow integers between `[` and `]`, and potentially parse them out as numbers too? 🤔

these are good points! i think we should in fact only allow integers between brackets, and parse them as numbers, because that's what `setPath` expects.

Brodzko:
> i think we should in fact only allow integers between brackets, and parse them as numbers, because that's what `setPath` expects.

Would that be considered a breaking change right now?

cjquines:
i'd argue that the previous behavior was a bug, as the intent is for this to be usable in `setPath`?

Brodzko:
> i'd argue that the previous behavior was a bug, as the intent is for this to be usable in `setPath`?

Personally that makes sense to me, but since it was _intentionally_ made that way, I wanted to make sure. Anyhow, I'll post an issue and would prefer to get back to that in another PR if that's OK :)

cjquines:
alright. i'd still push for adding typing tests though, while we're at it

src/stringToPath.test.ts
Brodzko: Added some explicit tests - right away caught that if the input type is `string` the types were broken, so fixed that too

src/stringToPath.test.ts
```diff
@@ -1,11 +1,68 @@
...
+        "42",
+      ]);
+    });
+  });
+
+  describe("types", () => {
+    test("simple const string are inferred", () => {
+      const input = "foo[bar.baz].qui";
+      const data = stringToPath(input);
+      assertType<["foo", "bar.baz", "qui"]>(data);
```
cjquines, line 48: **nit:** let's use the `expectTypeOf(...).toEqualTypeOf<...>()` form here, which does a stricter two-way comparison

src/stringToPath.test.ts
```diff
@@ -1,11 +1,68 @@
...
+      const data = stringToPath(input);
+      assertType<["foo", "bar.baz", "qui", "che"]>(data);
+    });
 
-test("should handle nested dot paths", () => {
-  const res = stringToPath("a.b[a.b].c");
-  expect<["a", "b", "a.b", "c"]>(res).toEqual(["a", "b", "a.b", "c"]);
+    test("dynamic strings cannot be inferred", () => {
+      // eslint-disable-next-line @typescript-eslint/no-inferrable-types
+      const bar: string = "bar";
```
cjquines, line 62: **nit:** we can skip the lint rule with a `"bar" as string` i think?

cjquines (approved):

src/stringToPath.test.ts
```diff
@@ -1,11 +1,68 @@
 import { stringToPath } from "./stringToPath";
 
-test("should convert a string to a deeply nested path", () => {
-  const res = stringToPath("a.b[0].c");
-  expect<["a", "b", "0", "c"]>(res).toEqual(["a", "b", "0", "c"]);
-});
+describe("stringToPath", () => {
```
eranhirsch, line 3: no need to wrap the whole thing in a describe, the tests are scoped with the file name, it just adds another level of indentation.

src/stringToPath.test.ts
```diff
@@ -1,11 +1,68 @@
...
+  describe("implementation", () => {
```
eranhirsch, line 4: my convention is to call this group "runtime", types are 'implemented' too

src/stringToPath.test.ts
eranhirsch: yes, in general, `as <type>` is also safer for tests because it forces typescript to treat it as that type, and `: <type> =` actually focuses typescript more on the validity of the assignment itself; it can still widen or narrow the type based on the rest of the code.

src/stringToPath.test.ts
```diff
@@ -1,11 +1,68 @@
...
+      assertType<["foo", "bar.baz", "qui"]>(data);
+    });
+
+    test("string constructed from const strings are inferred", () => {
+      const bar = "bar";
+      const baz = "baz";
+      const qui = "qui";
+      const input = `foo[${bar}.${baz}].${qui}.che`;
+      const data = stringToPath(input);
+      assertType<["foo", "bar.baz", "qui", "che"]>(data);
```
eranhirsch, line 57: assertType is a softer assertion, it only checks that data extends the type, not that they are equal (e.g. the type might not extend data).

```suggestion
      expectTypeOf(data).toEqualTypeOf<["foo", "bar.baz", "qui", "che"]>();
```

src/stringToPath.test.ts
```diff
@@ -1,11 +1,68 @@
...
+      const input = "foo[bar.baz].qui";
+      const data = stringToPath(input);
+      assertType<["foo", "bar.baz", "qui"]>(data);
+    });
+
+    test("string constructed from const strings are inferred", () => {
+      const bar = "bar";
+      const baz = "baz";
+      const qui = "qui";
+      const input = `foo[${bar}.${baz}].${qui}.che`;
```
eranhirsch, line 55: This isn't testing what you think it's testing, it's equivalent to just a const string, typescript doesn't do anything different with this type:

https://www.typescriptlang.org/play?#code/MYewdgzgLgBARgQwE4wLwwESKRg3AKFElkQC81My9DxoYBHAVwEsKMnnqi7mwAHRrHQADAGYgQAbQAkAb2wBfAHRyyCgLorZHZcAAWAU2EEA9CZgWYAPQD8+GsRi8BUAEwUxEydiVlNHJX0jAnwoAE8+AxgADQpwyJBRJ35BGAMADygDMAATCBh4g0Tkl1d8M0tbfAsbAoiipOdBdwys3PzC4qaoastaqCRGA16LAC4YUQQAGwhhsYnp2YIgA


```suggestion
      const input = `foo[bar.baz].qui.che`;
```

src/stringToPath.ts
```diff
@@ -1,9 +1,21 @@
+type StringToPath<T extends string> = string extends T
+  ? Array<string>
```
eranhirsch, line 2: Is there any value in this type for this case? we can also return `never` here instead so that we "fail up" and users get a compile time error when using this function wrongly.

eranhirsch (changes requested):

src/stringToPath.test.ts
Brodzko: It's testing exactly what I wanted to test, see https://github.com/remeda/remeda/pull/630#discussion_r1560139684 - I know it seems kind of like an obvious test-case, but I think it doesn't hurt to have it explicitly in the suite - shows we've thought of the case?

src/stringToPath.ts
Brodzko: Well I think the question is more "should the type reflect what the function does at runtime _as best as it can_?" vs "should the type protect TS users from accidentally using this function wrongly?"

As it is, if they get the `string[]` inferred, they will already get a compile error if they try to use the result with either `pathOr` or `setPath`, so the "intended" use-cases are protected. 

Returning `never` would imo make sense if we also make the runtime throw/return `undefined` or similar and we disallow "unknown" paths entirely. Right now it at least allows users to just _convert_ dotted paths to tupled paths even if they happen to not care for the strict safety/use it with other safe functions. WDYT?

src/stringToPath.test.ts
eranhirsch: This test and the test above it test the same thing, we don't need to test that typescript can't infer literal strings, that's on the typescript team to do. There are no cases where this test would break and the previous test wouldn't, and vice-versa. In that regard the syntax here is confusing as it looks like something it isn't, and might throw off a maintainer in the future.

Brodzko:
@cjquines what do you think?

src/stringToPath.ts
cjquines: hm. i do think that by far the most common case will be using it with constants, and that's also by far the most useful case. it's true that there's no way in the runtime level to throw if a path is "unknown", but that _is_ what the type system is for

in the case when someone _is_ using a path with variables, then it's not hard to convert it to an array anyway, which can be used with setPath. this is [recommended style when working with lodash](https://github.com/wix-incubator/eslint-plugin-lodash/blob/master/docs/rules/path-style.md) anyway

and it's always easier to expand in the future than contract

cjquines:
oops, fell through my notifs somehow. reviewed

src/stringToPath.ts
Brodzko: Alrighty, updated to infer `never` in case a `string` is passed in :)

Issue #624, opened 2024-04-08
Add `some` and `every`?


Brodzko:
Doing a bit of refactoring, going through the usages of `lodash` in our codebase and replacing them with `remeda` alternatives, I noticed there is no `some` and `every` alternatives for arrays. Is this intentional? I guess you could accomplish the same with
```ts
const every =
  <T>(fn: (val: T) => boolean) =>
  (arr: T[]) =>
    R.pipe(
      arr,
      R.map(x => () => fn(x)),
      R.allPass
    ); // similar for `some`
```
but that seems pretty clunky. Is the omitting of `some` and `every` intentional? If you're interested I'm happy to open a PR.

cjquines:
i think so. for example, i believe the only reason we have `map` and `filter` is because they're useful in pipes and have better than default types. but `some` and `every` aren't as useful in the middle of pipes, because typically you want to `R.pipe(...).some(...)` and `R.pipe(...).every(...)`. the other case i would think of where `every` would be useful is something like

```ts
const data: Array<string | number> = ["a", "b"];

if (data.every(R.isString)) {
  data; // Array<string>
}
```

but [Typescript already does this](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAKjgQwM5wEpwGZQiOAIigFMQSATZQgbgCg6KSBjAG2VLmYgDtV4qMZAC44AQShRkATwA8-KMB4BzOAB84PAK4gARiSgA+enWDY4ACkHIAdCQBuB6RYw3gqAMoxFKgJS+4AG86ODhrELgAeki4AD0AfjoAXyA).

Brodzko:
Right! I guess I can kind of see them being useful in control flow, for example your latter case, instead of having to write `if` statements, I could have a functional form like

```ts
const data: Array<string | number> = ["a", "b"];

const result = R.pipe(
  data
  R.conditional(
    [R.every(R.isString), doStringStuff],
    R.conditional.defaultCase(doLooseStuff)
  )
);
```

cjquines:
hm... i'd defer to the judgment of @eranhirsch here, because i don't have a clear vision myself of what the library should be. it's certainly changed over time; compare #54 to the set of functions we have now. i do think `some` and `every` are nice, composable "building blocks", and it does make the code easier to read in the example you gave, compared to the alternative...

eranhirsch:
[...]

PR #631, opened 2024-04-10
feat(every): Add `every`

Brodzko:
Following up on conversation in https://github.com/remeda/remeda/issues/624, I started with implementing `every`.

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/mapping.md
+++ b/mapping.md
@@ -27,8 +27,7 @@ documentation when migrating._
 | [`dropLast`](https://remedajs.com/docs/#dropLast)                   | [`dropRight`](https://lodash.com/docs/4.17.15#dropRight)                 | [`dropLast`](https://ramdajs.com/docs/#dropLast)                   |
 | [`dropLastWhile`](https://remedajs.com/docs/#dropLastWhile)         | [`dropRightWhile`](https://lodash.com/docs/4.17.15#dropRightWhile)       | [`dropLastWhile`](https://ramdajs.com/docs/#dropLastWhile)         |
 | [`dropWhile`](https://remedajs.com/docs/#dropWhile)                 | [`dropWhile`](https://lodash.com/docs/4.17.15#dropWhile)                 | [`dropWhile`](https://ramdajs.com/docs/#dropWhile)                 |
-| [`entries`](https://remedajs.com/docs/#entries)                     |
-| [`toPairs`](https://lodash.com/docs/4.17.15#toPairs)                | [`toPairs`](https://ramdajs.com/docs/#toPairs)                           |
+| [`entries`](https://remedajs.com/docs/#entries)                     | [`toPairs`](https://lodash.com/docs/4.17.15#toPairs)                     | [`toPairs`](https://ramdajs.com/docs/#toPairs)                     |
 | [`every`](https://remedajs.com/docs/#every)                         | [`every`](https://lodash.com/docs/4.17.15#every)                         |
 | [`evolve`](https://remedajs.com/docs/#evolve)                       |                                                                          | [`evolve`](https://ramdajs.com/docs/#evolve)                       |
 | [`filter`](https://remedajs.com/docs/#filter)                       | [`filter`](https://lodash.com/docs/4.17.15#filter)                       | [`filter`](https://ramdajs.com/docs/#filter)                       |
--- a/mapping.md
+++ b/mapping.md
@@ -111,6 +111,7 @@ documentation when migrating._
 | [`unique`](https://remedajs.com/docs/#unique)                       | [`uniq`](https://lodash.com/docs/4.17.15#uniq)                           | [`uniq`](https://ramdajs.com/docs/#uniq)                           |
 | [`uniqueBy`](https://remedajs.com/docs/#uniqueBy)                   | [`uniqBy`](https://lodash.com/docs/4.17.15#uniqBy)                       | [`uniqBy`](https://ramdajs.com/docs/#uniqBy)                       |
 | [`uniqueWith`](https://remedajs.com/docs/#uniqueWith)               | [`uniqWith`](https://lodash.com/docs/4.17.15#uniqWith)                   | [`uniqWith`](https://ramdajs.com/docs/#uniqWith)                   |
+| [`every`](https://remedajs.com/docs/#every)                         | [`every`](https://lodash.com/docs/4.17.15#every)                         |                                                                    |
 
 ## Helpful one-liners
 
--- a/mapping.md
+++ b/mapping.md
@@ -27,7 +27,9 @@ documentation when migrating._
 | [`dropLast`](https://remedajs.com/docs/#dropLast)                   | [`dropRight`](https://lodash.com/docs/4.17.15#dropRight)                 | [`dropLast`](https://ramdajs.com/docs/#dropLast)                   |
 | [`dropLastWhile`](https://remedajs.com/docs/#dropLastWhile)         | [`dropRightWhile`](https://lodash.com/docs/4.17.15#dropRightWhile)       | [`dropLastWhile`](https://ramdajs.com/docs/#dropLastWhile)         |
 | [`dropWhile`](https://remedajs.com/docs/#dropWhile)                 | [`dropWhile`](https://lodash.com/docs/4.17.15#dropWhile)                 | [`dropWhile`](https://ramdajs.com/docs/#dropWhile)                 |
-| [`entries`](https://remedajs.com/docs/#entries)                     | [`toPairs`](https://lodash.com/docs/4.17.15#toPairs)                     | [`toPairs`](https://ramdajs.com/docs/#toPairs)                     |
+| [`entries`](https://remedajs.com/docs/#entries)                     |
+| [`toPairs`](https://lodash.com/docs/4.17.15#toPairs)                | [`toPairs`](https://ramdajs.com/docs/#toPairs)                           |
+| [`every`](https://remedajs.com/docs/#every)                         | [`every`](https://lodash.com/docs/4.17.15#every)                         |
 | [`evolve`](https://remedajs.com/docs/#evolve)                       |                                                                          | [`evolve`](https://ramdajs.com/docs/#evolve)                       |
 | [`filter`](https://remedajs.com/docs/#filter)                       | [`filter`](https://lodash.com/docs/4.17.15#filter)                       | [`filter`](https://ramdajs.com/docs/#filter)                       |
 | [`find`](https://remedajs.com/docs/#find)                           | [`find`](https://lodash.com/docs/4.17.15#find)                           | [`find`](https://ramdajs.com/docs/#find)                           |
@@ -111,7 +113,6 @@ documentation when migrating._
 | [`unique`](https://remedajs.com/docs/#unique)                       | [`uniq`](https://lodash.com/docs/4.17.15#uniq)                           | [`uniq`](https://ramdajs.com/docs/#uniq)                           |
 | [`uniqueBy`](https://remedajs.com/docs/#uniqueBy)                   | [`uniqBy`](https://lodash.com/docs/4.17.15#uniqBy)                       | [`uniqBy`](https://ramdajs.com/docs/#uniqBy)                       |
 | [`uniqueWith`](https://remedajs.com/docs/#uniqueWith)               | [`uniqWith`](https://lodash.com/docs/4.17.15#uniqWith)                   | [`uniqWith`](https://ramdajs.com/docs/#uniqWith)                   |
-| [`every`](https://remedajs.com/docs/#every)                         | [`every`](https://lodash.com/docs/4.17.15#every)                         |                                                                    |
 
 ## Helpful one-liners
 
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -12,6 +12,20 @@ export type PredIndexedOptional<T, K> = (
   array?: ReadonlyArray<T>,
 ) => K;
 
+export type TypePred<T, S extends T> = (input: T) => input is S;
+
+export type TypePredIndexed<T, S extends T> = (
+  input: T,
+  index: number,
+  array: ReadonlyArray<T>,
+) => input is S;
+
+export type TypePredIndexedOptional<T, S extends T> = (
+  input: T,
+  index?: number,
+  array?: ReadonlyArray<T>,
+) => input is S;
+
 export type NonEmptyArray<T> = [T, ...Array<T>];
 
 export type Mapped<T extends IterableContainer, K> = {
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -12,20 +12,6 @@ export type PredIndexedOptional<T, K> = (
   array?: ReadonlyArray<T>,
 ) => K;
 
-export type TypePred<T, S extends T> = (input: T) => input is S;
-
-export type TypePredIndexed<T, S extends T> = (
-  input: T,
-  index: number,
-  array: ReadonlyArray<T>,
-) => input is S;
-
-export type TypePredIndexedOptional<T, S extends T> = (
-  input: T,
-  index?: number,
-  array?: ReadonlyArray<T>,
-) => input is S;
-
 export type NonEmptyArray<T> = [T, ...Array<T>];
 
 export type Mapped<T extends IterableContainer, K> = {
--- a/src/conditional.test.ts
+++ b/src/conditional.test.ts
@@ -1,5 +1,6 @@
 import { conditional } from "./conditional";
 import { equals } from "./equals";
+import { every } from "./every";
 import { isNumber } from "./isNumber";
 import { isString } from "./isString";
 import { pipe } from "./pipe";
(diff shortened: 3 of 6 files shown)


src/every.ts
```diff
@@ -0,0 +1,100 @@
...
+  predicate: Pred<T, boolean>,
+): boolean;
+
+/**
+ * Checks if every element of `data` passed the predicate `predicate`.
+ *
+ * NOTE: If you want to use this inside `conditional` predicates and preserve proper type inference,
+ * make sure to define the final type guard outside of the pipe. This is due to Typescript's limitation
+ * in generic type inference. See examples below for details.
+ *
```
cjquines, line 41: **question:** what do you feel about dropping this? i think this is a problem with `conditional` and not `every`, and so notes like this should go into usage of `conditional` (because they also apply to other predicates)

cjquines: also, i would say things like this should go in one of the conditional tests as like, a known issue

cjquines: (rather than the jsdoc, which appears every time we hover over the function)

src/every.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      input,
+      every((val) => {
+        expectTypeOf(val).toEqualTypeOf<number | string>();
+        return true;
+      }),
+    );
+    expect(data).toBe(true);
+  });
+
+  test("narrows types when used with filter", () => {
```
cjquines, line 83: **nit:** i wouldnt really think of this as a unit test... it tests `filter` more than it does `every` here, since we do already have a test of `every` narrowing properly?

src/every.ts
```diff
@@ -0,0 +1,100 @@
...
+ * @param predicate - A predicate function to run.
+ * @returns A `boolean` indicating if every element in `data` passes the predicate `pred`.
+ * @signature
+ *    R.every(arr, predicate)
+ * @example
+ *    R.every([1, 2, 3], R.isNumber) // => true
+ *    R.every(['a', 0, 'b'], R.isString) // => false
+ * @dataFirst
+ * @indexed
+ * @pipeable
```
cjquines, line 23: **issue:** i think we only use `@pipeable` for lazy functions, so we should remove this

cjquines: (or at least, we should)

cjquines:
oops sorry for the late review, was busy with my Day Job

src/every.ts
eranhirsch: My first pr after releasing v2 would be to rename pipeable as lazy

Brodzko:
@cjquines no worries, same here. Pushed updates to your previous comments :)

mapping.md
```diff
@@ -27,7 +27,9 @@ documentation when migrating._
 | [`dropLast`](https://remedajs.com/docs/#dropLast)                   | [`dropRight`](https://lodash.com/docs/4.17.15#dropRight)                 | [`dropLast`](https://ramdajs.com/docs/#dropLast)                   |
 | [`dropLastWhile`](https://remedajs.com/docs/#dropLastWhile)         | [`dropRightWhile`](https://lodash.com/docs/4.17.15#dropRightWhile)       | [`dropLastWhile`](https://ramdajs.com/docs/#dropLastWhile)         |
 | [`dropWhile`](https://remedajs.com/docs/#dropWhile)                 | [`dropWhile`](https://lodash.com/docs/4.17.15#dropWhile)                 | [`dropWhile`](https://ramdajs.com/docs/#dropWhile)                 |
-| [`entries`](https://remedajs.com/docs/#entries)                     | [`toPairs`](https://lodash.com/docs/4.17.15#toPairs)                     | [`toPairs`](https://ramdajs.com/docs/#toPairs)                     |
+| [`entries`](https://remedajs.com/docs/#entries)                     |
+| [`toPairs`](https://lodash.com/docs/4.17.15#toPairs)                | [`toPairs`](https://ramdajs.com/docs/#toPairs)                           |
```
cjquines, line 31: **todo:** formatting here is now incorrect

cjquines (approved):
lgtm, will merge after todo

src/every.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    const input: Array<number | string> = [1, 2];
+    const data = every(input, isNumber);
+    expect(data).toBe(true);
+    if (every(input, isNumber)) {
+      expectTypeOf(input).toEqualTypeOf<Array<number>>();
+    }
+  });
+});
+
+describe("data-last", () => {
```
cjquines, line 33: **suggestion:** rewrite these tests to use `pipe`

i think we mostly care about data-last usages in pipes, so i'd push for these tests to be in pipes

src/every.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+describe("indexed", () => {
+  test("works correctly", () => {
+    const input = [1, 2, 3];
+    const data = every.indexed(input, (_, index) => index > 5);
+    expect(data).toBe(false);
+  });
+});
+
+describe("in pipe", () => {
```
cjquines, line 69: **question:** what does this test?

there's no inference being tested here

cjquines (commented):
scratch that i actually have some more questions

src/_types.ts
```diff
@@ -12,6 +12,20 @@ export type PredIndexedOptional<T, K> = (
...
+  input: T,
+  index: number,
+  array: ReadonlyArray<T>,
+) => input is S;
+
+export type TypePredIndexedOptional<T, S extends T> = (
+  input: T,
+  index?: number,
+  array?: ReadonlyArray<T>,
+) => input is S;
```
eranhirsch, line 27: Don't define these here, all of these types are going away in v2. They aren't useful because the hide a very basic core concept of typing

src/every.ts
```diff
@@ -0,0 +1,94 @@
...
+  export function indexed<T, S extends T>(
+    predicate: TypePredIndexed<T, S>,
+  ): (data: ReadonlyArray<T>) => data is Array<S>;
+  export function indexed<T>(
+    predicate: PredIndexed<T, boolean>,
+  ): (data: ReadonlyArray<T>) => boolean;
+  export function indexed(): unknown {
+    return purry(_every(true), arguments);
+  }
+}
```
eranhirsch, line 94: Also the indexed version is not needed, make the regular version support the indexed version.

eranhirsch (changes requested):

src/every.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    const data = every(() => false)(input);
+    expect(data).toBe(true);
+  });
+
+  test("narrows types when a type predicate is passed", () => {
+    const input: Array<number | string> = [1, 2];
+    const data = every(isNumber)(input);
+    expect(data).toBe(true);
+    if (every(isNumber)(input)) {
+      expectTypeOf(input).toEqualTypeOf<Array<number>>();
```
eranhirsch, line 56: don't mix type testing with runtime testing; they run at different times in the test cycle and thus bind two unrelated concerns.

That's why I prefer having a "runtime" block and a "typing" block.

src/every.ts
```diff
@@ -0,0 +1,94 @@
...
+ * @example
+ *    // using `R.every` as a type predicate within `conditional`
+ *    const allNumbers = R.every(R.isNumber);
+ *    const result = R.pipe(
+ *      R.conditional(
+ *        [allNumbers, val => {...}]
+ *                  // ^? number[]
+ *        ...
+ *      )
+ *    )
```
eranhirsch, line 53: I don't know if we support multiple example tags in our docs. This seems like a really big example using a utility that people might not know or understand; it makes the example more complex than it should be.

Our examples aren't usually full-blown usage examples; they just show a basic, almost trivial, use case so that people understand what the params look like and how they relate to the output.

src/every.ts
```diff
@@ -0,0 +1,94 @@
...
+): (data: ReadonlyArray<T>) => data is Array<S>;
+export function every<T>(
+  predicate: (input: T) => boolean,
+): (data: ReadonlyArray<T>) => boolean;
+
+export function every(): unknown {
+  return purry(_every(false), arguments);
+}
+
+const _every =
```
eranhirsch, line 69: In v2 we are dropping these underscore marker for private implementations. Use a full name for the function, everyImplementation

src/_types.ts
Brodzko: Where would be a better place to put them then, while in v1?

src/every.ts
Brodzko: Removed 👍

Brodzko:
Sorry, got caught up at work. I'll try to get to the other PRs during the weekend too. 

Tried addressing most of you guys' comments and posted some questions to others :)

mapping.md
Brodzko: Not sure if I see what's wrong here exactly?

Brodzko: Ah 🤦 ngl that took me a _while_ to finally notice. Fixed, thanks! 🦅 👁️

src/_types.ts
Brodzko: Put them inline in the file, the reasoning behind putting them in `_types` was that I immediately needed the same thing for `some` in the next PR, but let's take it step by step

src/every.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export function every(): unknown {
+  return purry(everyImplementation, arguments);
+}
+
+const everyImplementation = <T>(
+  data: ReadonlyArray<T>,
+  predicate: PredIndexed<T, boolean>,
+): boolean =>
+  data.every((element, index, array) => predicate(element, index, array));
```
eranhirsch, line 63: Check out similar functions in V2, you don't need to create the arrow function here, it's redundant for our use-case
```suggestion
  data.every(predicate);
```

src/every.ts
```diff
@@ -0,0 +1,63 @@
+import type { PredIndexed } from "./_types";
+import { purry } from "./purry";
+
+type TypePredIndexed<T, S extends T> = (
+  input: T,
+  index: number,
+  array: ReadonlyArray<T>,
+) => input is S;
```
eranhirsch, line 8: Inline this, in v2 all functions have the function operand stated explicitly, these make it harder to read the function signature to understand what it accepts.

src/every.ts
```diff
@@ -0,0 +1,63 @@
+import type { PredIndexed } from "./_types";
```
eranhirsch, line 1: Similarly, these are removed in v2, don't use them

src/every.ts
Brodzko: Did this, nut had to disable `no-array-callback-reference` for it to be legal :)

src/_types.ts
Brodzko: @eranhirsch lgty?

src/conditional.test.ts
```diff
@@ -106,4 +107,41 @@ describe("typing", () => {
...
+      conditional([
+        every(isNumber),
+        (arr) => {
+          expectTypeOf(arr).toEqualTypeOf<Array<number | string>>();
+          return true;
+        },
+      ]),
+    );
+
+    expect(resultBroken).toBe(true);
```
eranhirsch, line 130: This is a runtime check in a typing test. Do we need this? it doesn't test anything that is related to the bug

src/conditional.test.ts
```diff
@@ -106,4 +107,41 @@ describe("typing", () => {
...
+      conditional([
+        allNumbers,
+        (arr) => {
+          expectTypeOf(arr).toEqualTypeOf<Array<number>>();
+          return true;
+        },
+      ]),
+    );
+
+    expect(resultCorrect).toBe(true);
```
eranhirsch, line 145: similarly

src/every.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  describe("data-first", () => {
+    test("returns correct values on simple arrays", () => {
+      const trueInput = [1, 2, 3];
+      const trueData = every(trueInput, (val) => val < 5);
+      expect(trueData).toBe(true);
+
+      const falseInput = ["foo", "bar", "buzz"];
+      const falseData = every(falseInput, (val) => val.length > 3);
+      expect(falseData).toBe(false);
+    });
```
eranhirsch, line 15: Keep tests simple and short; it's better to have more tests than to have tests do more; that way, when a test fails, the actual failure is reported. And we can debug the specific case instead of having to comment out non-failing parts.
```suggestion
    test("all values meet the criteria", () => {
      expect(every([1, 2, 3], (val) => val < 5)).toBe(true);
    });
    
    test("a value doesn't meet  the criteria", () => {
      expect(every(["foo", "bar", "buzz"], (val) => val.length > 3)).toBe(false);
    });
```

src/every.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    // empty set check
```
eranhirsch, line 17: The test name describes it, we don't need documentation inside tests
```suggestion
```

src/every.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    test("returns true on empty arrays regardless of predicate", () => {
+      const input = [] as const;
+      const data = every(input, () => false);
+      expect(data).toBe(true);
+    });
+
+    test("works with indexed predicate", () => {
+      const input = [1, 2, 3];
+      const data = every(input, (_, index) => index < 5);
+      expect(data).toBe(true);
```
eranhirsch, line 27: 👍🏼 nice, this is an advantage of moving indexed predicates to the main signature, we test them like regular logic and not disconnected from other concerns!

src/every.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    });
+  });
+
+  describe("data-last", () => {
+    test("returns correct values on simple arrays", () => {
+      const trueData = pipe(
+        [1, 2, 3],
+        every((val: number) => val < 5),
```
eranhirsch, line 35: Why do you need to state `number` explicitly? Are the types not working correctly without it?
```suggestion
        every((val) => val < 5),
```

src/every.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      );
+      expect(trueData).toBe(true);
+
+      const falseData = pipe(
+        ["foo", "bar", "buzz"],
+        every((val: string) => val.length > 3),
+      );
+      expect(falseData).toBe(false);
```
eranhirsch, line 43: Same concern as the above, split tests

src/every.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+      const falseInput = ["foo", "bar", "buzz"];
+      const falseData = every(falseInput, (val) => val.length > 3);
+      expect(falseData).toBe(false);
+    });
+
+    // empty set check
+    test("returns true on empty arrays regardless of predicate", () => {
+      const input = [] as const;
+      const data = every(input, () => false);
```
eranhirsch, line 20: You can use `constant(false)`

src/every.test.ts
```diff
@@ -0,0 +1,107 @@
...
+        every((_, index) => index < 5),
+      );
+      expect(data).toBe(true);
+    });
+  });
+});
+
+describe("typing", () => {
+  test("narrows types when a type predicate is passed", () => {
+    const input: Array<number | string> = [1, 2];
```
eranhirsch, line 68: prefer casting to explicit typing, i've seen cases where they don't act the same
```suggestion
    const input = [1, 2] as Array<number | string>;
```

src/every.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      input,
+      every((val, index, arr) => {
+        expectTypeOf(val).toEqualTypeOf<number>();
+        expectTypeOf(index).toEqualTypeOf<number>();
+        expectTypeOf(arr).toEqualTypeOf<ReadonlyArray<number>>();
+        return index < 5;
+      }),
+    );
+    expect(data).toBe(true);
+    expect(dataPipe).toBe(true);
```
eranhirsch, line 93: remove runtime concerns from type tests.
```suggestion
```

src/every.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const dataPipe = pipe(
+      input,
+      every((val, index, arr) => {
+        expectTypeOf(val).toEqualTypeOf<number>();
+        expectTypeOf(index).toEqualTypeOf<number>();
+        expectTypeOf(arr).toEqualTypeOf<ReadonlyArray<number>>();
+        return index < 5;
+      }),
+    );
```
eranhirsch, line 91: There's actually a lot of cases where the piped dataLast version breaking without the dataFirst version breaking, let's split these tests

src/every.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      expectTypeOf(input).toEqualTypeOf<Array<number>>();
+    }
+  });
+
+  test("works with indexed predicate", () => {
+    const input = [1, 2, 3];
+    const data = every(input, (val, index, arr) => {
+      expectTypeOf(val).toEqualTypeOf<number>();
+      expectTypeOf(index).toEqualTypeOf<number>();
+      expectTypeOf(arr).toEqualTypeOf<ReadonlyArray<number>>();
```
eranhirsch, line 79: We don't need to test these, it's part of the function signature, and they aren't inferred or have any "magic" that might break them.
```suggestion
```

src/every.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  test("infers types when used in pipe", () => {
+    const input = [1, 2, 3] as Array<number | string>;
+    const data = pipe(
+      input,
+      every((val) => {
+        expectTypeOf(val).toEqualTypeOf<number | string>();
+        return true;
+      }),
+    );
+    expect(data).toBe(true);
```
eranhirsch, line 105: runtime concern in a type test

src/every.ts
```diff
@@ -0,0 +1,56 @@
...
+ *    R.every([1, 2, 3], R.isNumber) // => true
+ *    R.every(['a', 0, 'b'], R.isString) // => false
+ * @dataFirst
+ * @indexed
+ * @category Array
+ */
+export function every<T, S extends T>(
+  data: ReadonlyArray<T>,
+  predicate: (input: T, index: number, array: ReadonlyArray<T>) => input is S,
+): data is Array<S>;
```
eranhirsch, line 21: The typing does not respect the shape of the input in the output. We can take the tuple as a generic parameter and use it to return a result that is the same shape.

```suggestion
export function every<T extends IterableContainer, S extends T[number]>(
  data: T,
  predicate: (input: T[number], index: number, array: T) => input is S,
): data is { [K in keyof T]: S };
```

This would allow someone to test something like `[string | number, ...Array<string | number>]` and get back `[string, ...Array<string>]`

The non-narrowing overload doesn't need this treatment because it returns a `boolean`

src/every.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @example
+ *    R.every(isNumber)([1, 2, 3]) // => true
+ *    R.every(isString)(['a', 0, 'b']) // => false
+ * @dataLast
+ * @indexed
+ * @category Array
+ */
+export function every<T, S extends T>(
+  predicate: (input: T, index: number, array: ReadonlyArray<T>) => input is S,
+): (data: ReadonlyArray<T>) => data is Array<S>;
```
eranhirsch, line 43: Same as above

```suggestion
export function every<T extends IterableContainer, S extends T[number]>(
  predicate: (input: T[number], index: number, array: T) => input is S,
): (data: T) => data is { [K in keyof T]: S };
```

eranhirsch (changes requested):
Sorry for the comments so late in the development. I've been working mainly on V2 so didn't have time to check other stuff. Because the release of v2 is happening very soon and it's likely we'll merge these changes only after releasing v2 it needs to match the style for v2.

eranhirsch:
It should happen in the next 2 weeks

eranhirsch:

cjquines:
hey @Brodzko , are you planning to reopen this? if not i might pick it up

PR #636, opened 2024-04-13
feat(difference, intersection): add multiset variants

eranhirsch:
Adding a variant difference for multiset semantics and update the deprecation warning.

### DIFF
--- a/src/difference.ts
+++ b/src/difference.ts
@@ -5,7 +5,7 @@ import { purry } from "./purry";
 /**
  * Excludes the values from `other` array.
  *
- * ! **DEPRECATED**: Use `R.filter(array, R.isNot(R.isIncludedIn(other)))`. Will be removed in v2!
+ * ! **DEPRECATED**: The runtime implementation of this function handles duplicate values inconsistently. In v2 a new implementation would replace it; this implementation is accessible to v1 users via the `multiset` variant of this function (`R.difference.multiset`), or to maintain the same runtime implementation use `R.filter(array, R.isNot(R.isIncludedIn(other)))` instead.
  *
  * @param array - The source array.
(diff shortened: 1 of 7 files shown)


src/difference.ts
```diff
@@ -5,7 +5,7 @@ import { purry } from "./purry";
 /**
  * Excludes the values from `other` array.
  *
- * ! **DEPRECATED**: Use `R.filter(array, R.isNot(R.isIncludedIn(other)))`. Will be removed in v2!
+ * ! **DEPRECATED**: The runtime implementation of this function handles duplicate values inconsistently. In v2 a new implementation would replace it; this implementation is accessible to v1 users via the `multiset` variant of this function (`R.difference.multiset`), or to maintain the same runtime implementation use `R.filter(array, R.isNot(R.isIncludedIn(other)))` instead.
```
cjquines, line 8: **suggestion:**

```suggestion
 * ! **DEPRECATED**: In v2, `R.difference.multiset` will replace `R.difference`. To keep current behavior, use `R.filter(array, R.isNot(R.isIncludedIn(other)))` instead.
```

cjquines (commented):
let's include an example with duplicate values to illustrate the difference?

cjquines (approved):

cjquines:
hooray 🎉 

does this make #368 no longer needed? are multiset semantics going to be our default moving-forward, and does that mean we want the other multiset helpers, or what?

eranhirsch:
> does this make https://github.com/remeda/remeda/pull/368 no longer needed?

Other utilities we don't offer or don't offer correctly would need to be ported, too. For now, it's a good template for me to copy from and a reminder to do it.

Issue #626, opened 2024-04-09
[Feature Request]: Support Maps in isDeepEqual


BuuBux:
Deprecated `equals` seems to have worked correctly with maps, I guess isDeepEqual as a successor should also support maps, for me it would be a really cool feature and essential.

cjquines:
i agree it should—do you want to send in a PR?

BuuBux:
I'll try to do my best but no promises

eranhirsch:
btw you mention that `equals` worked with Maps, but I don't think it ever did. It would simply return `true` for any 2 maps provided to it: https://codesandbox.io/p/sandbox/empty-cache-3tnrzs

BuuBux:
👏🏻 🎉

BuuBux:
> btw you mention that `equals` worked with Maps, but I don't think it ever did. It would simply return `true` for any 2 maps provided to it: https://codesandbox.io/p/sandbox/empty-cache-3tnrzs

You were right, i was doing equal for both `MapA.get('a')` and `MapA` and if we select 1 element from Map it was working when we compare Map it was false positive

PR #637, opened 2024-04-13
feat(isDeepEqual): add support for maps 

clemgbld:
Add support for maps for the isDeepEqual function

Fixes: #626

### DIFF
--- a/src/isDeepEqual.test.ts
+++ b/src/isDeepEqual.test.ts
@@ -155,6 +155,64 @@ describe("arrays", () => {
   });
 });
 
+describe("Maps", () => {
+  test("two empty Maps should be equal", () => {
+    expect(isDeepEqual(new Map(), new Map())).toBe(true);
+  });
+  it("two Maps with different size should not be equal", () => {
+    expect(isDeepEqual(new Map(), new Map([["a", 1]]))).toBe(false);
+  });
+
+  it("two Maps with different keys shoud not be equal", () => {
+    expect(
+      isDeepEqual(
+        new Map([
+          ["a", 1],
+          ["c", 2],
+        ]),
+        new Map([
+          ["a", 1],
+          ["b", 2],
+        ]),
+      ),
+    ).toBe(false);
+  });
+
+  it("two maps with the same keys but with different values should not be equal", () => {
+    expect(
+      isDeepEqual(
+        new Map([
+          ["a", 1],
+          ["b", 3],
+          ["c", 2],
+        ]),
+        new Map([
+          ["a", 1],
+          ["b", 2],
+          ["c", 2],
+        ]),
+      ),
+    ).toBe(false);
+  });
+
+  it("two Maps with the same non primitives data should be equal", () => {
+    expect(
+      isDeepEqual(
+        new Map([
+          ["a", { a: [1, 2, 3] }],
+          ["b", { a: [3] }],
+          ["c", { b: [4] }],
+        ]),
+        new Map([
+          ["a", { a: [1, 2, 3] }],
+          ["b", { a: [3] }],
+          ["c", { b: [4] }],
+        ]),
+      ),
+    ).toBe(true);
+  });
+});
+
 describe("Date objects", () => {
   test("equal date objects", () => {
     expect(
--- a/src/isDeepEqual.test.ts
+++ b/src/isDeepEqual.test.ts
@@ -159,6 +159,7 @@ describe("Maps", () => {
   test("two empty Maps should be equal", () => {
     expect(isDeepEqual(new Map(), new Map())).toBe(true);
   });
+
   it("two Maps with different size should not be equal", () => {
     expect(isDeepEqual(new Map(), new Map([["a", 1]]))).toBe(false);
   });
--- a/src/isDeepEqual.test.ts
+++ b/src/isDeepEqual.test.ts
@@ -166,6 +166,7 @@ describe("Maps", () => {
   it("two Maps with different size should not be equal", () => {
     expect(isDeepEqual(new Map(), new Map([["a", 1]]))).toBe(false);
   });
+
   it("two Maps with different keys shoud not be equal", () => {
     expect(
       isDeepEqual(
--- a/src/isDeepEqual.test.ts
+++ b/src/isDeepEqual.test.ts
@@ -156,6 +156,9 @@ describe("arrays", () => {
 });
 
 describe("Maps", () => {
+  it("works on shallow equal maps", () => {
+    expect(isDeepEqual(new Map([["a", 1]]), new Map([["a", 1]]))).toBe(true);
+  });
   test("two empty Maps should be equal", () => {
     expect(isDeepEqual(new Map(), new Map())).toBe(true);
   });
@@ -163,7 +166,6 @@ describe("Maps", () => {
   it("two Maps with different size should not be equal", () => {
     expect(isDeepEqual(new Map(), new Map([["a", 1]]))).toBe(false);
   });
-
   it("two Maps with different keys shoud not be equal", () => {
     expect(
       isDeepEqual(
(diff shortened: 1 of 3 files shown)


src/isDeepEqual.test.ts
cjquines, line 1: **question:** how does this work in the test case

```ts
const a = new Map();
const b = new Map();
a.set(a, a);
b.set(a, a);
expect(isDeepEqual(a, b)).toBe(...);
```

src/isDeepEqual.ts
```diff
@@ -121,6 +121,31 @@ function isDeepEqualImplementation<T, S>(data: S | T, other: S): data is S {
...
   }
 
+  if (data instanceof Map) {
+    if (data.size !== (other as unknown as Map<unknown, unknown>).size) {
+      return false;
+    }
+
+    const keys = Array.from(data.keys());
+
+    for (const key of keys) {
```
cjquines, line 131: **question:** is there a reason to not directly iterate over `data`?

```ts
for (const [key, value] of data) {
  // ...
}
```

src/isDeepEqual.test.ts
cjquines: **question:** what about

```ts
const a = new Map([[1, 2]]);
const b = {
  has: () => true,
  get: () => 2,
};
expect(isDeepEqual(a, b)).toBe(false);
```

eranhirsch: For 1, we don't support cyclic structures explicitly, we don't need to add tests for this.
For 2, we fail when prototypes don't match (https://github.com/remeda/remeda/blob/master/src/isDeepEqual.ts#L86) so this case doesn't matter.

src/isDeepEqual.ts
clemgbld: I wanted to do that initially but i got a ts error telling me that we should up the target to ES6 in the tsConfig in order to do that, should i do it ?

eranhirsch: no, we'll bump it in v2. But you don't need to use `Array.from` here...

clemgbld: I can't iterate directly over data.keys() either because i get the same kind of ts error
What do you suggest, using  Object.keys or doing a for loop instead of using Array.from ?

eranhirsch: I pulled this PR, i can see that `Map.prototype.keys` returns an Iterator which needs to be spread in ES5 so your solution is ideal for ES5. Can you please add a TODO above the line saying that this should be fixed in v2. I'll follow up when I merge this PR into v2.

src/isDeepEqual.ts
```diff
@@ -121,6 +121,31 @@ function isDeepEqualImplementation<T, S>(data: S | T, other: S): data is S {
...
+      }
+
+      if (
+        !isDeepEqualImplementation(
+          data.get(key),
+          (other as unknown as Map<unknown, unknown>).get(key),
+        )
+      ) {
+        return false;
+      }
```
eranhirsch, line 143: This is a big implementation, can you pull this logic to a helper function, it would also allow us to cast "other" to a Map<unknown, unknown> once and not have to deal with the type issue on every line.

src/isDeepEqual.ts
```diff
@@ -7,7 +7,7 @@ import { purry } from "./purry";
  * objects all props will be compared recursively. The built-in Date and RegExp
  * are special-cased and will be compared by their values.
  *
- * !IMPORTANT: Maps, Sets and TypedArrays, and symbol properties of objects  are
+ * !IMPORTANT: Sets and TypedArrays, and symbol properties of objects  are
```
eranhirsch, line 10: Phrasing is now a broken
```suggestion
 * !IMPORTANT: Sets, TypedArrays, and symbol properties of objects  are
```

src/isDeepEqual.test.ts
```diff
@@ -155,6 +155,64 @@ describe("arrays", () => {
   });
 });
 
+describe("Maps", () => {
+  test("two empty Maps should be equal", () => {
+    expect(isDeepEqual(new Map(), new Map())).toBe(true);
+  });
+  it("two Maps with different size should not be equal", () => {
```
eranhirsch, line 162: nit: new line
```suggestion

  it("two Maps with different size should not be equal", () => {
```

src/isDeepEqual.test.ts
```diff
@@ -155,6 +155,64 @@ describe("arrays", () => {
   });
 });
 
+describe("Maps", () => {
```
eranhirsch, line 158: There's no test case that tests that two simple maps that are equal come up equal, e.g.
```ts
describe("Maps", () => {
  it('works on shallow equal maps', () => {
    expect(isDeepEqual(new Map([['a', 1]]), new Map([['a', 1]])).toBe(true);
  });
```

eranhirsch (approved):
This looks great, thanks for your contribution. I only have a few nits

src/isDeepEqual.ts
```diff
@@ -64,6 +64,29 @@ export function isDeepEqual(): unknown {
   return purry(isDeepEqualImplementation, arguments);
 }
 
+function areMapsEqual<T extends Map<unknown, unknown>>(
+  data: T,
+  other: T,
+): boolean {
```
eranhirsch, line 70: 1. Name can be consistent with the name of the main function, `isMapsDeepEqual` or `isDeepEqualMaps`. Although the english is broken, it's the same "brokenness" of isDeepEqual (because it can also be named areDeepEqual)... in general, functions returning boolean should be named isXXX.
2. You don't need the generic here, it doesn't add anything to the typing of the function.
```suggestion
function isDeepEqualMaps(
  data: Map<unknown, unknown>,
  other: Map<unknown, unknown>,
): boolean {
```

src/isDeepEqual.test.ts
```diff
@@ -155,6 +155,66 @@ describe("arrays", () => {
...
+  it("works on shallow equal maps", () => {
+    expect(isDeepEqual(new Map([["a", 1]]), new Map([["a", 1]]))).toBe(true);
+  });
+  test("two empty Maps should be equal", () => {
```
eranhirsch, line 162: why did you remove all the newlines?
```suggestion

  test("two empty Maps should be equal", () => {
```

src/isDeepEqual.test.ts
```diff
@@ -155,6 +155,66 @@ describe("arrays", () => {
...
+    expect(isDeepEqual(new Map(), new Map())).toBe(true);
+  });
+
+  it("two Maps with different size should not be equal", () => {
+    expect(isDeepEqual(new Map(), new Map([["a", 1]]))).toBe(false);
+  });
+  it("two Maps with different keys shoud not be equal", () => {
```
eranhirsch, line 169:
```suggestion

  it("two Maps with different keys shoud not be equal", () => {
```

src/isDeepEqual.test.ts
```diff
@@ -155,6 +155,66 @@ describe("arrays", () => {
...
+          ["c", 2],
+        ]),
+        new Map([
+          ["a", 1],
+          ["b", 2],
+        ]),
+      ),
+    ).toBe(false);
+  });
+  it("two maps with the same keys but with different values should not be equal", () => {
```
eranhirsch, line 183:
```suggestion

  it("two maps with the same keys but with different values should not be equal", () => {
```

eranhirsch (approved):

src/isDeepEqual.ts
```diff
@@ -64,6 +64,29 @@ export function isDeepEqual(): unknown {
   return purry(isDeepEqualImplementation, arguments);
 }
 
+function isDeepEqualMaps(
+  data: Readonly<Map<unknown, unknown>>,
+  other: Readonly<Map<unknown, unknown>>,
```
eranhirsch, line 69: This doesn't make the map readonly, maps have their own readonly type like arrays do:

```suggestion
  data: ReadonlyMap<unknown, unknown>,
  other: ReadonlyMap<unknown, unknown>,
```

PR #638, opened 2024-04-14
docs(card): fix card header overflow

the-main-thing:
<details><summary>Added media query for card header to fix content overflow</summary>
While reading docs on my iPhone I've noticed that the menu at the top of the page is shorter than the page itself.
This happens due to overflow of `<CardHeader />`. When there is 3 tags and the github badge then whole row goes wider then the screen.

So I added media query for screens that is wider than `sm` breakpoint.

Screenshot of the bug
Screenshot with the fix applied

</details>


eranhirsch:
Thanks! looks great!

PR #640, opened 2024-04-14
feat(isDeepEqual): add supports for Set

clemgbld:
add supports for Set

### DIFF
--- a/src/isDeepEqual.test.ts
+++ b/src/isDeepEqual.test.ts
@@ -159,6 +159,14 @@ describe("Sets", () => {
       ),
     ).toBe(true);
   });
+  test("two sets with more than two items that are all equal should be equal", () => {
+    expect(
+      isDeepEqual(
+        new Set([{ a: 1 }, { b: 3 }, { c: 4 }, { a: 1 }]),
+        new Set([{ b: 3 }, { c: 4 }, { a: 1 }, { a: 1 }]),
+      ),
+    ).toBe(true);
+  });
 });
 
 describe("arrays", () => {
--- a/src/isDeepEqual.test.ts
+++ b/src/isDeepEqual.test.ts
@@ -127,6 +127,37 @@ describe("objects", () => {
   });
 });
 
+describe("Sets", () => {
+  test("two empty sets should be equal", () => {
+    expect(isDeepEqual(new Set(), new Set())).toBe(true);
+  });
+  test("two sets of different size should not be equal", () => {
+    expect(isDeepEqual(new Set([1]), new Set([1, 2]))).toBe(false);
+  });
+  test("two sets with different primitive values shoud not be equal", () => {
+    expect(isDeepEqual(new Set([1, 2]), new Set([1, 3]))).toBe(false);
+  });
+  test("two sets with the same primitive values but out of order should be equal", () => {
+    expect(isDeepEqual(new Set([3, 1]), new Set([1, 3]))).toBe(true);
+  });
+  test("two sets with the same non primitve values out of order should be equal", () => {
+    expect(
+      isDeepEqual(
+        new Set([
+          [{ a: 1 }, { b: 2 }],
+          [{ a: 1 }, { b: 2 }],
+          [{ c: 1 }, { d: 2 }],
+        ]),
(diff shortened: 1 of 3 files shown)


eranhirsch (changes requested):
This is a really cool technique, but we can't rely on the JSON representation to sort the set items because not everything is JSON serializable: Maps, Sets, RegExp, undefined, bigints (they even throw!), Classes, and objects with non-enumerable properties (like symbols), etc...

Although the conversion to a sorted array is a really cool optimization, without a globally correct sorting criteria we will need a different approach: most likely to compare items one by one to the other sets items and "cross off" equalities until we finish iterating over the whole set, this means we will do O(n^2) deep equality comparisons in the worst case, but I think that's the best we can offer so it's fine (as long as we mention it in the docs).

src/isDeepEqual.ts
```diff
@@ -174,3 +178,43 @@ function isDeepEqualMaps(
...
+  other: ReadonlySet<unknown>,
+): boolean {
+  if (data.size !== other.size) {
+    return false;
+  }
+
+  // TODO: Once we bump our typescript target we can iterate over the map keys and values directly.
+  const dataArr = Array.from(data.values());
+  const otherMap = arrayToMap(Array.from(other.values()));
+  const otherKeys = Array.from(otherMap.keys());
```
eranhirsch, line 193: These structures are overkill. Create an array of the items of the other array like you created for the data (e.g. `Array.from(other.values())` and use splice to remove items from it every time you find a matching element.

src/isDeepEqual.ts
```diff
@@ -174,3 +178,43 @@ function isDeepEqualMaps(
...
+      if (isDeepEqualImplementation(dataItem, otherMap.get(otherKey))) {
+        isFound = true;
+        otherMap.delete(otherKey);
+        break;
+      }
+    }
+    if (!isFound) {
+      return false;
+    }
+    isFound = false;
```
eranhirsch, line 208: scope isFound to the outer loop, it's only used within it's scope, and that way you don't need to reset it to `false`.
```suggestion
  for (const dataItem of dataArr) {
    let isFound = false;

    for (const otherKey of otherKeys) {
      if (isDeepEqualImplementation(dataItem, otherMap.get(otherKey))) {
        isFound = true;
        otherMap.delete(otherKey);
        break;
      }
    }
    if (!isFound) {
      return false;
    }
```

eranhirsch (changes requested):
Thanks for fixing the issue! This implementation creates redundant data structures which might affect performance for really large sets. I've added comments inline.

src/isDeepEqual.ts
```diff
@@ -174,3 +178,33 @@ function isDeepEqualMaps(
...
+  const dataArr = Array.from(data.values());
+  const otherArr = Array.from(other.values());
+
+  for (const dataItem of dataArr) {
+    let isFound = false;
+
+    for (let i = 0; i < otherArr.length; i++) {
+      if (isDeepEqualImplementation(dataItem, otherArr[i])) {
+        isFound = true;
+        otherArr.splice(i, i);
```
eranhirsch, line 200: This is incorrect, the second operand for slice is the number of items to remove, it should be 
```suggestion
        otherArr.splice(i, 1);
```

This also means the test plan doesn't cover cases where this would have broken. Please also add a test that tests sets with more than 2 items to make sure the code works as intended.

eranhirsch (changes requested):

src/isDeepEqual.test.ts
```diff
@@ -128,6 +128,39 @@ describe("objects", () => {
...
+    ).toBe(false);
+  });
+  test("two sets of Maps with the same values should be equal", () => {
+    expect(
+      isDeepEqual(
+        new Set([new Map([["a", 123]]), new Map([["b", 456]])]),
+        new Set([new Map([["b", 456]]), new Map([["a", 123]])]),
+      ),
+    ).toBe(true);
+  });
```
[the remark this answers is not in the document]
eranhirsch: I see you added a test for this after I wrote the comment

PR #642, opened 2024-04-17
fix(isDeepEqual): wrong dataLast types

eranhirsch:
There was a bug with the typing in the dataLast version that slipped through our testing. It prevented headless usage as an equality checker for functions like differenceWith.


eranhirsch:
@cjquines - I shipped this so I can merge it into `beta` because I want to continue working on v2. Feel free to review it retrospectively and surface any issues you detect.

PR #644, opened 2024-04-17
docs(v2): Migration Guide, A - E

eranhirsch:
Migration Guide for each function


cjquines (approved):
hm, is the philosophy behind this "release notes" or "migration guide"? because some of these changes i wouldn't put in a migration guide, e.g. the notes on `constant`

eranhirsch:
Feel free to comment on specifics. I don't know what approach to take here.

The goal is for people to update the npm dependency to v2, then have these as lookups when they either go over all their imports and see what might cause breakages in their code or effectively debug compile and runtime issues they observe after upgrading (there will be a bunch because of the removed `strict` and `indexed` variants, for example.

Each article should be self-contained and accompany the docs for the previous and new versions of that function. It shouldn't assume any "global" information about the v2 migration or about upgrading to it. I don't want to create a general-purpose migration guide that people would have to reference. It doesn't make sense for Remeda, which doesn't have a unifying component.

WDYT?

cjquines:
that makes sense then—i suppose i can see several places where stuff like new types could cause issues

eranhirsch:
@cjquines - Do you have any concrete suggestions for the content in this PR?

eranhirsch:
@cjquines, I've decided to simplify the notes, i've renamed the block itself "Breaking Changes in V2" and removed any notes that didn't talk specifically about potential breakages (like the constant and debounce notes).

PR #647, opened 2024-04-20
fix(`take`): Improve `take` typings for tuples

Brodzko:
Continuing in the spirit of https://github.com/remeda/remeda/pull/623, here are proper types for readonly tuples for `last`.

I _did try_ copying over `ArraySlice` from `type-fest` but
- it added _a lot_ of dependent types and
- it immediately crashed on infinite type instantiation anyway

I think this type should be safe for all cases of the existing implementation and much more performant.

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/take.test.ts
+++ b/src/take.test.ts
@@ -4,6 +4,14 @@ describe("data_first", () => {
   it("take", () => {
     expect(take([1, 2, 3, 4, 3, 2, 1] as const, 3)).toEqual([1, 2, 3]);
   });
+
+  it("returns the whole array if N is greater than length", () => {
+    expect(take([1, 2, 3] as const, 10)).toEqual([1, 2, 3]);
+  });
+
+  it("returns an empty array if N is negative", () => {
+    expect(take([1, 2, 3] as const, -1)).toEqual([]);
+  });
 });
 
 describe("data_last", () => {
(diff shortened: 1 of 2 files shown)


eranhirsch (changes requested):
In the past, these sorts of changes were done with a `strict` variant because changing the return type of an existing function would be considered a breaking change requiring us to bump a major version. Luckily, we are just about to bump a major version. Remeda v2 also has type-fest as a dependency so you can use types from type-fest directly.

I suggest porting this PR to the `beta` branch.

Also, please make sure you test as much as possible, with as many cases as you can think of so we are 100% sure we aren't returning broken types. The `strict` variant in the past allowed us to provide the better typed versions as a sort of "experimental" feature and get feedback while users could easily switch back to the non-strict versions if we made a mistake. With v2 we won't have that to protect us and it will block them from migrating to v2.

Some cases to test:
* strict tuples: [string, number boolean]
* literal unions: ["cat" | "dog", "mouse" | "chicken"]
* rest tail: [string, ...number[]]
* rest head: [boolean, ...string[]]
* mix of rest params and literal unions: ["cat" | "dog", ...(123 | 456)[]]
* Tuple is shorter than N: N == 3, [string]
* Tuple has a prefix shorter than N: N == 3, [string, ...number[]]
* Tuple has prefix longer than N: N == 2, [string, boolean, number, ...string[]]

And all these should be tested both with `readonly` and without.

Brodzko:
Trying to use type-fest ended up in excessively deep type instantiating - do you think I can keep a custom type like this? 

Good point about the test cases, will add soon

Brodzko:
Closing in favor of https://github.com/remeda/remeda/pull/649

PR #649, opened 2024-04-20
fix(v2): Improve `take` typings for tuples

Brodzko:
Based on https://github.com/remeda/remeda/pull/647#pullrequestreview-2013028424, porting that MR into v2 and improved typing tests.

Please do let me know if there is anything else needed to merging into `beta` - I noticed pre-commit hooks failing due to a bunch of TS errors that are unrelated to my changes, so I assume that is expected for now?

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/take.test.ts
+++ b/src/take.test.ts
@@ -26,24 +26,137 @@ describe("data_last", () => {
 });
 
 describe("typings", () => {
-  it("infers tuple types properly", () => {
-    const valid = take([1, 2, "foo", "bar"] as const, 3);
-    expectTypeOf(valid).toEqualTypeOf<[1, 2, "foo"]>();
+  it("handles strict tuples", () => {
+    const strict = ["1", 2, true] as [string, number, boolean];
+    expectTypeOf(take(strict, 2)).toEqualTypeOf<[string, number]>();
 
-    const negative = take([1, 2, 3] as const, -1);
-    expectTypeOf(negative).toEqualTypeOf<[]>();
+    const strictReadonly = ["1", 2, true] as readonly [string, number, boolean];
+    expectTypeOf(take(strictReadonly, 2)).toEqualTypeOf<[string, number]>();
+  });
 
-    const greater = take([1, 2, 3] as const, 10);
-    expectTypeOf(greater).toEqualTypeOf<[1, 2, 3]>();
+  it("handles literal unions", () => {
+    const unions = ["cat", "mouse"] as ["cat" | "dog", "mouse" | "chicken"];
+    expectTypeOf(take(unions, 1)).toEqualTypeOf<["cat" | "dog"]>();
 
-    const unknown = take([1, 2, 3], 2);
-    expectTypeOf(unknown).toEqualTypeOf<Array<number>>();
+    const unionsReadonly = ["cat", "mouse"] as readonly [
+      "cat" | "dog",
+      "mouse" | "chicken",
+    ];
+    expectTypeOf(take(unionsReadonly, 1)).toEqualTypeOf<["cat" | "dog"]>();
+  });
+
+  it("handles rest tails", () => {
+    const restTail = ["1", 2, 3] as [string, ...Array<number>];
+    expectTypeOf(take(restTail, 1)).toEqualTypeOf<[string]>();
+    expectTypeOf(take(restTail, 2)).toEqualTypeOf<[string, ...Array<number>]>();
+
+    const restTailReadonly = ["1", 2, 3] as readonly [string, ...Array<number>];
+    expectTypeOf(take(restTailReadonly, 1)).toEqualTypeOf<[string]>();
+    expectTypeOf(take(restTailReadonly, 2)).toEqualTypeOf<
+      [string, ...Array<number>]
+    >();
+  });
+
+  it("handles rest heads", () => {
+    const restHead = [1, 2, 3, "foo"] as [...Array<number>, string];
+    expectTypeOf(take(restHead, 3)).toEqualTypeOf<[...Array<number>, string]>();
+
+    const restHeadReadonly = [1, 2, 3, "foo"] as readonly [
+      ...Array<number>,
+      string,
+    ];
+    expectTypeOf(take(restHeadReadonly, 3)).toEqualTypeOf<
+      [...Array<number>, string]
+    >();
+  });
+
+  it("handles a mix of unions and rest params", () => {
+    const mix = ["cat", 123, 123, 456] as ["cat" | "dog", ...Array<123 | 456>];
+    expectTypeOf(take(mix, 1)).toEqualTypeOf<["cat" | "dog"]>();
+    expectTypeOf(take(mix, 3)).toEqualTypeOf<
+      ["cat" | "dog", ...Array<123 | 456>]
+    >();
+
+    const mixReadonly = ["cat", 123, 123, 456] as readonly [
+      "cat" | "dog",
+      ...Array<123 | 456>,
+    ];
+    expectTypeOf(take(mixReadonly, 1)).toEqualTypeOf<["cat" | "dog"]>();
+    expectTypeOf(take(mixReadonly, 3)).toEqualTypeOf<
+      ["cat" | "dog", ...Array<123 | 456>]
+    >();
+  });
+
+  it("handles N greater than tuple length", () => {
+    const short = ["foo"] as [string];
+    expectTypeOf(take(short, 10)).toEqualTypeOf<[string]>();
+
+    const shortReadonly = ["foo"] as readonly [string];
+    expectTypeOf(take(shortReadonly, 10)).toEqualTypeOf<[string]>();
+  });
+
+  it("handles shorter heads", () => {
+    const shortHead = [1, "foo", true, false] as [
+      number,
+      string,
+      ...Array<boolean>,
+    ];
+    expectTypeOf(take(shortHead, 3)).toEqualTypeOf<
+      [number, string, ...Array<boolean>]
+    >();
+
+    const shortHeadReadonly = [1, "foo", true, false] as readonly [
+      number,
+      string,
+      ...Array<boolean>,
+    ];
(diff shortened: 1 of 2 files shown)


eranhirsch:
Regarding your precommit hooks, don't forget to `npm ci` when moving between the master branch and the beta branch, the dependencies have changed considerably and you can't use the tooling without updating them first.

src/take.ts
```diff
@@ -29,14 +46,18 @@ export function take<T>(array: ReadonlyArray<T>, n: number): Array<T>;
...
 export function take(...args: ReadonlyArray<unknown>): unknown {
   return purry(takeImplementation, args, lazyImplementation);
 }
 
-const takeImplementation = <T>(array: ReadonlyArray<T>, n: number): Array<T> =>
-  _reduceLazy(array, lazyImplementation(n));
+const takeImplementation = <T extends ReadonlyArray<unknown>, N extends number>(
+  array: T,
+  n: N,
+): unknown => _reduceLazy(array, lazyImplementation(n));
```
eranhirsch, line 60: If we are changing this function with breaking changes lets also change the implementation to use Array.prototype.slice to perform the dataFirst implementation. It's funky that we use our own reducer for this.

Brodzko: Is this what you mean? 

Also, what do you think about supporting negative range, same as `slice` while we're at it? I.e. `R.take(arr, -2)` - return last 2 elements of `arr`

eranhirsch: Would the typing work?

Brodzko: Well, been looking at it and I'm not sure we can get them working the same as for taking from he head part. Especially in "headed" cases like `[number, string, ...Array<boolean>]` there really doesn't seem to be a "good" way to define the output array.

However while thinking about this I also realized the same problem exists for regular `take` in "tailed" cases and that my tests do not make sense in those cases either.

Take `const arr = [1, 2] as [...Array<number>, string]`. What should be the output type of `take(arr, n)` for any `n`? Best we could do is maybe something like `Array<number> | [...Array<number>, string]`, or `Array<number | string>`, but neither seems as a good choice?

Brodzko: Anyhow if we solve this ^, than I think for `N < 0`, `Take<T, N> = Reverse<Take<Reverse<T>, -N>>` should do the trick?

src/take.test.ts
```diff
@@ -16,3 +24,139 @@ describe("data_last", () => {
...
+      "cat" | "dog",
+      "mouse" | "chicken",
+    ];
+    expectTypeOf(take(unionsReadonly, 1)).toEqualTypeOf<["cat" | "dog"]>();
+  });
+
+  it("handles rest tails", () => {
+    const restTail = ["1", 2, 3] as [string, ...Array<number>];
+    expectTypeOf(take(restTail, 1)).toEqualTypeOf<[string]>();
+    expectTypeOf(take(restTail, 2)).toEqualTypeOf<[string, ...Array<number>]>();
```
eranhirsch, line 51: i'd expect the type of this to be `[string] | [string, number]`

src/take.test.ts
```diff
@@ -16,3 +24,139 @@ describe("data_last", () => {
...
+    const restTailReadonly = ["1", 2, 3] as readonly [string, ...Array<number>];
+    expectTypeOf(take(restTailReadonly, 1)).toEqualTypeOf<[string]>();
+    expectTypeOf(take(restTailReadonly, 2)).toEqualTypeOf<
+      [string, ...Array<number>]
+    >();
+  });
+
+  it("handles rest heads", () => {
+    const restHead = [1, 2, 3, "foo"] as [...Array<number>, string];
+    expectTypeOf(take(restHead, 3)).toEqualTypeOf<[...Array<number>, string]>();
```
eranhirsch, line 62: similarly this could be `[string] | [number, string] | [number, number, string] | [number, number, number]`

src/take.test.ts
```diff
@@ -16,3 +24,139 @@ describe("data_last", () => {
...
+      [...Array<number>, string]
+    >();
+  });
+
+  it("handles a mix of unions and rest params", () => {
+    const mix = ["cat", 123, 123, 456] as ["cat" | "dog", ...Array<123 | 456>];
+    expectTypeOf(take(mix, 1)).toEqualTypeOf<["cat" | "dog"]>();
+    expectTypeOf(take(mix, 3)).toEqualTypeOf<
+      ["cat" | "dog", ...Array<123 | 456>]
+    >();
```
eranhirsch, line 78: And similarly: `["cat" | "dog"] | ["cat" | "dog", 123 | 456] | ["cat" | "dog", 123 | 456, 123 | 456]`

src/take.test.ts
```diff
@@ -16,3 +24,139 @@ describe("data_last", () => {
...
+
+    const shortHeadReadonly = [1, "foo", true, false] as readonly [
+      number,
+      string,
+      ...Array<boolean>,
+    ];
+    expectTypeOf(take(shortHeadReadonly, 3)).toEqualTypeOf<
+      [number, string, ...Array<boolean>]
+    >();
+  });
```
eranhirsch, line 116: Is this test redundant, it feels like it's already tested as part of the tuples with rest tails...

src/take.test.ts
```diff
@@ -16,3 +24,139 @@ describe("data_last", () => {
...
+    ];
+
+    expectTypeOf(take(longHeadReadonly, 2)).toEqualTypeOf<[number, string]>();
+    expectTypeOf(take(longHeadReadonly, 6)).toEqualTypeOf<
+      [number, string, boolean, ...Array<string>]
+    >();
+  });
+
+  it("handles regular arrays", () => {
+    const input = [1, 2, 3];
```
eranhirsch, line 145: 1. cast this to an array, i'm not sure the type isn't a tuple here...
2. you don't need values for type tests, only types...
```suggestion
    const input = [] as Array<number>;
```

src/take.test.ts
```diff
@@ -16,3 +24,139 @@ describe("data_last", () => {
...
+    expectTypeOf(take(input, 10)).toEqualTypeOf<Array<number>>();
+  });
+
+  it("infers types in pipe", () => {
+    const input = [1, 2, 3, 4] as [1, 2, 3, 4];
+    const result = pipe(input, take(2));
+
+    expectTypeOf(result).toEqualTypeOf<[1, 2]>();
+
+    const inputReadonly = [1, 2, 3, 4] as readonly [1, 2, 3, 4];
```
eranhirsch, line 157:
```suggestion
    const inputReadonly = [1, 2, 3, 4] as const;
```

src/take.ts
```diff
@@ -1,7 +1,20 @@
...
+  T extends ReadonlyArray<unknown>,
+  N extends number,
+  Acc extends ReadonlyArray<unknown> = [],
+> = `${N}` extends `-${number}`
+  ? []
+  : Acc["length"] extends N
+    ? Acc
+    : T extends readonly [infer Head, ...infer Tail]
+      ? TakeAcc<Tail, N, [...Acc, Head]>
+      : [...Acc, ...T];
```
eranhirsch, line 14: I'm not sure about this branch. Can we return a narrowed type that would take N into account? See my questions in the tests.

eranhirsch (changes requested):
Typing `take` is similar to how you'd implement a `hasAtMost` type predicate. If the input is `number[]` and N=3, you'd want the output to be either `[number?, number?, number?]` or `[] | [number] | [number, number] | [number, number, number]` (preferably), and not `number[]`.

It's not the typing is incorrect right now, it's just that it's less useful than it could be, if we are already looking into narrowing the type. The most common inputs in general would probably be regular arrays `string[]`, fixed tuples like `[string, string, string]`, and maybe non-empty arrays `[boolean, ...boolean[]]`, so those are the types we should focus on providing a better experience for. We currently do the fixed tuple well, but not the other two.

Regarding handling negative N values - I think the current implementation makes the most sense, as it aligns better with what the expectation of the semantics of the function is, currently there is a correlation between the size of N and the number of items you get, so if N1 > N2 the number of elements you get back is |N1| >= |N2|, but if you support negative values, then this doesn't hold anymore: N1 = 1, N2 = -3 --> |N1| === 1, |N2| === 3.

Another issue, there are no type tests where N is a number and not a const. Assume that a lot of users use `take` with a non-const `number`. We need to make sure that our typing makes sense for them.

You can test N as a number by doing `1 as number` as an operand.

Brodzko:
> Typing `take` is similar to how you'd implement a `hasAtMost` type predicate. If the input is `number[]` and N=3, you'd want the output to be either `[number?, number?, number?]` or `[] | [number] | [number, number] | [number, number, number]` (preferably), and not `number[]`.

I see your point, but am not sure about the ergonomics of types like that. Consider e.g. for `[...Array<number>, string]` and `N=3` the resulting type would have to be

```ts
type CorrectType = [string] | [number, string] | [number, number] | [number, number, string] | [number, number, number]
```

For large `N` this would quickly become unreadable and maybe even hit some performance issues when type checking?
The fact it would necessitate following calls to `hasAtLeast` to narrow the type further can also be seen as both good and bad I suppose. 

I assume for `N = number` this would end up being like this?
```ts
type TypeForNumber = Array<number> | [...Array<number>, string]`
```

eranhirsch:
It's a question of practicality. In most cases, if you are doing `take(N)` with a **const** N you are either using a small N because you want to access specific items in the array (e.g. `arr[3]`), or a large N where you probably don't care about the type of *any* of the individual elements (maybe returning a paginated result from a server). In the former case, every bit of typing information is helpful; it can reduce the need to perform redundant runtime checks or handle overly wide typing; in the latter case, you can cast your number `as number` and "turn off" all these advanced typing if the typing is causing issues.

Typescript performance isn't a problem until we have concrete information that it is. From my experience with `hasAtLeast`, typescript can handle up to about 50 iterations of recursive types before it fails.

TL;DR - my point is, for low constant N's, the better the typing, the more useful it is, and the typing can always be turned off for larger Ns if they do create an issue, so we should strive to maximize the type utility here.

PR #651, opened 2024-04-22
docs: Migration Guide, F

eranhirsch:
There's a lot of functions starting with F which had changes

### DIFF
--- a/docs/src/content/v1-migration/find.md
+++ b/docs/src/content/v1-migration/find.md
@@ -1,0 +1,55 @@
+# Typing
+
+When using with a type-predicate, the returned item would now be narrowed to the
+type of the type-predicate, similar to how `filter` with a type-predicate
+narrows the resulting items in the array.
+
+The `indexed` variant was removed; the base implementation takes the same
+parameters. If you are using `indexed` you can simply remove it without any
+other changes.
+
+# Runtime
+
+The predicate function now takes 2 additional parameters: `index` - The index of
+the current element being processed in array, and `data` - the array the
+function was called upon (the same signature as the built-in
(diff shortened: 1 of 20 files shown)


docs/src/content/v1-migration/find.md
```diff
@@ -0,0 +1,55 @@
+# Typing
+
+When using with a type-predicate, the returned item would now be narrowed to the
+type of the type-predicate, similar to how `filter` with a type-predicate
+narrows the resulting items in the array.
+
```
cjquines, line 6: **nit:** let's make this second paragraph to match example order, and for consistency with other `indexed` changes...

docs/src/content/v1-migration/forEachObj.md
```diff
@@ -0,0 +1,70 @@
...
+  console.log(`${key} => ${value}`);
+});
+
+// Now
+forEachObj({ a: 1, b: 2 }, (value, key) => {
+  console.log(`${key} => ${value}`);
+});
+```
+
+### Symbol Keys
```
cjquines, line 40: **nit:** `Keys` -> `keys`

cjquines: (our headings are in sentence case)

cjquines: lmao this is an extremely tiny nit, i will probably just give it a copyediting pass when they're all written

eranhirsch: These become repetitive to write and are quite draining for me, I'm happy that another person is reading them and getting a feel for them so they stay consistent.

PR #652, opened 2024-04-24
docs(v2): Migration Guide, G - N

eranhirsch:


docs/src/content/v1-migration/keys.md
```diff
@@ -0,0 +1,50 @@
+# Typing
+
+If you weren't using the `strict` variant, The returned type is now stricter and
```
cjquines, line 3: **nit:** 

```suggestion
If you weren't using the `strict` variant, the returned type is now stricter and
```

docs/src/content/v1-migration/intersection.md
```diff
@@ -0,0 +1,75 @@
+# Typing
+
+The typing for the `other` operand was softened so that it could accept any
```
cjquines, line 3: **nit:** softened -> widened (which is the term ts uses in the docs)

```suggestion
The typing for the `other` operand was widened so that it could accept any
```

docs/src/content/v1-migration/intersection.md
```diff
@@ -0,0 +1,75 @@
...
+## Examples
+
+### Narrowed result
+
+```ts
+const DATA1 = [] as (string | number)[];
+const DATA2 = [] as (number | boolean)[];
+
+// In the current implementation typescript would error for DATA2 because it
+// isn't the same type as DATA1, this would no longer error.
```
cjquines, line 32: **suggestion:** for parallelism with the was/now examples

```suggestion
// Was a type error, now okay:
```

docs/src/content/v1-migration/mapKeys.md
```diff
@@ -0,0 +1,41 @@
+# Typing
+
+The mapping function is no longer typed to accept symbol keys, it casts `number`
+keys as strings, and the values of `symbol` keys are also not part of the typing
+of the value parameter. Additionally, a third parameter was added with the whole
```
cjquines, line 5: **suggestion:** let's add a

```md
This aligns with the runtime behavior of [`Object.keys`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys).
```

here too?

docs/src/content/v1-migration/groupBy.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+### Better typing
+
+```ts
+type Pet = { type: "cat" | "dog"; name: string };
+
+const grouped = groupBy(
+  //  ^? Partial<Record<"cat" | "dog", NonEmptyArray<Pet>>>
+  // Was: Record<PropertyKey, NonEmptyArray<Pet>>
```
cjquines, line 60: **thought:** should we change the order of these?

i was confused at first that the was comes after the now (which is not the case for the examples, but is the case for the types and results). maybe it'd be better to be consistent? but i'm not sure. anyway, we can change this all together if we decide it's clearer

docs/src/content/v1-migration/mapKeys.md
```diff
@@ -0,0 +1,41 @@
...
+const result = mapKeys(
+  //  ^? Partial<Record<"yes" | "no", string>>, Was: Record<"yes" | "no", string>
+  { a: "hello" } as Record<string, string>,
+  (key) => (key.length > 3 ? "yes" : "no"),
+);
+
+// We don't know in compile time if any of the entries in the input would result
+// in creating either a "yes" or "no" key and thus might not exist on the output
+// object, therefore it's a mistake to assume that the type of the prop would be
+// `string`.
```
cjquines, line 38: **suggestion:**

```suggestion
// It's possible none of the keys mapped to "yes", so the type here is
// `string | undefined`:
```

docs/src/content/v1-migration/mapKeys.md
cjquines: (and same with mapValues?)

cjquines (approved):

docs/src/content/v1-migration/groupBy.md
eranhirsch: I put the new type first in all the type examples where I use the `^?` sign and add the "was" afterward. I started doing it intuitively, but now it's everywhere. It's because I thought the expectation from the `^?` would be that it would tell you the existing type, and the "was: " is like a comment on top of it.

wdyt?

cjquines: yeah i'm not sure... i agree that it's a bit late since we already have other type examples...

i was thinking we could have something like

```ts
  //  ^?
  // Was: Record<PropertyKey, NonEmptyArray<Pet>>
  // Now: Partial<Record<"cat" | "dog", NonEmptyArray<Pet>>>
```

it takes more vertical space, but i also dk how much that's a consideration since a lot of our migration notes are already long lol

cjquines: but also if it was like an inline comment it'd be like

```ts
const result = doSomething(data); // => Was: string | undefined; Now: string
```

or

```ts
const result = doSomething(data); // => Was: string | undefined
                                  //    Now: string
```

cjquines: i don't like any of these solutions, so i'm fine with how it is now—if we change our minds later, i can take point on changing everything

eranhirsch: Let's defer this to later if we get feedback that the docs aren't good

PR #673, opened 2024-05-05
docs(v2): Restyle migration guide box

cjquines:
* make the box collapsible (this is what i'm least sure of)
* make the headings smaller
* make the code blocks wider

image


cjquines:
could also add some function tag to mark stuff that had breaking changes

docs/src/components/v1/migration-box.tsx
```diff
@@ -0,0 +1,46 @@
+import type { ReactNode } from "react";
+import { Button } from "@/components/ui/button";
+import {
+  CollapsibleContent,
+  Collapsible as CollapsibleRoot,
```
eranhirsch, line 5: Is there a reason we need this rename?
```suggestion
  Collapsible,
```

eranhirsch (approved):
I think that if we go with this approach we can drop the "No changes" blocks for functions with migration notes, and maybe change the navigation label for the v1 docs to something like "Docs (Legacy, V1)" or something else that makes sense.

docs/src/components/markdown.astro
```diff
@@ -3,14 +3,13 @@ import { asHtml } from "@/lib/as-html";
...
-const { markdown } = Astro.props;
+const { markdown, className } = Astro.props;
 ---
 
 <article
-  class="prose prose-slate dark:prose-invert prose-code:text-primary prose-code:before:content-none prose-code:after:content-none"
+  class={`prose prose-slate dark:prose-invert prose-code:text-primary prose-code:before:content-none prose-code:after:content-none ${className ?? ""}`}
   set:html={markdown === undefined ? undefined : asHtml(markdown)}
->
-  <slot />
```
eranhirsch, line 15: I'm confused, how does this work if you removed the slot? We use the slotted approach for the migration boxes...

eranhirsch (approved):
Thanks for addressing my comments.

docs/src/components/markdown.astro
cjquines: `set:html` is the thing that does this here; this was apparently against the https://ota-meshi.github.io/eslint-plugin-astro/rules/no-conflict-set-directives/ rule

eranhirsch: but doesn't this mean we need to send the markdown as a param instead of a child in the migration box?

cjquines: huh yeah, i checked and i guess they're both needed????

PR #676, opened 2024-05-10
fix(v2): rewrite clone

eranhirsch:
I'm procrastinating releasing v2 so in the meantime I took a look at clone.

I deprecated the support for `.clone()` function inside the cloned object. It wasn't part of the documentation and felt like it doesn't fit in the library.


src/clone.ts
```diff
@@ -1,68 +1,110 @@
...
-  return value != null && typeof (value as any).clone === "function"
-    ? (value as any).clone()
-    : cloneImplementation(value, [], []);
+export function clone(): <T>(data: T) => T;
+
+export function clone(...args: ReadonlyArray<unknown>): unknown {
+  return purry(cloneImplementation, args);
 }
 
 // from https://github.com/ramda/ramda/blob/master/source/internal/_clone.js
```
cjquines, line 39: **nit:** we can remove this right?

src/clone.ts
```diff
@@ -1,68 +1,110 @@
...
+  }
+
+  if (
+    typeof value !== "object" ||
+    value === null ||
+    value instanceof Date ||
+    value instanceof RegExp
+  ) {
+    // We can use the built-in deep cloning function.
+    return structuredClone(value);
```
cjquines, line 57: **thought:** for some reason i thought structuredClone had bad compat but it's fine

cjquines (approved):

src/clone.ts
eranhirsch: Our tests run on Node 18 too, so if they pass i'm going to assume that the support is adequete for our use-case.

PR #677, opened 2024-05-10
docs(v2): Refactor + Articles Infra

eranhirsch:
Adding a heading section to the v1 docs that would provide general context for the v2 migration


docs/src/components/markdown.astro
```diff
@@ -13,5 +13,8 @@ const { markdown, className } = Astro.props;
   class={`prose prose-slate dark:prose-invert prose-code:text-primary prose-code:before:content-none prose-code:after:content-none ${className ?? ""}`}
   set:html={markdown === undefined ? undefined : asHtml(markdown)}
 >
-  <slot />
+  {
+    // Only use the slot if we haven't already set the inner HTML via `set:html`
+    markdown === undefined ? <slot /> : undefined
+  }
```
cjquines, line 19: **thought:** oh this makes a lot more sense

eranhirsch:
This is a huge PR, I played around a lot with astro to try and see how we can make it the easiest to add content to the site, and re-use as much as possible now that V1 is going to be part of the site for a while.

The main things in this PR are:
* Adding the ability to render content as part of the docs pages
* Normalizing the usage of shadcn, removing any manual changes done previously in favour of simpler modifications "from the outside".
* Creating a version selector in the navbar instead of 2 entries in the top bar.

docs/src/content/docs/v1/minimum-versions.md
```diff
@@ -0,0 +1,31 @@
...
+version is supported by all current supported Node versions (18+) and by _~94%_
+of browsers.
+
+Browsers and runtimes that don't support ES2022 can still use functions that
+don't rely on anything that runtime doesn't support, but we don't test for those
+cases and won't be able to support any issues.
+
+## Typescript 5.1+
+
+## ESM
```
cjquines, line 31: **question:** will these have descriptions eventually? should we TODO this?

docs/src/components/function-api.astro
```diff
@@ -1,28 +1,38 @@
...
-          <CodeBlock code={method.example} />
-        </section>
-      )}
-    </div>
-  )
-}
+<section class="flex flex-col gap-2">
+  {
+    variant !== undefined && (
+      <h6 class="text-sm font-bold text-secondary-foreground underline">
```
cjquines, line 23: **nit:** bold [underline](https://practicaltypography.com/underlining.html) is a bit too much emphasis imo, just the bold is fine

docs/tsconfig.json
```diff
@@ -4,13 +4,14 @@
   "compilerOptions": {
     "baseUrl": ".",
     "paths": {
-      "@/*": ["src/*"]
+      "@/*": ["src/*"],
+      "@shadcn/*": ["src/components/ui/*"]
```
cjquines, line 8: **thought:** not sure how i feel about this...

i think it's nice, but it does obscure the import

if you look at [shadcn itself](https://ui.shadcn.com/docs/components/accordion#usage), it always imports from `@/components/ui`

docs/src/pages/index.astro
cjquines, line 1: **question:** is there some way to get this on markdown? i guess the lead section and the install stuff can stay here but i feel like the codeblocks can be put in markdown, but idk enough about astro

docs/src/content/docs/v1/migration-intro.md
cjquines, line 1: **praise:** nice

docs/src/content/config.ts
```diff
@@ -0,0 +1,28 @@
...
+       * the content together.
+       */
+      category: z.string().min(1),
+
+      /**
+       * The priority defines the sorting order *within* the category.
+       */
+      priority: z.number().positive().int().optional(),
+
+      isV1: z.literal(true as const).optional(),
```
cjquines, line 23: **nit:** maybe this is premature, but i'd rather this be a literal of versions rather than a boolean

docs/src/components/version-selector.tsx
cjquines, line 1: **praise:** thanks; i was gonna do something like this but you got to it first!

docs/src/components/tag-badge.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const TAG_COLOR = {
+  Lazy: "bg-green-600 text-green-50 hover:bg-green-600/80 dark:bg-green-500 dark:text-green-950",
+
+  Indexed:
+    "bg-purple-600 text-purple-50 hover:bg-purple-600/80 dark:bg-purple-500 dark:text-purple-950",
+
+  Strict:
+    "bg-rose-600 text-rose-50 hover:bg-rose-600/80 dark:bg-rose-500 dark:text-rose-950",
+} as const satisfies Readonly<Record<Tag, string>>;
```
cjquines, line 14: **thought:** this is one of the things i dislike about the way we compile tailwind... you'd think we could just do a variable with the color, but that doesn't work!

cjquines (approved):

docs/src/content/docs/v1/minimum-versions.md
eranhirsch: They will. I wanted to send out the actual content in a separate PR but didn't want to handle merge conflicts, so I put the content here until you approve this PR.

docs/tsconfig.json
eranhirsch: I agree; it was an idea I had to make it more straightforward for junior contributors to understand where these components come from if they want to change or add new ones. I feel that Shadcn does a lousy job documenting itself in a project. I can see how a project deteriorates in quality as people who don't know what Shadcn is start mucking around with the copied components.

docs/src/components/tag-badge.tsx
eranhirsch: I don't think these colors are even very good, but because in v2 we only have one tag, I wouldn't put too much effort into optimizing this. The result is cute and makes sense, and this is the easiest way to do it.

docs/src/components/version-selector.tsx
eranhirsch: I looked around, copied this design specifically from eslint, they have a similar approach of a single "live" version, and a single previous version which is completely deprecated where the docs are only there to aid the transition (they don't maintain docs for v7, v6, etc...).

docs/src/content/config.ts
eranhirsch: I'm thinking of doing this via the folder (`v1/`) instead of having any sort of version flag for the same reason that I've mentioned in other places: I don't want us to "maintain" a v1 site, hopefully, what we're doing now is the last we'll touch anything related to v1. I prefer not to build the whole site in a way that signals multi-version support.

My goal with this abstraction is to add extra documentation articles to the docs so that we remove the explanations from the index page so we can make it splashier and less verbose.

docs/src/pages/index.astro
eranhirsch: We can move the whole thing to something like MDX; it's probably pretty easy to do. However, I still think we'll change the main page considerably and don't want to put in the effort yet. We can defer to another PR; this PR is long enough as it is.

PR #679, opened 2024-05-16
docs(v2): style readme, move mapping

cjquines:
* cuts the readme
* moves mapping to docs site (unlinked from header for now)


docs/src/pages/mapping.astro
```diff
@@ -0,0 +1,11 @@
+---
```
eranhirsch, line 1: This shouldn't be part of the pages dir, put it in the layout dir instead, with a `<slot />` instead of the actual content, and set the layout in the markdown frontmatter. 

You can create a new layout with this and use that layout in the frontmatter for mapping-content.md, then rename mapping-content.md to mapping.md and astro would create a page for it without needing another component just to wrap it.

https://docs.astro.build/en/guides/markdown-content/#frontmatter-layout

cjquines: oof my lack of astro knowledge is showing

i'll get to this after #677 merges

docs/src/layouts/mapping.astro
cjquines, line 1: **thought:** probably this layout can be called something else if we eventually add more docs pages other than mapping, but we can deal with that later

cjquines, line 1: **thought**: probably this layout can be called something else if we eventually add more docs pages other than mapping, but we can deal with that later

eranhirsch (changes requested):
I think the README is too short. I think a lot of people checkout the github and npmjs pages just to see how serious the project is, and having a bunch of links to a docs site without anything else would be a red flag. Looking at all other similar projects they all have build status, a summary of key points, installation instructions, etc...

README.md
```diff
@@ -7,133 +7,49 @@ The first "data-first" and "data-last" utility library designed especially for T
...
-```
+## Features
 
-## Why Remeda?
+* First-class TypeScript support, with types that are as specific as possible. 
+* Supports data-first (`R.filter(array, fn)`) and data-last (`R.filter(fn)(array)`) approaches.
+* Lazy evaluation support with `pipe` and `piped`.
+* Runtime and types are both extensively tested, with full code coverage.
+* Tree-shakable, supports CJS and ESM.
+* Fully documented with JSDoc, supports in-editor function documentation.
```
eranhirsch, line 23: <3

README.md
```diff
@@ -7,133 +7,49 @@ The first "data-first" and "data-last" utility library designed especially for T
...
+## Documentation
 
-```bash
-npm i remeda@beta
-yarn add remeda@beta
-```
+Full documentation on [remedajs.com/docs](https://remedajs.com/docs).
 
-Then in .js or .ts
+For Lodash and Ramda users, check the function mapping on [remedajs.com/mapping](https://remedajs.com/mapping).
```
eranhirsch, line 14: I got used to the chatgpt style, I think longer sentences are better. Something along the lines of: "If you are migrating from Lodash or Ramda, check out our mapping to help you convert your project.

README.md
```diff
@@ -7,133 +7,49 @@ The first "data-first" and "data-last" utility library designed especially for T
...
 ![Dependencies](https://img.shields.io/librariesio/release/npm/remeda)
 
-## Installation
+## Documentation
 
-```bash
-npm i remeda@beta
-yarn add remeda@beta
-```
+Full documentation on [remedajs.com/docs](https://remedajs.com/docs).
```
eranhirsch, line 12: Similarly, something like "read the full docs and API reference"

README.md
```diff
@@ -7,133 +7,49 @@ The first "data-first" and "data-last" utility library designed especially for T
...
-// Remeda
-R.pick(obj, ["firstName", "lastName"]);
+### Installation
 
-// Ramda
-R.pick(["firstName", "lastName"], obj);
-
-// Lodash
-_.pick(obj, ["firstName", "lastName"]);
+```bash
```
eranhirsch, line 29: Is the hipster thing to also include pnpm and bun?

README.md
```diff
@@ -7,133 +7,49 @@ The first "data-first" and "data-last" utility library designed especially for T
...
-  { name: "bill", age: 33, gender: "m" },
-];
-
-// Remeda
-R.pipe(
-  users,
-  R.filter((x) => x.gender === "f"),
-  R.groupBy((x) => x.age),
-);
+import * as R from "remeda";
```
eranhirsch, line 37: Let's also mention that imports could be done by name. I think some people might find this syntax confusing.

```ts
import { map } from "remeda";
```

eranhirsch (approved):
Happy to get the README in better shape. My longer term goal is to break down the advantages into 3-4 distinct themes:
* Useful utilities (e.g. partition)
* Amazing types for compile time safety (e.g. map)
* Hybrid dataFirst and dataLast to make any style of code work in any part of the project.
* Implicit Lazy evaluation.

README.md
cjquines: i feel like if we were really hipster we would pick only bun and stick with that!

but fair point, i'll add it, it's already in the front page of the docs anyway

README.md
cjquines: i personally prefer brevity over longer sentences, but in this case i think something like

```md
Migrating from Lodash or Ramda? Check the function mapping on [remedajs.com/mapping](https://remedajs.com/mapping).
```

is better

PR #681, opened 2024-05-24
docs(v2): New front page

eranhirsch:
* Migrate the docs to use MDX instead of hardcoded html.
* Add the mapping to the navbar.


docs/src/pages/index.mdx
```diff
@@ -0,0 +1,178 @@
+---
+layout: ../layouts/wiki.astro
+---
+
+import InstallCommands from "@/components/install-commands.astro";
+
+# Remeda
+
+Fully written in TypeScript for powerful, type-safe, and functional utilities.
```
cjquines, line 9: **suggestion:** maybe something like

```suggestion
Powerful, type-safe, functional utilities for TypeScript.
```

cjquines: maybe we can even make it bigger font-size-wise or something, but can think about that later

docs/src/pages/index.mdx
```diff
@@ -0,0 +1,178 @@
...
+[`takeFirstBy`](/docs#takeFirstBy).
+
+```ts
+type User = {
+  firstName: string;
+  lastName: string;
+  lastLogin: number;
+};
+
+decalre const USERS: User[];
```
cjquines, line 30: **nit:** typo

```suggestion
declare const USERS: User[];
```

but also maybe we can use an actual example here? hm

docs/src/pages/index.mdx
```diff
@@ -0,0 +1,178 @@
...
+Remeda offers a wide range of common everyday utility functions so that you
+don't have to deal with the minutiae of writing them yourself. We ensure that
+our algorithms are efficient, optimized, and cover all edge cases properly. We
+offer a wide range of functions, from simple utilities like
+[`partition`](/docs#partition) and building objects with
+[`pullObject`](/docs#pullObject), to _complex_ algorithms like
+[`takeFirstBy`](/docs#takeFirstBy).
+
+```ts
+type User = {
```
cjquines, line 24: **nit:** let's add the import to show usage

```suggestion
```ts
import * as R from "remeda";

type User = {
```

docs/src/pages/index.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+Fully written in TypeScript for powerful, type-safe, and functional utilities.
+
+<InstallCommands />
+
+## 🧰 Expansive
+
+Remeda offers a wide range of common everyday utility functions so that you
+don't have to deal with the minutiae of writing them yourself. We ensure that
+our algorithms are efficient, optimized, and cover all edge cases properly. We
```
cjquines, line 17: **question:** i don't think we can say that our algorithms are "optimized" given that we don't run any benchmarks?

i can accept "efficient" though, given that stuff like `takeFirstBy` is n log n rather than n^2 or something

docs/src/pages/index.mdx
```diff
@@ -0,0 +1,178 @@
...
+offer a wide range of functions, from simple utilities like
+[`partition`](/docs#partition) and building objects with
+[`pullObject`](/docs#pullObject), to _complex_ algorithms like
```
cjquines, line 20: **nit:** let's drop the italics here

```suggestion
[`pullObject`](/docs#pullObject), to complex algorithms like
```

docs/src/pages/index.mdx
```diff
@@ -0,0 +1,178 @@
...
+```
+
+## 🦾 Functional
+
+Functional programming is beneficial, making code more readable. However, there
+are situations where you don't need "pipes" and want to call just a single
+function. In Remeda, **all** functions have automatic currying built-in. This
+means the same imported function can be called either regularly (what we call
+_dataFirst_) or with the first parameter, _data_, omitted (what we call
+_dataLast_). When a dataLast invocation is detected, Remeda creates a curried
```
cjquines, line 70: **question:** do we want to switch to using "dataFirst" and "dataLast" everywhere?

so far convention has been to use "data-first" and "data-last", which is standard, see e.g. [this article](https://www.javierchavarri.com/data-first-and-data-last-a-comparison/) or [lodash wiki](https://github.com/lodash/lodash/wiki/FP-Guide). i think this is grammatically correct, as data-first is a [compound modifier](https://en.wikipedia.org/wiki/Compound_modifier)

whatever we decide to use though, we should be consistent—we can edit the readme after this if we decide to prefer dataFirst and dataLast

docs/src/pages/index.mdx
```diff
@@ -0,0 +1,178 @@
...
+```ts
+// Get first 3 unique values
+const DATA = [1, 2, 2, 3, 3, 4, 5, 6];
+
+const result = R.pipe(
+  DATA,
+  R.map((x) => {
+    console.log("iterate", x);
+    return x;
+  }),
```
cjquines, line 151: **nit:** replace with `R.tap`? it's a bit less readable though, hm

docs/src/pages/index.mdx
```diff
@@ -0,0 +1,178 @@
...
+  DATA,
+  R.filter((x) => x.gender === "f"),
+  R.groupBy((x) => x.age),
+);
+
+// Ramda
+R.pipe(
+  R.filter((x) => x.gender === "f"),
+  R.groupBy((x) => x.age),
+)(DATA); // broken typings in TS :(
```
cjquines, line 115: **suggestion:** ooh, what if we try something like

```suggestion
// Ramda
R.pipe(
  R.filter((x) => x.gender === "f"),
  //              ^ ❌ 'x' is of type 'unknown'
  R.groupBy((x) => x.age),
  //                 ^ ❌ Property 'age' does not exist on type 'P'
)(DATA);
```

docs/src/pages/index.mdx
```diff
@@ -0,0 +1,178 @@
...
+_(DATA)
+  .filter((x) => x.gender === "f")
+  .groupBy((x) => x.age)
+  .value();
+
+// Lodash-fp
+_.flow(
+  _.filter((x) => x.gender === "f"),
+  _.groupBy((x) => x.age),
+)(DATA); // broken typings in TS :(
```
cjquines, line 127: **suggestion:** the library has now renamed to [lodash/fp](https://github.com/lodash/lodash/wiki/FP-Guide) i think

```suggestion
// Lodash/fp
fp.flow(
  fp.filter((x) => x.gender === "f"),
  //               ^ ❌ 'x' is of type 'unknown'
  fp.groupBy((x) => x.age),
  //                ^ ❌ 'x' is of type 'unknown'
)(DATA);
```

docs/src/pages/index.mdx
```diff
@@ -0,0 +1,178 @@
...
+R.pipe(
+  R.filter((x) => x.gender === "f"),
+  R.groupBy((x) => x.age),
+)(DATA); // broken typings in TS :(
+
+// Lodash
+_(DATA)
+  .filter((x) => x.gender === "f")
+  .groupBy((x) => x.age)
+  .value();
```
cjquines, line 121: **question:** hm, why is this example here? the types are all correct in this case, even the result type

docs/src/pages/index.mdx
```diff
@@ -0,0 +1,178 @@
...
+// iterate 1
+// iterate 2
+// iterate 2
+// iterate 3
+```
+
+## 🔬 Trustworthy!
+
+All this wouldn't be worth much if it didn't work. That's why Remeda has an
+extensive test suite that offers 100% coverage. We also test all functions on
```
cjquines, line 166: **praise:** nice

in my mind this is our claim to fame over ts-belt, which otherwise checks all the boxes here

docs/src/pages/index.mdx
cjquines: …i just now realize we use "data first" (with spaces) in the documentation…

docs/src/layouts/wiki.astro
```diff
@@ -0,0 +1,12 @@
+---
+import Prose from "@/components/prose.astro";
+import BaseLayout from "./base.astro";
+---
+
+<BaseLayout>
+  <main class="p-6">
+    <Prose class="container max-w-4xl">
```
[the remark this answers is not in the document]
cjquines: actually maybe we should just put `prose-p:max-w-2xl` in `Prose`, i think the "Migrating to v2" docs look so narrow…

cjquines (approved):
lots of small comments, but none blocking—we can always make more docs changes after we publish v2

docs/src/pages/index.mdx
eranhirsch: My issue with "for TypeScript" is that it doesn't differentiate between runtime and typing, type-fest and ts-toolbelt are TypeScript utility libraries, they focus on something different than us, and we don't offer exported types at all.

docs/src/pages/index.mdx
eranhirsch: I agree, I'll change it

docs/src/pages/index.mdx
eranhirsch: The tag was renamed from "data-first" to "dataFirst" because typedoc doesn't support it, and once I did that I got used to calling it like that, but I think it's fine to change it to what is the most common way.

docs/src/pages/index.mdx
eranhirsch: I like it!

docs/src/pages/index.mdx
eranhirsch: Isn't that the next example though?

docs/src/pages/index.mdx
eranhirsch: It's less understandable as an example, and causes prettier to put everything in a single line which makes it less readable too, I think that as an example this is clearer

docs/src/pages/index.mdx
cjquines: but "Fully written in TypeScript" doesn't do that either, i think. the main reason i made this suggestion was that it feels weird to lead the tagline with that, the fact that our library is "fully written" in typescript isn't as important as, say, the fact that it's type-safe, or typescript-first, or whatever

i also think "functional utilities" already says that it's a runtime level utility library. compare [ts-belt's tagline](https://github.com/mobily/ts-belt) ("Fast, modern, and practical utility library for FP in TypeScript.") or [radash's tagline](https://github.com/sodiray/radash) ("Functional utility library - modern, simple, typed, powerful")

we could put "for TS/JS" like, say, [tanstack query's tagline](https://tanstack.com/query/latest) ("Powerful asynchronous state management for TS/JS, React, Solid, Vue, Svelte and Angular")

eranhirsch: I didn't like the tagline either, but after trying to come up with something for a while, this felt like the one that covered my main points most closely. Let's try to come up with a better one; I just couldn't.

I don't want TS/JS because we don't want to support non-TS usage. Without typing, you can run into issues where we don't do runtime validation on inputs, and the code could explode. I recommend that people on JS projects not use Remeda and prefer Lodash, which does runtime validation (at least sometimes).

I mentioned that the library is written in TypeScript because none of the big libraries are. I thought this would communicate to people who think that a TypeScript codebase (like their own) is a higher-quality project than one in JS, so a library that is fully written in TypeScript is a sign of its quality.

cjquines: alright then! this wasn't a blocking comment anyway

Issue #678, opened 2024-05-15
Incorrect type assertion in negative case for hasSubObject 


JohanAlteruna:
Hi, 
I noticed that `hasSubObject`, when used as a type guard, behaves unexpectedly in the negative case:

```typescript
import * as R from "remeda";

const superObject = { foo: 1, bar: 2 };
const subObject = { foo: 3 };

if(R.hasSubObject(superObject, subObject)) {
    superObject // correctly typed as { foo: number; bar: number; }
} else {
    superObject // typed as never; should probably be { foo: number; bar: number } as well
}
```

Clearly, in this example, `subObject` is not a sub-object of `superObject`, and so the negative case is a valid outcome of the check.

[TS playground link](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wEpwGZQiOAIigFMQSATZQgbgCg6BjCAO1XlQFcwSoB5AEYArEo3gBeOAG8cECAC44ARgA0cAciiKATHAC+9Zmw6cBgkWLiSZ2OYoDM++nWDYAFBgB0ACzQBlU3NRGDcuHn5hYLUuM0ixAEp46To4VLgw3iDLAHpsuBgATx4KFHQbOzgWThABXhp1TUUqmrr9Oj04EgAbVBJktPTuTLj4XPyiylLKkgA3XnagA)

eranhirsch:
This makes sense, thanks for reporting

PR #683, opened 2024-05-29
fix(hasSubObject): incorrect type narrowing in negate condition

lesha1201:
Fixes: #678

The only way I come up to fix the issue with the negate condition in case when the guard resolves to the same type as `data` is by using something that will make the guard always different from `data`. One solution can be by using a brand type. Brand type is an internal type that doesn't affect anything but can help us identify the type result better. The users of the library can safely ignore that brand type.

This is the first occurrence of a brand type so let me know if I need to change its format/naming, etc.

Next, in a separate PR, I'm going to try to improve the overall typing of `hasSubObject` to make the narrowing more smarter. I already did some progress but stuck on making it work well with `pipe` function.

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/hasSubObject.test.ts
+++ b/src/hasSubObject.test.ts
@@ -1,19 +1,19 @@
-import { type HasSubObjectBrand, hasSubObject } from "./hasSubObject";
+import { hasSubObject } from "./hasSubObject";
 import { pipe } from "./pipe";
 
 describe("data first", () => {
-  test("works with empty sub-object", () => {
+  it("works with empty sub-object", () => {
     expect(hasSubObject({ a: 1, b: "b", c: 3 }, {})).toBe(true);
     expect(hasSubObject({}, {})).toBe(true);
   });
 
-  test("works with primitives", () => {
+  it("works with primitives", () => {
     expect(hasSubObject({ a: 1, b: "b", c: 3 }, { a: 1, b: "b" })).toBe(true);
     expect(hasSubObject({ a: 1, b: "c", c: 3 }, { a: 1, b: "b" })).toBe(false);
     expect(hasSubObject({ a: 2, b: "b", c: 3 }, { a: 1, b: "b" })).toBe(false);
   });
 
-  test("works with deep objects", () => {
+  it("works with deep objects", () => {
     expect(hasSubObject({ a: { b: 1, c: 2 } }, { a: { b: 1, c: 2 } })).toBe(
       true,
     );
@@ -22,19 +22,19 @@ describe("data first", () => {
     );
   });
 
-  test("checks for matching key", () => {
+  it("checks for matching key", () => {
     const data = {} as { a?: undefined };
     expect(hasSubObject(data, { a: undefined })).toBe(false);
   });
 });
 
 describe("data last", () => {
-  test("empty sub-object", () => {
+  it("works with empty sub-object", () => {
     expect(pipe({ a: 1, b: 2, c: 3 }, hasSubObject({}))).toBe(true);
     expect(pipe({}, hasSubObject({}))).toBe(true);
   });
 
-  test("works with primitives", () => {
+  it("works with primitives", () => {
     expect(pipe({ a: 1, b: "b", c: 3 }, hasSubObject({ a: 1, b: "b" }))).toBe(
       true,
     );
@@ -46,7 +46,7 @@ describe("data last", () => {
     );
   });
 
-  test("works with deep objects", () => {
+  it("works with deep objects", () => {
     expect(
       pipe({ a: { b: 1, c: 2 } }, hasSubObject({ a: { b: 1, c: 2 } })),
     ).toBe(true);
@@ -59,7 +59,7 @@ describe("data last", () => {
 
 describe("typing", () => {
   describe("data-first", () => {
-    test("must have matching keys and values", () => {
+    it("must have matching keys and values", () => {
       expectTypeOf(hasSubObject({ a: 2 }, { a: 1 })).toEqualTypeOf<boolean>();
 
       // @ts-expect-error [ts2353] - missing a key
@@ -81,7 +81,7 @@ describe("typing", () => {
       hasSubObject({ a: 2 }, { a: 1 } as const);
     });
 
-    test("allows nested objects", () => {
+    it("allows nested objects", () => {
       // ok - nested object has extra keys
       hasSubObject({ a: { b: 4 } }, { a: { b: 1, c: 2 } });
 
@@ -92,45 +92,49 @@ describe("typing", () => {
       hasSubObject({ a: { b: 4, c: "c" } }, { a: { b: 1, c: 2 } });
     });
 
-    test("narrows type", () => {
+    it("narrows with empty object", () => {
       const obj = {} as { a?: string; b?: number };
 
       if (hasSubObject(obj, {})) {
-        expectTypeOf(obj).toEqualTypeOf<
-          HasSubObjectBrand & { a?: string; b?: number }
-        >();
+        expectTypeOf(obj).toMatchTypeOf<{ a?: string; b?: number }>();
       } else {
-        expectTypeOf(obj).toEqualTypeOf<{ a?: string; b?: number }>();
+        expectTypeOf(obj).toMatchTypeOf<{ a?: string; b?: number }>();
       }
+    });
+
+    it("narrows with same object", () => {
+      const obj = {} as { a?: string; b?: number };
 
       if (hasSubObject(obj, obj)) {
-        expectTypeOf(obj).toEqualTypeOf<
-          HasSubObjectBrand & { a?: string; b?: number }
-        >();
+        expectTypeOf(obj).toMatchTypeOf<{ a?: string; b?: number }>();
       } else {
-        expectTypeOf(obj).toEqualTypeOf<{ a?: string; b?: number }>();
+        expectTypeOf(obj).toMatchTypeOf<{ a?: string; b?: number }>();
       }
+    });
+
+    it("narrows optional field to required", () => {
+      const obj = {} as { a?: string; b?: number };
 
       if (hasSubObject(obj, { a: "a" })) {
-        expectTypeOf(obj).toEqualTypeOf<
-          HasSubObjectBrand & { a: string; b?: number }
-        >();
+        expectTypeOf(obj).toMatchTypeOf<{ a: string; b?: number }>();
       } else {
-        expectTypeOf(obj).toEqualTypeOf<{ a?: string; b?: number }>();
+        expectTypeOf(obj).toMatchTypeOf<{ a?: string; b?: number }>();
       }
+    });
+
+    it("narrows field to constant type", () => {
+      const obj = {} as { a?: string; b?: number };
 
       if (hasSubObject(obj, { a: "a" } as const)) {
-        expectTypeOf(obj).toEqualTypeOf<
-          HasSubObjectBrand & { readonly a: "a"; b?: number }
-        >();
+        expectTypeOf(obj).toMatchTypeOf<{ readonly a: "a"; b?: number }>();
       } else {
-        expectTypeOf(obj).toEqualTypeOf<{ a?: string; b?: number }>();
+        expectTypeOf(obj).toMatchTypeOf<{ a?: string; b?: number }>();
       }
     });
   });
 
   describe("data-last", () => {
-    test("must have matching keys and values", () => {
+    it("must have matching keys and values", () => {
       expectTypeOf(
         pipe({ a: 2 }, hasSubObject({ a: 1 })),
       ).toEqualTypeOf<boolean>();
@@ -154,7 +158,7 @@ describe("typing", () => {
       pipe({ a: 2 }, hasSubObject({ a: 1 } as const));
     });
 
-    test("allows nested objects", () => {
+    it("allows nested objects", () => {
       // ok - nested object has extra keys
       pipe({ a: { b: 4 } }, hasSubObject({ a: { b: 1, c: 2 } }));
 
--- a/src/hasSubObject.test.ts
+++ b/src/hasSubObject.test.ts
@@ -1,4 +1,4 @@
-import { hasSubObject } from "./hasSubObject";
+import { type HasSubObjectBrand, hasSubObject } from "./hasSubObject";
 import { pipe } from "./pipe";
 
 describe("data first", () => {
@@ -96,15 +96,35 @@ describe("typing", () => {
       const obj = {} as { a?: string; b?: number };
 
       if (hasSubObject(obj, {})) {
+        expectTypeOf(obj).toEqualTypeOf<
+          HasSubObjectBrand & { a?: string; b?: number }
+        >();
+      } else {
+        expectTypeOf(obj).toEqualTypeOf<{ a?: string; b?: number }>();
+      }
+
+      if (hasSubObject(obj, obj)) {
+        expectTypeOf(obj).toEqualTypeOf<
+          HasSubObjectBrand & { a?: string; b?: number }
+        >();
+      } else {
         expectTypeOf(obj).toEqualTypeOf<{ a?: string; b?: number }>();
       }
 
       if (hasSubObject(obj, { a: "a" })) {
-        expectTypeOf(obj).toEqualTypeOf<{ a: string; b?: number }>();
+        expectTypeOf(obj).toEqualTypeOf<
+          HasSubObjectBrand & { a: string; b?: number }
+        >();
+      } else {
+        expectTypeOf(obj).toEqualTypeOf<{ a?: string; b?: number }>();
       }
 
       if (hasSubObject(obj, { a: "a" } as const)) {
-        expectTypeOf(obj).toEqualTypeOf<{ a: "a"; b?: number }>();
+        expectTypeOf(obj).toEqualTypeOf<
+          HasSubObjectBrand & { readonly a: "a"; b?: number }
+        >();
+      } else {
+        expectTypeOf(obj).toEqualTypeOf<{ a?: string; b?: number }>();
       }
     });
   });
--- a/src/hasSubObject.ts
+++ b/src/hasSubObject.ts
@@ -1,6 +1,8 @@
-import type { Simplify } from "type-fest";
+import type { Merge } from "type-fest";
 import { isDeepEqual } from "./isDeepEqual";
 import { purry } from "./purry";
+
+export type HasSubObjectBrand = { __remeda: "hasSubObject" };
 
 /**
  * Checks if `subObject` is a sub-object of `object`, which means for every
@@ -19,10 +21,9 @@ import { purry } from "./purry";
  * @category Guard
  */
 export function hasSubObject<T, S extends Partial<T>>(
-  data: T,
+  data: Merge<T, S> | T,
   subObject: S,
-): // TODO: want to use data is Merge<T, S> here, but it's a typescript error. fix after we allow @ts-expect-error in the build process
-data is Simplify<S & T>;
+): data is HasSubObjectBrand & Merge<T, S>;
 
 /**
  * Checks if `subObject` is a sub-object of `object`, which means for every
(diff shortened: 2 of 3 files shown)


lesha1201:
> Next, in a separate PR, I'm going to try to improve the overall typing of hasSubObject to make the narrowing more smarter. I already did some progress but stuck on making it work well with pipe function.

Opened PR - https://github.com/remeda/remeda/pull/686.

src/hasSubObject.ts
```diff
@@ -19,10 +21,9 @@ import type { Simplify } from "./type-fest/simplify";
  * @category Guard
  */
 export function hasSubObject<T, S extends Partial<T>>(
-  data: T,
+  data: Merge<T, S> | T,
   subObject: S,
-): // TODO: want to use data is Merge<T, S> here, but it's a typescript error. fix after we allow @ts-expect-error in the build process
-data is Simplify<S & T>;
+): data is HasSubObjectBrand & Merge<T, S>;
```
lesha1201, line 26: Reverted this to `Simplify<S & T>` in https://github.com/remeda/remeda/pull/686 because `Merge` is quite a complex type and I couldn't make `pipe` to work well in this case.  `Simplify` worked well in all cases.

lesha1201:
Fixes: https://github.com/remeda/remeda/issues/678

src/hasSubObject.test.ts
```diff
@@ -96,15 +96,35 @@ describe("typing", () => {
       const obj = {} as { a?: string; b?: number };
 
       if (hasSubObject(obj, {})) {
+        expectTypeOf(obj).toEqualTypeOf<
+          HasSubObjectBrand & { a?: string; b?: number }
+        >();
```
cjquines, line 101: i'm hesitant to include a brand type, because it'd impact how it works with the rest of the library. for example, in this branch, if we had `keys(obj)`, its type would be `Array<"a" | "b" | "__remeda">`, rather than the expected `Array<"a" | "b">`. i don't think the tradeoff here is worth it

lesha1201: @cjquines Yeah, I thought about this but couldn't come up with a real world example where it would affect the user. Anyway, I agree that it might be too much adding type guard so feel free to close PR. I couldn't come up with a solution without a brand type.

cjquines: are your changes in #686 independent of this?

my vote is a weak no, but i'd also like to hear @eranhirsch 's thoughts

lesha1201: > are your changes in https://github.com/remeda/remeda/pull/686 independent of this?

@cjquines Yes, they are. https://github.com/remeda/remeda/pull/686 handles a lot of cases but also might be too much for such function.

src/hasSubObject.ts
```diff
@@ -1,7 +1,9 @@
-import type { Simplify } from "type-fest";
+import type { Merge } from "type-fest";
 import { isDeepEqual } from "./isDeepEqual";
 import { purry } from "./purry";
 
+export type HasSubObjectBrand = { __remeda: "hasSubObject" };
```
eranhirsch, line 5: I haven't deep dived into this PR yet so this isn't a full review.

Shouldn't the branding property use a symbol? It would prevent that property from being part of the object for most use-cases (like JSON serialization, our object functions, etc...).

lesha1201: @eranhirsch Yes, I've just been reading about this in [an article](https://egghead.io/blog/using-branded-types-in-typescript) and was going to write about it  :).

Here is an example of how Effect implements brand type - https://github.com/Effect-TS/effect/blob/main/packages/effect/src/Brand.ts#L30-L36

I will change it if you give the green light for the usage of brand type in that PR.

eranhirsch: If the usage of the branded types doesn't require changes to the tests and they still pass, I'm OK with the change; if this change requires changing the tests to accommodate it, then it's problematic.

lesha1201: @eranhirsch By tests, do you mean only type tests, only regular tests or both?

It doesn't require changes for the regular tests. It requires changes for the type tests where we use `toEqualTypeOf` but doesn't where we use `toMatchTypeOf`.

eranhirsch: Yea, that makes sense. I think those are OK as the `MatchTypeOf` would mean that this isn't a breaking change for anyone.

@cjquines — Using symbols instead of string keys and values makes sense to me. What's your stance here?

cjquines: if symbols were used, then i'm fine with shipping this PR!

lesha1201: Then I will update the PR (probably at the beginning of the next week) and will let you know when it's done.

lesha1201: @eranhirsch @cjquines Done.

src/hasSubObject.ts
```diff
@@ -1,7 +1,10 @@
-import type { Simplify } from "type-fest";
+import type { Merge } from "type-fest";
+import type { Brand } from "./internal/types";
 import { isDeepEqual } from "./isDeepEqual";
 import { purry } from "./purry";
 
+export type HasSubObjectBrand = Brand<"hasSubObject">;
```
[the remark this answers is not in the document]
eranhirsch: We made a big effort to ensure that our types treat symbol props properly, but the basic types built into TypeScript could still create some confusing edge cases.

https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAQgJQIIDkAiMC8MDKBPAWwCMQAbACgHIiAnAQzABNKBKAbgChRJY4tdCJCtVacu4aDAAa-AN4wA2olRoAugC54AGhgALAKalSITQEYATAGYYAXxh0IMbtE7PYATTmLl6DTABERP46BkYmMBbWdg5OElBiHFB4AA760qb8SakgAGbSCgDW+ni5MFn6pVKqHAD0NTANAHoA-ByJKWnuGdjlpe6FxaW9ee6qnHUNMC1AA

Using a symbol key **and value** here makes sense. Symbols are great when you need to differentiate between intents when passing objects around; they make it harder to override or accidentally use the wrong values.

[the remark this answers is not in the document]
lesha1201: @eranhirsch Makes sense, thanks for the example. I will update it to use symbol for the value. Something like this:

```ts
declare const hasSubObjectBrandId: unique symbol;

export type HasSubObjectBrand = Brand<hasSubObjectBrandId>
```

But before I do this, I have another question. I realized that using that single brand type won't handle similar case in nested conditions. For example:

```ts
const obj = { a: 5, b: 5 };

if (hasSubObject(obj, { a: 5 })) {
  obj; // { a: number; b: number; } & HasSubObjectBrand

  if (hasSubObject(obj, { b: 5 })) {
    obj; // { a: number; b: number; } & HasSubObjectBrand
  } else {
    obj; // never <--
  }
} else {
  obj; // { a: number; b: number; }
}
```

It can be solved by having several unique brand types for `hasSubObject` and by recursively checking what brand type `Merge<T, S>` extends:

```ts
declare const hasSubObjectBrandId: unique symbol;
declare const hasSubObjectBrandId2: unique symbol;

export type HasSubObjectBrand = {
  readonly [hasSubObjectBrandId]: typeof hasSubObjectBrandId;
};
// Note: naming isn't final
export type HasSubObjectBrand2 = {
  readonly [hasSubObjectBrandId2]: typeof hasSubObjectBrandId2;
};

export function hasSubObject<T, S extends Partial<T>>(
  data: Merge<T, S> | T,
  subObject: S,
): data is Merge<T, S> extends HasSubObjectBrand
  ? Merge<T, S> & HasSubObjectBrand2
  : Merge<T, S> & HasSubObjectBrand;
  
const obj = { a: 5, b: 5 };
const subObj = { b: 5 };

if (hasSubObject(obj, subObj)) {
  obj; // { a: number; b: number; } & HasSubObjectBrand

  if (hasSubObject(obj, subObj)) {
    obj; // { a: number; b: number; } & HasSubObjectBrand & HasSubObjectBrand2 
  } else {
    obj; // { a: number; b: number; } & HasSubObjectBrand
  }
} else {
  obj; // { a: number; b: number; }
}
```

I'm not sure if we need to handle that case. What do you think? Should we handle it? If so, how deep we should handle this?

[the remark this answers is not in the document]
cjquines: i think it's fine if we don't handle this case; the workaround is

```ts
if (hasSubObject(obj, { ...subObj1, ...subObj2 }) {
  // ...
} else if (hasSubObject(obj, subObj1)) {
  // ...
} else if (hasSubObject(obj, subObj2)) {
  // ...
} else {
  // ...
}
```

right? if there's a complaint in the future, like if someone wants to nest more than two hasSubObjects, then we can look into this again, but for now i think it's fine?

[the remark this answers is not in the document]
lesha1201: > if there's a complaint in the future, like if someone wants to nest more than two hasSubObjects, then we can look into this again, but for now i think it's fine?

@cjquines Agree. There are different workarounds and the case probably isn't popular so it shouldn't be a blocker.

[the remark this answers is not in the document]
lesha1201: @eranhirsch @cjquines Updated PR to use symbol for the value of the brand type.

src/internal/types.ts
```diff
@@ -1,5 +1,9 @@
 import type { IsAny, IsNever } from "type-fest";
 
+declare const __brand: unique symbol;
+
+export type Brand<T> = { [__brand]: T };
```
eranhirsch, line 5: 1. Should brand only accept symbols?
2. To make the API more fail-proof I'd limit the usage to also branding the object, so something like this:
```suggestion
export type Branded<T, Brand extends symbol> = T & { [__brand]: Brand };
```

src/hasSubObject.ts
```diff
@@ -1,7 +1,12 @@
-import type { Simplify } from "type-fest";
+import type { Merge } from "type-fest";
+import type { Brand } from "./internal/types";
 import { isDeepEqual } from "./isDeepEqual";
 import { purry } from "./purry";
 
+declare const hasSubObjectBrandId: unique symbol;
+
+export type HasSubObjectBrand = Brand<typeof hasSubObjectBrandId>;
```
eranhirsch, line 8: 1. The name feels off to me. Why is it called "has"? The "has" prefix should be used for boolean values.
2. Don't export the type here, we don't include types in our exports as we don't consider our types part of the offering. See my comment in the test file, you don't need this exported.
3. Let's pull more of the common type up so it's easier to follow the logic. Taking my comment from "types.ts" about calling the utility type Branded.
```suggestion
declare const BRAND_SUB_OBJECT: unique symbol;

type SubObject<T, S> = Branded<Merge<T, S>, typeof BRAND_SUB_OBJECT>;
```

Now functions need to return `is SubObject<T, S>` which fully covers the intent of the type without having to understand the branding concept.

src/hasSubObject.test.ts
```diff
@@ -1,4 +1,4 @@
-import { hasSubObject } from "./hasSubObject";
+import { type HasSubObjectBrand, hasSubObject } from "./hasSubObject";
```
eranhirsch, line 1: don't import the type, we talked about using `toMatchTypeOf` instead of `toEqualTypeOf` to bypass the branded issue.
```suggestion
import { hasSubObject } from "./hasSubObject";
```

src/hasSubObject.test.ts
```diff
@@ -96,15 +96,35 @@ describe("typing", () => {
       const obj = {} as { a?: string; b?: number };
 
       if (hasSubObject(obj, {})) {
+        expectTypeOf(obj).toEqualTypeOf<
+          HasSubObjectBrand & { a?: string; b?: number }
```
eranhirsch, line 100: Use `toMatchTypeOf` and don't include the brand.
```suggestion
        expectTypeOf(obj).toMatchTypeOf<
          { a?: string; b?: number }
```

eranhirsch (changes requested):
This is a really cool solution to the problem. Good job!

src/internal/types.ts
lesha1201: @eranhirsch 

1. The initial idea was to create a general-purpose utility type but since it's an internal type and we don't expect anything other than symbol I think it makes sense to constrain it.
2. I thought about it but ended up with just `Brand` type to have more flexibility. I will change it to what you proposed.

src/hasSubObject.ts
lesha1201: @eranhirsch 

1. The naming has this format `<functionName><name>` and since function is called `hasSubObject` it has the name `HasSubObjectBrand`.
2. 👌
3. 👌

src/hasSubObject.test.ts
```diff
@@ -96,15 +96,35 @@ describe("typing", () => {
       const obj = {} as { a?: string; b?: number };
 
       if (hasSubObject(obj, {})) {
```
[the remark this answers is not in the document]
lesha1201: @eranhirsch In #683, I used a block statement ([link](https://github.com/remeda/remeda/pull/686/files#diff-aa20fb5170cbafcaeb26ac51cb81c5c91a4de4f0b89fa3d7d42fb7f34546a091R158-R192)) to isolated each case since there are a lot of cases and it's hard to name. Will this work for this PR as well?

src/hasSubObject.ts
eranhirsch: The naming confused me but you are completely right that the Has part comes from the function name 🤦🏼 , my bad.

src/hasSubObject.test.ts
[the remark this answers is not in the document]
eranhirsch: I haven't looked at the other PR yet, I prefer each PR be a standalone thing in terms of code quality, as we might not ship that other PR at the same time. I prefer splitting the tests along the lines I mentioned, I like each test have a few asserts, the whole point of tests is to split asserts on logical boundaries.

src/internal/types.ts
lesha1201: Done

src/hasSubObject.ts
lesha1201: Done.

Decided to keep the convention for naming `<functionName><name>` to make it easier to identify where this type comes from for the users. Using `SubObject` name might be confusing because the type guard doesn't resolve to sub-object but to a merged super-object and sub-object. Used a generic name `HasSubObjectGuard` instead to save connection of the type to that function and show that it's a type from a type guard function. Other options:

- `HasSubObjectData`
- `HasSubObjectMergedData`
- `HasSubObjectNarrowedData`
- `MergedData` or `MergedObject`
- `NarrowedData` or `NarrowedObject`
- `DataWithSubObject`
- `DataHasSubObject`

Let me know if I should rename it and I will do it.

By the way, wrapping it in a separate type makes it less convenient to see the type information from TS for the users:

| After | Before |
| ---- | ---- |
| image | image

It's not a big deal, just letting you know.

src/hasSubObject.test.ts
lesha1201: Done

src/hasSubObject.test.ts
```diff
@@ -96,15 +96,35 @@ describe("typing", () => {
...
+        >();
+      } else {
         expectTypeOf(obj).toEqualTypeOf<{ a?: string; b?: number }>();
       }
 
       if (hasSubObject(obj, { a: "a" })) {
-        expectTypeOf(obj).toEqualTypeOf<{ a: string; b?: number }>();
+        expectTypeOf(obj).toEqualTypeOf<
+          HasSubObjectBrand & { a: string; b?: number }
+        >();
```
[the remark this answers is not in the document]
lesha1201: Done

src/hasSubObject.test.ts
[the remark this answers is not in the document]
lesha1201: Done. Also changed all `test` to `it` to make it more similar to how `hasAtLeast.test.ts` is written.

lesha1201:
@eranhirsch Handled all your comments and left comments to the prior discussions.

src/hasSubObject.test.ts
```diff
@@ -92,25 +92,49 @@ describe("typing", () => {
       hasSubObject({ a: { b: 4, c: "c" } }, { a: { b: 1, c: 2 } });
     });
 
-    test("narrows type", () => {
+    it("narrows with empty object", () => {
```
eranhirsch, line 95: this is actually "**doesn't** narrow on empty object"

src/hasSubObject.test.ts
```diff
@@ -92,25 +92,49 @@ describe("typing", () => {
...
 
       if (hasSubObject(obj, {})) {
-        expectTypeOf(obj).toEqualTypeOf<{ a?: string; b?: number }>();
+        expectTypeOf(obj).toMatchTypeOf<{ a?: string; b?: number }>();
+      } else {
+        expectTypeOf(obj).toMatchTypeOf<{ a?: string; b?: number }>();
       }
+    });
+
+    it("narrows with same object", () => {
```
eranhirsch, line 105: and "**doesn't** narrow on same object"

src/hasSubObject.ts
```diff
@@ -1,7 +1,15 @@
-import type { Simplify } from "type-fest";
+import type { Merge } from "type-fest";
+import type { Branded } from "./internal/types";
 import { isDeepEqual } from "./isDeepEqual";
 import { purry } from "./purry";
 
+declare const HAS_SUB_OBJECT_BRAND: unique symbol;
+
+type HasSubObjectGuard<T, S> = Branded<
```
eranhirsch, line 8: I'm still not sure about the name, I prefer a more descriptive name and not just a formulaic one. Maybe "RefinedObject", e.g. `hasSubObject<T, S>(data: T, subObj: S): data is RefinedObject<T, S>`, I feel this is the most readable version of the name...

@cjquines , wdyt?

eranhirsch (approved):
This looks great, I'll let @cjquines go over this before we ship it. I have 2 nits inline about the test naming.

src/hasSubObject.ts
lesha1201: @eranhirsch I left some options here - https://github.com/remeda/remeda/pull/683#discussion_r1629100613. What do you think about `WithSubObject`? `RefinedObject` sounds good too.

src/hasSubObject.test.ts
lesha1201: @eranhirsch I thought about this but then thought that this test is about narrowing when using a type guard function and that it works properly in case of using an empty object: "it properly narrows type when using an empty object for the `subObject`". So I'm not sure we should use `doesn't` here.

In the test, we ensure that the type narrowing properly works when using an empty object in both branches of the condition.

Do you think I still should use `doesn't` here?

src/hasSubObject.ts
cjquines: maybe something like

```ts
type WithSubObject<T> = Branded<T, typeof HAS_SUB_OBJECT_BRAND>;

function hasSubObject<T, S>(data: T, subObj: S): data is WithSubObject<Merge<T, S>>;
```

that way we still get nice hover info

image

cjquines: not sure how that reads though... `obj` doesn't just have  `{ b?: number; readonly a: "a"; }` with a subobject it _is_ that

in this case we want something small and ignorable, like `Is` or `Wrap`

(the ideal would be `data is Simplify<WithSubObject<T, S>>`, but that doesn't typecheck?)

cjquines (approved):

src/hasSubObject.ts
lesha1201: @cjquines 

> not sure how that reads though... obj doesn't just have { b?: number; readonly a: "a"; } as a subobject type. it is that type

Yeah, in this case it's probably better to have `HasSubObjectGuard`, `HasSubObjectBranded`, `RefinedObject`, `BrandedObject`.

> in this case maybe we want something small and ignorable, like Is or Wrap? read: obj has the type Is<{ b?: number; readonly a: "a"; }>?

We can use `Branded` but I'm not sure about this since we also have an internal type with such name.

> the ideal would be data is Simplify<WithSubObject<T, S>>, but that doesn't typecheck?

We can make it typecheck (even with `Merge`). But in this case, I would probably remove `Merge` at all and revert `Simplify` usage and use it as the following:

image

image


In another PR, I had to use `Simplify` instead of `Merge` anyway - https://github.com/remeda/remeda/pull/683#discussion_r1620560556.

cjquines: let's stick with that then—the simplify option looks better to me?

eranhirsch:
@cjquines, let's get this shipped if you don't have any objections, type names can be changed as we don't export them

cjquines:
sgtm, i'll merge

PR #685, opened 2024-05-29
build(v2): Switch CI to use main instead of master

eranhirsch:
main has been checked out from the 'beta' branch so now represents the HEAD for v2 and remeda in general.

This PR makes the CI treat main as a valid branch for releases. by making it breaking it would de-facto release v2.

BREAKING CHANGE: Releases Remeda v2.


cjquines (approved):
let's try it

PR #686, opened 2024-05-30
feat(hasSubObject): improve typing

lesha1201:
> This PR depends on https://github.com/remeda/remeda/pull/683. See 35afa2fee17aa0fe16af3d467c18b22fabe23a44 for changes.

**Depends on:** https://github.com/remeda/remeda/pull/683

- make type narrowing smarter
- handle cases with overlapping union types (e.g. `{ a: string | number}, { a: number | boolean }`)
- improve typing for nested objects

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/hasSubObject.test.ts
+++ b/src/hasSubObject.test.ts
@@ -59,95 +59,118 @@ describe("data last", () => {
 
 describe("typing", () => {
   describe("data-first", () => {
-    it("must have matching keys and values", () => {
+    it("returns boolean", () => {
       expectTypeOf(hasSubObject({ a: 2 }, { a: 1 })).toEqualTypeOf<boolean>();
+    });
 
+    it("doesn't require sub-object to have all keys", () => {
       // ok
       hasSubObject({ a: 1, b: 2 }, { b: 2 });
+    });
 
+    it("doesn't allow sub-object to have distinct keys from super-object", () => {
       // @ts-expect-error [ts2322] - non-matching key
       hasSubObject({ b: 2 }, { a: 1 });
 
       // @ts-expect-error [ts2322] - non-matching key
       hasSubObject({ b: 2 }, { b: 2, a: 1 });
+    });
 
-      // @ts-expect-error [ts2322] - different value type
+    it("doesn't allow keys to have different types", () => {
+      // @ts-expect-error [ts2322] - `number` isn't assignable to `string`
       hasSubObject({ a: "a" }, { a: 1 });
 
-      // @ts-expect-error [ts2322] - different value type
+      // @ts-expect-error [ts2322] - object isn't assignable to `string`
       hasSubObject({ a: "a" }, { a: { b: 2 } });
 
-      // @ts-expect-error [ts2322] - different value type
+      // @ts-expect-error [ts2322] - `string` isn't assignable to object
       hasSubObject({ a: { b: 2 } }, { a: "a" });
+    });
 
-      // ok - wider value type
+    it("allows keys to have different types when types are overlapping", () => {
+      // ok - union type in super-object
       hasSubObject({ a: "a" as number | string }, { a: 1 });
 
-      // ok - wider value type
+      // ok - union type in sub-object
       hasSubObject({ a: "a" }, { a: 1 as number | string });
 
-      // ok - wider value type
+      // ok - union type in both super-object and sub-object
       hasSubObject({ a: "a" as number | string }, { a: 1 as boolean | number });
+    });
 
-      // @ts-expect-error [ts2345] - only allow valid objects to be passed
-      hasSubObject({ a: 2 } as unknown, { a: 1 });
-
+    it("allows const types", () => {
       // ok - const value
       hasSubObject({ a: 2, b: 2 }, { a: 1 } as const);
     });
 
+    it("only allows valid objects to be passed", () => {
+      // @ts-expect-error [ts2345] - only allow valid objects to be passed
+      hasSubObject({ a: 2 } as unknown, { a: 1 });
+
+      // @ts-expect-error [ts2345] - only allow valid objects to be passed
+      hasSubObject({ a: 2 }, { a: 1 } as unknown);
+    });
+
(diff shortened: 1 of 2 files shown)


eranhirsch:
Where does this PR stand now that #683 shipped? @lesha1201, are you planning on rebasing this?

lesha1201:
@eranhirsch Yes, I will rebase it probably this week.

lesha1201:
@eranhirsch Rebased.

src/hasSubObject.test.ts
cjquines, line 1: **suggestion:** i know you're following the convention in the existing file of putting multiple tests in a single block, but i think we can split up the blocks here meaningfully?

src/hasSubObject.ts
```diff
@@ -25,10 +75,10 @@ type HasSubObjectGuard<T, S> = Simplify<
  * @dataFirst
  * @category Guard
  */
-export function hasSubObject<T, S extends Partial<T>>(
-  data: T,
-  subObject: S,
-): data is HasSubObjectGuard<T, S>;
+export function hasSubObject<
+  T extends Record<string, unknown>,
```
cjquines, line 79: **question:** why `string` as key? shouldn't this be any `PropertyKey`

lesha1201: @cjquines Good catch, will fix it.

src/hasSubObject.test.ts
lesha1201: @cjquines I will try to split them.

lesha1201: @cjquines Done

src/hasSubObject.test.ts
```diff
@@ -126,47 +211,395 @@ describe("typing", () => {
...
-    it("must have matching keys and values", () => {
+    it("returns boolean", () => {
       expectTypeOf(
         pipe({ a: 2 }, hasSubObject({ a: 1 })),
       ).toEqualTypeOf<boolean>();
+    });
+
+    it("doesn't require sub-object to have all keys", () => {
+      // ok
+      hasSubObject({ b: 2 })({ a: 1, b: 2 });
```
cjquines, line 314: **nit:** let's cut the data-last tests that aren't in a pipe—we don't usually test these

src/hasSubObject.test.ts
```diff
@@ -126,47 +211,395 @@ describe("typing", () => {
...
+      })({
+        a: { b: 1 as boolean | number },
+      });
+      pipe(
+        { a: { b: 1 as boolean | number } },
+        hasSubObject({ a: { b: 1 as number | string } }),
+      );
+    });
+
+    it("narrows with empty object", () => {
```
cjquines, line 481: **question:** can we rewrite these to be in a pipe? `filter(hasSubObject)` for example

src/hasSubObject.ts
```diff
@@ -9,6 +9,56 @@ type HasSubObjectGuard<T, S> = Simplify<
...
+      : A[Key] & B[Key] extends object
+        ? B[Key]
+        : A[Key];
+}> & {
+  [Key in
+    | Exclude<keyof A, keyof B>
+    | Exclude<keyof B, keyof A>]: Key extends keyof B ? B[Key] : never;
+};
+
+type HasSubObjectData<
```
cjquines, line 26: **nit:** i don't really like this name, but i can't think of a better one right now. same with `HasSubObjectSubObject`

src/hasSubObject.test.ts
lesha1201: @cjquines Yes, we can rewrite it but in this case we won't be testing the negate type guard condition (`else`). Is it ok for data-last test cases?

Do you want something like this?

```tsx
const data = [] as Array<{ a?: string; b?: number }>;

pipe(data, filter(hasSubObject({ a: "a" })), (x) => {
  expectTypeOf(x).toMatchTypeOf<Array<{ a: string; b?: number }>>();
});
```

src/hasSubObject.test.ts
lesha1201: @cjquines I think it's better to leave them because their behavior can be different. Tests with `pipe` depends on `pipe`'s implementation.

src/hasSubObject.ts
lesha1201: @cjquines Alternative is to remove `HasSubObject` prefix and have:

- `HasSubObjectData` -> `Data`
- `HasSubObjectSubObject` -> `SubObject` or `DataSubObject`
- `HasSubObjectObjectValue` -> `DataObjectValue` or `ObjectValue`

src/hasSubObject.test.ts
cjquines: `filter(isNot(hasSubObject))` would be the most likely usage in the field, i think?

src/hasSubObject.test.ts
cjquines: i agree this is usually the case, but by far the most common use case for data-last functions are in pipes, and testing it there brings it closer to what people use

src/hasSubObject.test.ts
lesha1201: @cjquines It's not the same as the negate type guard case (`else` branch). `isNot` creates a new type guard. If someone needs to use data-last function with `else` condition then it won't be covered by tests though the case is quite possible. I think it's better to have tests for both. One tests the data-last function as it is. Another tests the data-last function with a real use case.

src/hasSubObject.test.ts
lesha1201: @cjquines That's why I left both. One tests it as it is and another tests a more common scenario.

src/hasSubObject.test.ts
cjquines: fair enough. it'd be nice if we got the `isNot` case working, but this is fine for now

src/hasSubObject.ts
cjquines: i'm fine going ahead with this name for now, since we can always change it later

PR #708, opened 2024-06-02
docs(mapping): Move mappings to astro content infra

eranhirsch:
Using astro content to streamline our mapping docs:
1. Splitting lodash and ramda, rarely will the user of one lib care about the mapping for the other
2. Indexing the mapping on the mapped function, and not the remeda equivalent, as that's the direction the user cares about.
3. Allowing both a simple mapping and additional content in markdown to add comments, examples, caveats, etc...
4. The ability to not provide a remeda function and instead provide the alternative, in cases where we don't have an equivalent function.
5. This allows adding docs to the mapping, like prerequisite steps or general instructions.
6. Allows mapping any library, without getting into issues with the size of the table.
7. Less merge conflicts because there isn't a single monolithic object to edit.


cjquines:
do we want to maintain this in the docs folder? in #519 it was suggested we put it in jsdoc. i guess a similar question would be like, do we still want to keep the metadata in the jsdoc, then, if we're going to have this?

eranhirsch:
> in #519 it was suggested we put it in jsdoc. i guess a similar question would be like, do we still want to keep the metadata in the jsdoc, then, if we're going to have this?

I think this solution makes more sense, putting it in the JSDoc means that everyone using Remeda has some abstract "dependency" on Lodash and Ramda. It doesn't make sense. Originally I thought it's a clean way to bundle everything together, mainly so we can stop maintaining the mapping file, but with this design we can decouple everything. We can build more mappings without impacting other mappings or the Remeda code itself. Mapping live on their own.

eranhirsch:
> do we want to maintain this in the docs folder? 

It could, but I feel this is a different concern. It allows us to add metadata specific to mappings (like the "remeda" field). It makes the **docs** folder a classical wiki for adding articles to a docs page while the **mapping** objects focus on just the mapping entry.

docs/src/pages/migrate/ramda.md
cjquines, line 1: **question:** is the intent to eventually also move this to individual files?

docs/src/content/mapping/lodash/some.md
```diff
@@ -0,0 +1,17 @@
+---
+category: Collection
+---
+
+_Not provided by Remeda._
+
+Use the native JS [`Array.prototype.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) instead.
+
+_**@see** ["You don't need Lodash"](https://you-dont-need.github.io/You-Dont-Need-Lodash-Underscore/#/?id=_some)_
```
cjquines, line 9: **question:** does this work? what does this look like?

docs/src/content/mapping/lodash/set.md
```diff
@@ -0,0 +1,52 @@
...
+  only the array representation is accepted. Use the helper function
+  [`stringToPath`](/docs#stringToPath) to translate string paths to array paths.
+- Unlike the Lodash `set` function, In Remeda the provided value must match the
+  type of the prop at that path, the function does not support creating "sparse"
+  objects.
+- For better type-safety, Remeda offers two additional functions to handle
+  paths of length **1**. Use [`set`](/docs#set) to update an existing prop in an
+  object (best type-safety and with IDE type-ahead); and
+  [`addProp`](/docs#addProp) To add a new prop to the object (or to override
+  it's _type_)
```
cjquines, line 18: **nit:**

```suggestion
  object (best type-safety and with IDE type-ahead), and
  [`addProp`](/docs#addProp) to add a new prop to the object (or to override
  its _type_)
```

docs/src/content/mapping/lodash/sampleSize.md
```diff
@@ -0,0 +1,33 @@
+---
+category: Collection
+remeda: sample
+---
+
+- Unlike Lodash, In Remeda the `sample` function doesn't also randomize the
+  **order** of the items in the result. Use the `shuffle` function to also
+  randomize the result.
```
cjquines, line 8: **suggestion:**

```suggestion
- Unlike Lodash, In Remeda the `sample` function returns items in the
  **same order** they appear in the input. Use the `shuffle` function to also
  randomize the result.
```

docs/src/content/mapping/lodash/sample.md
```diff
@@ -0,0 +1,41 @@
+---
+category: Collection
+remeda: sample
+---
+
+- The `sample` function in Remeda takes a required second parameter of the size
+  of the sample. Use **1**.
+- The Remeda function returns an array; to extract the value either use
+  native destructuring or index access, or use [`first`](/docs#first).
```
cjquines, line 9: **question:** maybe `only`?

docs/src/content/mapping/lodash/reverse.md
```diff
@@ -0,0 +1,16 @@
+---
+category: Array
+remeda: reverse
+---
+
+Lodash mutate's the input but in Remeda a new array is returned instead.
```
cjquines, line 6: **nit:**

```suggestion
Lodash mutates the input but in Remeda a new array is returned instead.
```

docs/src/content/mapping/lodash/pickBy.md
```diff
@@ -0,0 +1,6 @@
+---
+category: Object
+remeda: pickBy
+---
+
+_TODO: Check how lodash handles symbols and document any differences._
```
cjquines, line 6: **todo:** comment this out?

```suggestion
<!-- TODO: Check how lodash handles symbols and document any differences. -->
```

cjquines: same with the rest, unless you intend to write these up before merging

cjquines (commented):
this is awesome, thanks for all the effort so far!!

docs/src/pages/migrate/ramda.md
eranhirsch: yes, it's a lot of work because I'm double-validating that the existing mapping is correct at the same time...

docs/src/content/mapping/lodash/some.md
eranhirsch: image

cjquines: hm...... it's not great, but it's fine for now

cjquines (approved):
thanks for starting this huge effort!

Issue #554, opened 2024-03-01
`setPath` only works for existing properties


bombillazo:
Hello, looking at the `setPath` logic, it seems to check if the path value on the object is not null/undefined before attempting to insert the new value. How can we set a new value to the object that is not yet defined?

cjquines:
the trouble with this is with undefined deeply nested paths:

```ts
type T = Record<string, { a: number; b: number }>;
const obj: T = {};

R.setPath(obj, ["test", "a"], 2);
```

there's no way to do this type-safely if `obj.test` isn't defined

bombillazo:
Hey, yeah i figured if proper typing is a tenet of this library, it's gonna be hard to infer/produce a dynamic, nested property...

It would be awesome, though. Not sure if one can create a dynamic type from a path, and make a union of the input param generic type with that dynamic type to type the output.

Brodzko:
I stumbled upon a similar problem recently - I think it _technically_ might even be possible to write a function like that type-safely(ish), but there might be too many decisions on how to handle edge-cases to have this on a library level. 
For example, what if the path contains an array index that is non-zero, but its parent property does not exist in target? Do you put the value at index 0 (thus not really upholding contract), or risk creating a sparse array?

I'm not sure there are canonical answers to this and other edge-case questions, so it feels like something that should remain in user-land.

I ended up ignoring array properties (didn't need them) and recursively creating a new object using `objOf` + some type magic to create a type from path in the end.

eranhirsch:
This is actually a bug! setPath can lead to an exception being thrown, and the typing doesn't do enough to protect against it:

```ts
setPath({} as Record<string, { b: number }>, ["a", "b"], 123);
```

PR #715, opened 2024-06-04
fix(set): prevent setting `undefined` on optional props

eranhirsch:
The current type would allow setting an optional prop with `undefined` even when it isn't a valid value for the prop.

I also expanded the typing to make the output more useful by making optional props required and have a fixed value (when possible).

### DIFF
--- a/src/addProp.ts
+++ b/src/addProp.ts
@@ -1,3 +1,4 @@
+import type { UpsertProp } from "./internal/types";
 import { purry } from "./purry";
 
 /**
@@ -19,11 +20,11 @@ import { purry } from "./purry";
  * @dataFirst
  * @category Object
  */
-export function addProp<
-  T extends Record<PropertyKey, unknown>,
-  K extends PropertyKey,
-  V,
->(obj: T, prop: K, value: V): T & { [x in K]: V };
+export function addProp<T, K extends PropertyKey, V>(
+  obj: T,
+  prop: K,
+  value: V,
+): UpsertProp<T, K, V>;
 
 /**
  * Add a new property to an object.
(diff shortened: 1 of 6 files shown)


src/addProp.ts
```diff
@@ -1,8 +1,15 @@
...
 
 /**
  * Add a new property to an object.
  *
+ * The function doesn't do any checks on the input object. If the property
+ * already exists it will be overwritten, and the type of the new value is not
+ * checked against the previous type.
+ *
+ * Use `set` to override values explicitly with better protections.
+ *
```
cjquines, line 12: **thought:** i now wonder if we can add an equivalent to `setPath`, which would fulfill #554, called `addPath`?

src/set.ts
```diff
@@ -1,8 +1,12 @@
+import type { UpsertProp } from "./internal/types";
 import { purry } from "./purry";
 
 /**
  * Sets the `value` at `prop` of `object`.
  *
+ * To add a new property to an object, or to override it's type use `addProp`
```
cjquines, line 7: **nit:**

```suggestion
 * To add a new property to an object, or to override its type, use `addProp`
```

cjquines (approved):

src/addProp.ts
eranhirsch: I think we can overhaul the whole set of functions that replace Lodash's `set` function. All these changes came about while I was writing the migration for Lodash. Some of the changes might be breaking, so we might need to wait for a v3 (which I hope to release sooner, maybe in six months).

I'm happy with the way the new migration docs are turning out because it's obvious that the simple mapping table hid many complexities, which can now be worked on more clearly.

PR #721, opened 2024-06-07
fix: Don't return partial for branded or template keys

### LINKED ISSUE #719: Records with template literal or branded keys return Partial with groupBy
# What

Grouping by branded type and then iterating over entries leads to invalid `undefined` added to the type. 

In Remeda v1 this worked correctly so the behavior has changed.

# Reproduce

```ts
import * as R from 'remeda'

type Branded<K, T> = K & { _type: T }
type UserID = Branded<string, 'UserId'>

const items = [
  { owner: 'U1' as UserID, name: 'Phone' },
  { owner: 'U2' as UserID, name: 'Pencil' },
]

const itemsByOwner = R.groupBy(items, (item) => {
  return item.owner
})

const entries = R.entries(itemsByOwner)

for (const entry of entries) {
  const owner = entry[0]
  const ownersItems = entry[1] // ownersItems can be undefined but it should never be undefined. This seems to be caused by groupBy returning Partial record
  console.log(owner, ownersItems)
}

```

cjquines:
closes #719

### DIFF
--- a/src/groupBy.test.ts
+++ b/src/groupBy.test.ts
@@ -176,6 +176,56 @@ describe("typing", () => {
     >();
   });
 
+  test("symbol", () => {
+    const mySymbol = Symbol("mySymbol");
+    const data = groupBy(
+      [
+        { [mySymbol]: "cat", b: 123 },
+        { [mySymbol]: "dog", b: 456 },
+      ],
+      () => mySymbol,
+    );
+    expectTypeOf(data).toEqualTypeOf<
+      Partial<
+        Record<
+          typeof mySymbol,
+          NonEmptyArray<{ [mySymbol]: string; b: number }>
+        >
+      >
+    >();
+  });
+
+  test("branded string", () => {
+    type Branded<K, T> = K & { _type: T };
+    type UserID = Branded<string, "UserId">;
+
+    const data = groupBy(
+      [
+        { owner: "U1" as UserID, name: "Phone" },
+        { owner: "U2" as UserID, name: "Pencil" },
+      ],
+      (x) => x.owner,
+    );
+    expectTypeOf(data).toEqualTypeOf<
+      Record<UserID, NonEmptyArray<{ owner: UserID; name: string }>>
+    >();
+  });
+
+  test("template string", () => {
+    type UserID = `UserID_${number}`;
+
+    const data = groupBy(
+      [
+        { owner: "UserID_1" as UserID, name: "Phone" },
+        { owner: "UserID_2" as UserID, name: "Pencil" },
+      ],
+      (x) => x.owner,
+    );
+    expectTypeOf(data).toEqualTypeOf<
+      Record<UserID, NonEmptyArray<{ owner: UserID; name: string }>>
+    >();
+  });
+
   describe("Filtering on undefined grouper result", () => {
     test("regular", () => {
       const { even, ...rest } = groupBy([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], (x) =>
--- a/src/groupBy.test.ts
+++ b/src/groupBy.test.ts
@@ -176,56 +176,6 @@ describe("typing", () => {
     >();
   });
 
-  test("symbol", () => {
-    const mySymbol = Symbol("mySymbol");
-    const data = groupBy(
-      [
-        { [mySymbol]: "cat", b: 123 },
-        { [mySymbol]: "dog", b: 456 },
-      ],
-      () => mySymbol,
-    );
-    expectTypeOf(data).toEqualTypeOf<
-      Partial<
-        Record<
-          typeof mySymbol,
-          NonEmptyArray<{ [mySymbol]: string; b: number }>
-        >
-      >
-    >();
-  });
-
-  test("branded string", () => {
-    type Branded<K, T> = K & { _type: T };
-    type UserID = Branded<string, "UserId">;
-
-    const data = groupBy(
-      [
-        { owner: "U1" as UserID, name: "Phone" },
-        { owner: "U2" as UserID, name: "Pencil" },
-      ],
-      (x) => x.owner,
-    );
-    expectTypeOf(data).toEqualTypeOf<
-      Record<UserID, NonEmptyArray<{ owner: UserID; name: string }>>
-    >();
-  });
-
-  test("template string", () => {
-    type UserID = `UserID_${number}`;
-
-    const data = groupBy(
-      [
-        { owner: "UserID_1" as UserID, name: "Phone" },
-        { owner: "UserID_2" as UserID, name: "Pencil" },
-      ],
-      (x) => x.owner,
-    );
-    expectTypeOf(data).toEqualTypeOf<
-      Record<UserID, NonEmptyArray<{ owner: UserID; name: string }>>
-    >();
-  });
-
   describe("Filtering on undefined grouper result", () => {
     test("regular", () => {
       const { even, ...rest } = groupBy([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], (x) =>
--- a/src/internal/types.test.ts
+++ b/src/internal/types.test.ts
@@ -1,99 +1,158 @@
-import { type Branded, type IfSimpleRecord } from "./types";
+import { type Branded, type IfBoundedRecord } from "./types";
 
-describe("IfSimpleRecord", () => {
-  test("string, number, symbol", () => {
-    expectTypeOf<
-      IfSimpleRecord<Record<string, unknown>>
-    >().toEqualTypeOf<true>();
+declare const SymbolFoo: unique symbol;
+declare const SymbolBar: unique symbol;
 
+describe("IfBoundedRecord", () => {
+  test("string", () => {
     expectTypeOf<
-      IfSimpleRecord<Record<number, unknown>>
-    >().toEqualTypeOf<true>();
-
-    expectTypeOf<
-      IfSimpleRecord<Record<number | string, unknown>>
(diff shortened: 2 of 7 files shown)


cjquines:
hm why did the codesandbox build fail?

src/groupBy.test.ts
```diff
@@ -176,6 +176,56 @@ describe("typing", () => {
...
+      [
+        { owner: "UserID_1" as UserID, name: "Phone" },
+        { owner: "UserID_2" as UserID, name: "Pencil" },
+      ],
+      (x) => x.owner,
+    );
+    expectTypeOf(data).toEqualTypeOf<
+      Record<UserID, NonEmptyArray<{ owner: UserID; name: string }>>
+    >();
+  });
```
eranhirsch, line 227: I appreciate that testing on the real functions that use the types makes sense, as these are the interfaces our users encounter, but I think it makes sense to narrow the tests to just `IfSimpleRecord` and test it directly with a test file for types.ts... e.g. `internal/types.test.ts`

src/groupBy.test.ts
```diff
@@ -176,6 +176,56 @@ describe("typing", () => {
     >();
   });
 
+  test("symbol", () => {
+    const mySymbol = Symbol("mySymbol");
+    const data = groupBy(
+      [
+        { [mySymbol]: "cat", b: 123 },
+        { [mySymbol]: "dog", b: 456 },
```
eranhirsch, line 184: I'm confused; why does it matter that the object has a prop with `mySymbol`? isn't the grouping function's return value the only thing we care about here?

src/groupBy.test.ts
```diff
@@ -176,6 +176,56 @@ describe("typing", () => {
...
+      ],
+      () => mySymbol,
```
eranhirsch, line 186: use constant?
```suggestion
      constant(mySymbol),
```

src/groupBy.test.ts
```diff
@@ -176,6 +176,56 @@ describe("typing", () => {
...
+  test("branded string", () => {
+    type Branded<K, T> = K & { _type: T };
+    type UserID = Branded<string, "UserId">;
+
+    const data = groupBy(
+      [
+        { owner: "U1" as UserID, name: "Phone" },
+        { owner: "U2" as UserID, name: "Pencil" },
+      ],
+      (x) => x.owner,
```
eranhirsch, line 207: use prop?
```suggestion
      prop("owner"),
```

src/groupBy.test.ts
```diff
@@ -176,6 +176,56 @@ describe("typing", () => {
...
+          typeof mySymbol,
+          NonEmptyArray<{ [mySymbol]: string; b: number }>
+        >
+      >
+    >();
+  });
+
+  test("branded string", () => {
+    type Branded<K, T> = K & { _type: T };
+    type UserID = Branded<string, "UserId">;
```
eranhirsch, line 200: I prefer pulling out types to the module level, I don't have anything smart for a reason, it is just always a bit jarring to see a type definition inside a test

eranhirsch, line 200: We now have an internal branded type, we can use it.

src/groupBy.test.ts
```diff
@@ -176,6 +176,56 @@ describe("typing", () => {
...
+
+    const data = groupBy(
+      [
+        { owner: "U1" as UserID, name: "Phone" },
+        { owner: "U2" as UserID, name: "Pencil" },
+      ],
```
eranhirsch, line 206: you don't need to define items in the array for type tests; it detracts from what the test is about; you can cast an empty array to whatever type you want to test against...

```suggestion
      [] as Array<{ owner: UserId, name: string }>,
```

eranhirsch (commented):
This is amazing, I didn't think it will be possible to solve the template string cases.

I want to make sure we cover as many cases of template strings we can think of:
- `${"cat" | "dog"}_${number}`
- `prefix_${string}`
- `prefix_${1 | 2}`
etc...

src/internal/types.ts
```diff
@@ -20,25 +29,44 @@ export type Mapped<T extends IterableContainer, K> = {
...
+ */
+type IsBoundedString<T> = T extends string
+  ? Split<T, "">[number] extends infer U
+    ? [`${number}`] extends [U]
+      ? false
+      : [string] extends [U]
+        ? false
+        : true
+    : never
+  : never;
```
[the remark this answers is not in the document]
cjquines: oops this was a mistake, i'll convert to false

src/internal/types.ts
```diff
@@ -20,25 +29,44 @@ export type Mapped<T extends IterableContainer, K> = {
...
+/**
+ * Checks if a type is a bounded key: a union of bounded strings, numeric
+ * literals, or symbol literals.
+ */
+type IsBoundedKey<T> = T extends unknown
+  ? IsStringLiteral<T> extends true
+    ? IsBoundedString<T>
+    : IsNumericLiteral<T> extends true
+      ? true
+      : IsSymbolLiteral<T>
```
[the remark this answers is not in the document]
cjquines: hm… in the case that someone does declare a type `Record<symbol, string>`, i think we'd want it to be a simple record, though? like, in theory, someone could write

```ts
const a = {} as Record<symbol, string>;
// etc.
```

though i don't know why they'd do that.

src/groupBy.test.ts
cjquines: agreed, i think i'll just discard these tests

src/groupBy.test.ts
cjquines: test removed, so this is outdated—i didn't use prop initially for consistency

src/groupBy.test.ts
cjquines: (test removed)

src/groupBy.test.ts
cjquines: (test removed)

cjquines: right, we do now!

src/groupBy.test.ts
cjquines: (test removed)

src/internal/types.ts
```diff
@@ -24,25 +33,54 @@ export type Mapped<T extends IterableContainer, K> = {
...
- * combination). Another way of thinking about simple records is that the number
- * of properties are unbound (they can be empty, or they can have 1000 props).
- *
- * TODO: Template string literals (e.g. `prefix_${number}`) should also be
- * considered simple, but we don't know how to check for them...
+ * Check if a type is a simple record. A simple record is one that has an
+ * unbounded number of properties. For example, a record with a union of string
+ * or number keys is simple, and so is a record with `prefix_${number}` keys.
  */
 export type IfSimpleRecord<
```
eranhirsch, line 40: Now that the difference is the boundness of the key this name isn't very good anymore, it doesn't communicate the difference well. 
```ts
Record<`${MyEnum}_suffix_${number}`, MyObject>
```

isn't any simpler than `{ a: number }`.

src/internal/types.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    >().toEqualTypeOf<true>();
+
+    expectTypeOf<
+      IfSimpleRecord<Record<number | string, unknown>>
+    >().toEqualTypeOf<true>();
+
+    expectTypeOf<
+      IfSimpleRecord<Record<symbol, unknown>>
+    >().toEqualTypeOf<true>();
+  });
```
eranhirsch, line 20: no reason not to split them up, and maybe have a slightly more descriptive name 🙄 ?

eranhirsch (commented):
This is really impressive! Good job!

src/internal/types.test.ts
```diff
@@ -0,0 +1,99 @@
+import { type Branded, type IfSimpleRecord } from "./types";
+
+describe("IfSimpleRecord", () => {
```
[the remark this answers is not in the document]
cjquines: 1 should already be covered? `Record<a, string>` is the same as `{ a: string }`

but good point with 2, i added some tests

src/internal/types.ts
cjquines: renamed to `IfBoundedRecord` (and moved around the args)

src/internal/types.ts
```diff
@@ -24,25 +34,57 @@ export type Mapped<T extends IterableContainer, K> = {
 export type IterableContainer<T = unknown> = ReadonlyArray<T> | readonly [];
 
 /**
- * We define a Simple record as one that doesn't define it's properties beyond
- * their general type. This is the case for string and number keys (and their
- * combination). Another way of thinking about simple records is that the number
- * of properties are unbound (they can be empty, or they can have 1000 props).
+ * Check if a type is guaranteed to be a bounded record: a record with a fixed
```
eranhirsch, line 37: finite?

```suggestion
 * Check if a type is guaranteed to be a bounded record: a record with a finite
```

src/internal/types.ts
```diff
@@ -24,25 +34,57 @@ export type Mapped<T extends IterableContainer, K> = {
...
+ * set of keys.
  *
- * TODO: Template string literals (e.g. `prefix_${number}`) should also be
- * considered simple, but we don't know how to check for them...
+ * For example, a record with a union of string or number keys is unbounded,
```
eranhirsch, line 40: 1. Would this be better as a JSDoc `@example`? 
2. Would this be clearer in code vs words (e.g. `Record<string, unknown>`)
3. Why start with a complex example (union) instead of the trivial one?

eranhirsch (approved):

PR #722, opened 2024-06-07
fix(pipe): early exit on hasMany

kosmotema:
Currently when using `pipe`, `take` does nothing if it is followed by `flat` or `flatMap`. Here is a small example:

```ts
import * as R from 'remeda';

console.log(
  R.pipe(
    [
      [1, 2],
      [3, 4],
      [5, 6],
    ],
    R.take(1),
    R.flat()
  )
);
```

This code should output to the console `[1, 2]`, but actually outputs `[1, 2, 3, 4, 5, 6]`
This is because using `take` causes `isDone` to change to `true`, but `flat` (or any other function with `hasMany` in its iterator) has an early return of `false`, that ignores the value of `isDone` (so, instead of taking only first pair in the example code, `pipe` will take all pairs)

---

Make sure that you:

- [ ] New methods added to `docs/src/pages/mapping.md`

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/pipe.test.ts
+++ b/src/pipe.test.ts
@@ -1,4 +1,6 @@
 import { filter } from "./filter";
+import { flat } from "./flat";
+import { flatMap } from "./flatMap";
 import { identity } from "./identity";
 import { map } from "./map";
 import { pipe } from "./pipe";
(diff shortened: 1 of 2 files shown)


src/pipe.test.ts
```diff
@@ -115,4 +117,39 @@ describe("lazy", () => {
...
+        [
+          [7, 8],
+          [8, 9],
+        ],
+      ],
+      take(2),
+      flatMap((x) => x[1]),
+    );
+    expect(result).toEqual([2, 3, 5, 6]);
+  });
```
eranhirsch, line 154: This test doesn't add to the previous test; we don't need both, the problem was not `flat` or `flatMap`, it was the `hasMany`.

```suggestion
```

src/pipe.test.ts
```diff
@@ -115,4 +117,39 @@ describe("lazy", () => {
     expect(count2).toHaveBeenCalledTimes(2);
     expect(result).toEqual([100, 200]);
   });
+
+  it("lazy take + flat", () => {
```
eranhirsch, line 121: Can you rename the test to reflect the issue better, take and flat are just a way to reproduce it. I'd call this "lazy early exit with hasMany"

eranhirsch (approved):
Great catch and great fix! thank you!

src/pipe.test.ts
kosmotema: Removed the test

src/pipe.test.ts
kosmotema: Renamed it according to your suggestion

eranhirsch (approved):

PR #724, opened 2024-06-08
chore(codesandbox): Fix CI errors

eranhirsch:
Trying to fix codesandbox failing our builds. From the docs it seems that the plugin doesn't need any config to work. Maybe our config has grown old.


eranhirsch:
I don't know what's breaking the build. There's no error message in the logs, and I couldn't bisect any change we made on our side.

cc: @cjquines, @TkDodo?

TkDodo:
We have the same issue in query. I reached out to the csb team:
- https://x.com/tkdodo/status/1799064502132220131

PR #728, opened 2024-06-11
fix: key and value typing of unions of object types.

### LINKED ISSUE #727: `mapKeys` with union input record gives result with `never` values
here's a failing test

```ts
  test("union of records", () => {
    const result = mapKeys(
      {} as Record<number, string> | Record<string, string>,
      constant("hello" as string),
    );
    expectTypeOf(result).toEqualTypeOf<Record<string, string>>();
  });
```

the actual type of `result` right now is `Record<string, never>`

grigorischristainas:
Closes #727.

I thought it was an inference issue so I tried to resolve it by modifying the types and using `infer`. Moreover, I tried to keep using the internal types `EnumerableStringKeyOf` and `EnumerableStringKeyedValueOf`. 

I would be happy to integrate any further adjustments as needed. 

---

Make sure that you:

- [ ] New methods added to `docs/src/pages/mapping.md`

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/forEachObj.test.ts
+++ b/src/forEachObj.test.ts
@@ -70,4 +70,22 @@ describe("typing", () => {
       expectTypeOf(value).toEqualTypeOf<boolean | string>();
     });
   });
+
+  test("union of records", () => {
+    forEachObj(
+      {} as Record<number, string> | Record<string, number>,
+      (value, key) => {
+        expectTypeOf(key).toEqualTypeOf<string>();
+        expectTypeOf(value).toEqualTypeOf<number | string>();
+      },
+    );
+
+    pipe(
+      {} as Record<number, string> | Record<string, number>,
+      forEachObj((value, key) => {
+        expectTypeOf(key).toEqualTypeOf<string>();
+        expectTypeOf(value).toEqualTypeOf<number | string>();
+      }),
+    );
+  });
 });
--- a/src/forEachObj.test.ts
+++ b/src/forEachObj.test.ts
@@ -72,16 +72,15 @@ describe("typing", () => {
   });
 
   test("union of records", () => {
-    forEachObj(
-      {} as Record<number, string> | Record<string, number>,
-      (value, key) => {
-        expectTypeOf(key).toEqualTypeOf<string>();
-        expectTypeOf(value).toEqualTypeOf<number | string>();
-      },
-    );
+    const data = {} as Record<number, string> | Record<string, number>;
+
+    forEachObj(data, (value, key) => {
+      expectTypeOf(key).toEqualTypeOf<string>();
+      expectTypeOf(value).toEqualTypeOf<number | string>();
+    });
 
     pipe(
-      {} as Record<number, string> | Record<string, number>,
+      data,
       forEachObj((value, key) => {
         expectTypeOf(key).toEqualTypeOf<string>();
         expectTypeOf(value).toEqualTypeOf<number | string>();
--- a/src/internal/types.test.ts
+++ b/src/internal/types.test.ts
@@ -1,4 +1,10 @@
-import { type Branded, type IfBoundedRecord } from "./types";
+import { type EmptyObject } from "type-fest";
+import {
+  type Branded,
+  type IfBoundedRecord,
+  type EnumerableStringKeyedValueOf,
+  type EnumerableStringKeyOf,
+} from "./types";
 
 declare const SymbolFoo: unique symbol;
 declare const SymbolBar: unique symbol;
@@ -154,5 +160,149 @@ describe("IfBoundedRecord", () => {
         Record<`${string}_${number}_${string}_${number}_${string}`, unknown>
       >
     >().toEqualTypeOf<false>();
+  });
+});
+
+describe("EnumerableStringKeyOf", () => {
+  test("string keys", () => {
+    expectTypeOf<
+      EnumerableStringKeyOf<Record<string, unknown>>
+    >().toEqualTypeOf<string>();
+  });
+
+  test("number keys", () => {
+    expectTypeOf<
+      EnumerableStringKeyOf<Record<number, unknown>>
+    >().toEqualTypeOf<`${number}`>();
+  });
+
+  test("union of records", () => {
+    expectTypeOf<
+      EnumerableStringKeyOf<
+        Record<`prefix_${string}`, unknown> | Record<number, unknown>
+      >
+    >().toEqualTypeOf<`${number}` | `prefix_${string}`>();
+  });
+
+  test("union keys", () => {
+    expectTypeOf<
+      EnumerableStringKeyOf<Record<number | `prefix_${string}`, unknown>>
+    >().toEqualTypeOf<`${number}` | `prefix_${string}`>();
+  });
+
+  test("symbol keys", () => {
+    expectTypeOf<
+      EnumerableStringKeyOf<Record<string | symbol, unknown>>
+    >().toEqualTypeOf<string>();
+
+    expectTypeOf<
+      EnumerableStringKeyOf<{ [SymbolFoo]: number; a: unknown }>
+    >().toEqualTypeOf<"a">();
+
+    expectTypeOf<
+      EnumerableStringKeyOf<Record<string | typeof SymbolFoo, unknown>>
+    >().toEqualTypeOf<string>();
+  });
+
+  test("optional keys", () => {
+    expectTypeOf<
+      EnumerableStringKeyOf<{ a: unknown; b?: unknown }>
+    >().toEqualTypeOf<"a" | "b">();
+  });
+
+  test("branded types", () => {
+    expectTypeOf<
+      EnumerableStringKeyOf<Record<Branded<string, symbol>, unknown>>
+    >().toEqualTypeOf<`${Branded<string, symbol>}`>();
+  });
+});
+
+describe("EnumerableStringKeyedValueOf", () => {
+  test("string values", () => {
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<Record<PropertyKey, string>>
+    >().toEqualTypeOf<string>();
+  });
+
+  test("number values", () => {
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<Record<PropertyKey, number>>
+    >().toEqualTypeOf<number>();
+  });
+
+  test("union of records", () => {
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<
+        Record<PropertyKey, "cat"> | Record<PropertyKey, "dog">
+      >
+    >().toEqualTypeOf<"cat" | "dog">();
+
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<
+        Record<PropertyKey, number> | Record<PropertyKey, string>
+      >
+    >().toEqualTypeOf<number | string>();
+  });
+
+  test("union values", () => {
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<Record<PropertyKey, number | string>>
+    >().toEqualTypeOf<number | string>();
+  });
+
+  test("literal values", () => {
+    expectTypeOf<EnumerableStringKeyedValueOf<{ a: 1 }>>().toEqualTypeOf<1>();
+
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<{ a: "1" | "2" | 1 }>
+    >().toEqualTypeOf<"1" | "2" | 1>();
+  });
+
+  test("optional values", () => {
+    expectTypeOf<EnumerableStringKeyedValueOf<{ a: 1; b?: 4 }>>().toEqualTypeOf<
+      1 | 4
+    >();
+
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<{ a: string; b?: number }>
+    >().toEqualTypeOf<number | string>();
+  });
+
+  test("nullish and undefined values", () => {
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<{
+        a: string | undefined;
+        b: string | null;
+      }>
+    >().toEqualTypeOf<string | null | undefined>();
+
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<{
+        a?: number | null;
+        b?: number | null | undefined;
+      }>
+    >().toEqualTypeOf<number | null | undefined>();
+  });
+
+  test("symbol keys", () => {
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<{ [SymbolFoo]: string }>
+    >().toEqualTypeOf<never>();
+
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<{ [SymbolFoo]: string; b: "1" }>
+    >().toEqualTypeOf<"1">();
+
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<
+        Record<PropertyKey | typeof SymbolFoo, string>
+      >
+    >().toEqualTypeOf<string>();
+  });
+
+  test("empty object", () => {
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<EmptyObject>
+    >().toEqualTypeOf<never>();
   });
 });
--- a/src/internal/types.ts
+++ b/src/internal/types.ts
@@ -8,6 +8,7 @@ import {
   type KeysOfUnion,
   type Simplify,
   type Split,
+  type EmptyObject,
 } from "type-fest";
 
 declare const __brand: unique symbol;
@@ -96,15 +97,25 @@ type IsBoundedString<T> = T extends string
  *
  * @see EnumerableStringKeyedValueOf
  */
-export type EnumerableStringKeyOf<T> = `${Exclude<keyof T, symbol>}`;
+export type EnumerableStringKeyOf<T> =
+  T extends Record<infer K, unknown> ? `${Exclude<K, symbol>}` : never;
+
+/**
+ * Extracts the value type from an object type T.
+ */
+export type InferredRecordValue<T> = T extends EmptyObject
+  ? T[keyof T]
+  : T extends Record<PropertyKey, infer V>
+    ? V
+    : never;
 
 /**
  * A union of all values of properties in T which are not keyed by a symbol,
  * following the definition of `Object.values` and `Object.entries`.
  */
-export type EnumerableStringKeyedValueOf<T> = {
+export type EnumerableStringKeyedValueOf<T> = InferredRecordValue<{
   [K in keyof T]-?: K extends symbol ? never : T[K];
-}[keyof T];
+}>;
 
 /**
  * This is the type you'd get from doing:
--- a/src/internal/types.ts
+++ b/src/internal/types.ts
@@ -103,9 +103,17 @@ export type EnumerableStringKeyOf<T> =
     : never;
 
 /**
+ * A union of all values of properties in T which are not keyed by a symbol,
+ * following the definition of `Object.values` and `Object.entries`.
+ */
+export type EnumerableStringKeyedValueOf<T> = ValuesOf<{
+  [K in keyof T]-?: K extends symbol ? never : T[K];
+}>;
+
+/**
  * Extracts the value type from an object type T.
  */
-export type ValuesOf<T> =
+type ValuesOf<T> =
   // EmptyObject is used to infer value type as never instead of unknown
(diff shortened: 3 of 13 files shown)


cjquines (changes requested):
thanks for the pr!

i suspect we have a similar issue with other places where we use `EnumerableStringKeyedValueOf`, which i think are `pickBy`, `omitBy`, `mapValues`, `forEachObj`, and `values`. for example

```ts
  test("union of records", () => {
    forEachObj(
      {} as Record<number, string> | Record<string, number>,
      (value, key) => {
        expectTypeOf(key).toEqualTypeOf<string | `${number}`>();
        expectTypeOf(value).toEqualTypeOf<number | string>();
      },
    );
  });
```

i suspect the correct fix would have to be at the `EnumerableStringKeyOf` and `EnumearableStringKeyedValueOf` level?

grigorischristainas:
Hi @cjquines, I have committed the changes and added some tests. Could you please review them and let me know if any further adjustments are necessary?

src/internal/types.ts
```diff
@@ -96,15 +97,25 @@ type IsBoundedString<T> = T extends string
...
  */
-export type EnumerableStringKeyOf<T> = `${Exclude<keyof T, symbol>}`;
+export type EnumerableStringKeyOf<T> =
+  T extends Record<infer K, unknown> ? `${Exclude<K, symbol>}` : never;
+
+/**
+ * Extracts the value type from an object type T.
+ */
+export type InferredRecordValue<T> = T extends EmptyObject
+  ? T[keyof T]
```
cjquines, line 107: **question:** why is there a `T extends EmptyObject` check here? what goes wrong without it?

can you add a comment for this

src/mapValues.test.ts
```diff
@@ -116,7 +116,7 @@ describe("typing", () => {
 
       mapValues(userValues, (value, key) => {
         expectTypeOf(value).toEqualTypeOf<number>();
-        expectTypeOf(key).toEqualTypeOf<`${UserID}`>();
+        expectTypeOf(key).toMatchTypeOf<`${UserID}`>();
```
cjquines, line 119: **question:** why doesn't `toEqualTypeOf` work here?

i think units should always assert the strongest possible assertion, and in this case i can't see how `key` can be narrower than `${UserID}`

src/omitBy.ts
```diff
@@ -19,15 +19,19 @@ type PickSymbolKeys<T extends object> = {
...
-type PartialEnumerableKeys<T extends object> = Simplify<
-  IfBoundedRecord<
-    T,
-    PickSymbolKeys<T> & {
-      -readonly [P in keyof T as P extends symbol ? never : P]?: Required<T>[P];
-    },
-    ReconstructedRecord<T>
-  >
->;
+type PartialEnumerableKeys<T extends object> = T extends unknown
```
cjquines, line 22: **nit:** whenever we have to do the `T extends unknown` trick, we should add the distributive conditional comment, i think? see definition for `IsBoundedKey`

src/internal/types.ts
```diff
@@ -96,15 +97,25 @@ type IsBoundedString<T> = T extends string
  *
  * @see EnumerableStringKeyedValueOf
  */
-export type EnumerableStringKeyOf<T> = `${Exclude<keyof T, symbol>}`;
+export type EnumerableStringKeyOf<T> =
+  T extends Record<infer K, unknown> ? `${Exclude<K, symbol>}` : never;
```
cjquines, line 101: **suggestion:** let's add tests for this and `EnumerableStringKeyedValueOf` to `types.test.ts`

testing would be wide cases (`Record<string, string>`, `Record<string | number, string>`, etc.), literals (`{ a: 1 }`, etc.), union types (like we have here), things with `symbol` (which is the intent behind this type in the first place), things with optional properties

cjquines (dismissed):
this looks awesome, thanks for the thoroughness here! i've got one suggestion to add more tests to `types.test.ts`, but i'm fine with shipping this as-is if you don't have time to add some.

src/internal/types.ts
grigorischristainas: This was added mainly due to a failing test case for `values`: 
```ts
it("should correctly type functions", () => {
    const result = values(doNothing());
    expectTypeOf(result).toEqualTypeOf<Array<never>>();
});
```
Without the check for `EmptyObject`, result is typed as `uknown[]` instead of `never[]`. I agree that it is a bit confusing and a comment would be useful. If you have a better suggestion, I would be happy to integrate it; otherwise, I will proceed with adding a comment.

src/mapValues.test.ts
grigorischristainas: I got an error when using `toEqualTypeOf`, which I suspect has to do with type inference (?). I got it to work with `assertType` but it is not being used in codebase, so I changed it with `toMatchTypeOf`. Have you encountered a similar issue before ?

src/omitBy.ts
grigorischristainas: Sure, I will add the distributive conditional comment, it will be useful for future reference!

src/internal/types.ts
grigorischristainas: I agree, this would be very useful. I will add them in `types.test.ts` within the next few days.

src/mapValues.test.ts
cjquines: strange indeed… i'm fine with this for now

src/internal/types.ts
cjquines: makes sense, let's comment

src/internal/types.ts
```diff
@@ -96,15 +97,25 @@ type IsBoundedString<T> = T extends string
...
+
+/**
+ * Extracts the value type from an object type T.
+ */
+export type InferredRecordValue<T> = T extends EmptyObject
```
eranhirsch, line 106: The "Inferred" part in the name does not describe what this type is; it's how it's calculated. This doesn't communicate clearly what this type is. Similarly, Records, conceptually, are a subset of all objects for simple mappings. It's true that when doing complex typing, it's a nice shorthand for more complicated syntax, but it's not the intent of the term.

Naming stuff is often a matter of taste, so I'm not putting my foot down on this, but having read a lot of old code to try and detangle Remeda in the first place, a good name goes a long way to simplify long-term maintenance.

It seems like the correct name here would be `Value` or `ValuesOf`, as in `ValuesOf<MyType>`

src/keys.test.ts
```diff
@@ -111,6 +112,25 @@ describe("typing", () => {
       expectTypeOf(result).toEqualTypeOf<[]>();
     });
 
+    test("unions of records", () => {
+      const dataFirst = keys(
+        {} as Record<number, string> | Record<string, string>,
```
eranhirsch, line 117: You should use `unknown` for the values type, to make it even more general.
```suggestion
        {} as Record<number, unknown> | Record<string, unknown>,
```

src/keys.test.ts
```diff
@@ -111,6 +112,25 @@ describe("typing", () => {
...
+      );
+
+      expectTypeOf(dataFirst).toEqualTypeOf<
+        Array<`${number}`> | Array<string>
+      >();
+
+      const dataLast = pipe(
+        {} as Record<number, string> | Record<string, string>,
```
eranhirsch, line 125: You can pull out this as a data variable and reuse it in the 2 tests instead of repeating the type. That way, if someone changes the type for one test, they don't drift accidentally.

```suggestion
        data,
```

src/forEachObj.test.ts
```diff
@@ -70,4 +70,22 @@ describe("typing", () => {
       expectTypeOf(value).toEqualTypeOf<boolean | string>();
     });
   });
+
+  test("union of records", () => {
+    forEachObj(
+      {} as Record<number, string> | Record<string, number>,
+      (value, key) => {
+        expectTypeOf(key).toEqualTypeOf<string>();
```
eranhirsch, line 78: wouldn't the key here need to be:
```ts
`${number}` | string
```

Is that being squished by TypeScript because they are both strings?

src/forEachObj.test.ts
```diff
@@ -70,4 +70,22 @@ describe("typing", () => {
...
+        expectTypeOf(value).toEqualTypeOf<number | string>();
+      },
+    );
+
+    pipe(
+      {} as Record<number, string> | Record<string, number>,
```
eranhirsch, line 84: Similar comments as above, keep a single "data" variable with the type so that both tests are synced and we don't risk changing one without changing the other.
```suggestion
      data,
```

src/keys.ts
```diff
@@ -50,7 +50,9 @@ type IndicesAfterSpread<
       : Iterations["length"];
 
 type ObjectKeys<T> =
-  T extends Record<PropertyKey, never> ? [] : Array<EnumerableStringKeyOf<T>>;
+  T extends Record<PropertyKey, never>
+    ? []
+    : Array<EnumerableStringKeyOf<Required<T>>>;
```
eranhirsch, line 55: Shouldn't the `Required` wrapper part be part of the `EnumerableStringKeyOf` type and not here?

src/mapKeys.test.ts
```diff
@@ -105,4 +105,20 @@ describe("typing", () => {
     const result = mapKeys({ a: "b" }, constant(123));
     expectTypeOf(result).toEqualTypeOf<Partial<Record<123, string>>>();
   });
+
+  test("union of records", () => {
+    const dataFirst = mapKeys(
+      {} as Record<number, string> | Record<string, string>,
```
eranhirsch, line 111: 1. similar to previous comments, extract a common `data` 
variable.
2. We don't care about the value type, we should use `unknown` instead of `string` to keep the test generic.
```suggestion
      data,
```

src/mapValues.test.ts
eranhirsch: Let's add a TODO so we can test this out separately without delaying this PR

src/mapValues.test.ts
```diff
@@ -160,4 +160,24 @@ describe("typing", () => {
       d: boolean;
     }>();
   });
+
+  test("unions of records", () => {
+    const dataFirst = mapValues(
+      {} as Record<number, string> | Record<string, string>,
```
eranhirsch, line 166: similar to previous comments: 1. pull a `data` variable out, and use `unknown` in places we don't care about a specific type, like the values here.

src/omitBy.ts
```diff
@@ -19,15 +19,19 @@ type PickSymbolKeys<T extends object> = {
...
+        T,
+        PickSymbolKeys<T> & {
+          -readonly [P in keyof T as P extends symbol
+            ? never
+            : P]?: Required<T>[P];
+        },
+        ReconstructedRecord<T>
+      >
+    >
+  : unknown;
```
eranhirsch, line 34: Shouldn't this be `never`?

src/values.test.ts
```diff
@@ -53,6 +53,13 @@ describe("typing", () => {
     expectTypeOf(result).toEqualTypeOf<Array<boolean>>();
   });
 
+  it("should correctly type union of Records", () => {
+    const result = values<Record<number, string> | Record<string, string>>({
+      a: "b",
+    });
```
eranhirsch, line 59: Don't use an explicit generic of `values`, define the type for the data itself.
```suggestion
    const result = values({
      a: "b",
    } as Record<number, string> | Record<string, string>);
```

eranhirsch (commented):
I just have a few nits about styling and test ergonomics.

src/internal/types.ts
grigorischristainas: Thank you for your suggestion, `ValuesOf` is indeed a more precise name.

src/forEachObj.test.ts
grigorischristainas: I think it is indeed being squished by TypeScript as the first one is a subset of string. I tried this example below:
```ts
type A = `${number}`
type B = string
type C = A | B // C is of type string
```

src/keys.ts
grigorischristainas: You are right thank you, I will integrate it in `EnumerableStringKeyOf`.

src/keys.test.ts
grigorischristainas: I agree it would be better to extract a common `data` variable. I will integrate this change to the rest of the tests too, as you suggested.

grigorischristainas:
@cjquines @eranhirsch thank you both for your feedback! I have integrated the requested changes and additionally added tests for internal types `EnumerableStringKeyOf` and `EnumerableStringKeyedValueOf`. I would be happy to integrate any additional adjustments if needed.

src/internal/types.ts
```diff
@@ -8,11 +8,12 @@ import {
   type KeysOfUnion,
   type Simplify,
   type Split,
+  type EmptyObject,
 } from "type-fest";
 
 declare const __brand: unique symbol;
 
-export type Branded<T, Brand extends symbol> = T & { [__brand]: Brand };
+export type Branded<T, Brand extends PropertyKey> = T & { [__brand]: Brand };
```
eranhirsch, line 16: Why the change? We want to limit the usage of branded types as much as possible, but this opens up the usage to regular strings...

src/internal/types.ts
```diff
@@ -96,15 +97,30 @@ type IsBoundedString<T> = T extends string
...
-export type EnumerableStringKeyOf<T> = `${Exclude<keyof T, symbol>}`;
+export type EnumerableStringKeyOf<T> =
+  Required<T> extends Record<infer K, unknown>
+    ? `${Exclude<K, symbol>}`
+    : never;
+
+/**
+ * Extracts the value type from an object type T.
+ */
+export type ValuesOf<T> =
```
eranhirsch, line 108: This isn't used outside this file so it shouldn't be exported. We also usually put non-exported "private" types below the exported ones so it's easier to read through the file.

src/mapKeys.test.ts
```diff
@@ -105,4 +105,14 @@ describe("typing", () => {
...
+
+  test("union of records", () => {
+    const data = {} as Record<number, string> | Record<string, string>;
+
+    const dataFirst = mapKeys(data, constant("hello" as string));
+    expectTypeOf(dataFirst).toEqualTypeOf<Record<string, string>>();
+
+    const dataLast = pipe(data, mapKeys(constant("hello" as string)));
+    expectTypeOf(dataLast).toEqualTypeOf<Record<string, string>>();
+  });
```
eranhirsch, line 117: I don't think this test tests what this PR fixes. Because you use constant for the new key, there's nothing testing the result of the union. You should test different value types (and the key doesn't matter): `Record<PropertyKey, "cat"> | Record<PropertyKey, "dog">`, or test the mapper itself like you test in forEach.

src/mapValues.test.ts
```diff
@@ -106,7 +107,6 @@ describe("typing", () => {
 
   describe("branded types", () => {
     test("should infer types correctly in the mapper", () => {
-      type Branded<K, T> = K & { _type: T };
```
eranhirsch, line 109: 👍🏼

src/values.test.ts
```diff
@@ -53,6 +53,14 @@ describe("typing", () => {
     expectTypeOf(result).toEqualTypeOf<Array<boolean>>();
   });
 
+  it("should correctly type union of Records", () => {
+    const result = values({
+      a: "b",
+    } as Record<number, string> | Record<string, string>);
+
+    expectTypeOf(result).toEqualTypeOf<Array<string>>();
```
eranhirsch, line 61: This doesn't test the main issue this PR deals with clearly enough; the records should have different value types so that we can see that the result is the correct union of them; using string for both just skips the complexity, and it doesn't matter what the key is.
```suggestion
    const result = values({
      a: "b",
    } as Record<PropertyKey, "cat"> | Record<PropertyKey, "dog">);

    expectTypeOf(result).toEqualTypeOf<Array<string>>();
```

src/internal/types.test.ts
```diff
@@ -156,3 +161,144 @@ describe("IfBoundedRecord", () => {
...
+
+  test("number keys", () => {
+    expectTypeOf<
+      EnumerableStringKeyOf<Record<number, unknown>>
+    >().toEqualTypeOf<`${number}`>();
+  });
+
+  test("union of records", () => {
+    expectTypeOf<
+      EnumerableStringKeyOf<Record<number, unknown> | Record<string, unknown>>
```
eranhirsch, line 180: Instead of testing number and string, which cause typescript to squish the result type, you can use a template strings which won't be squished with the number string:

```suggestion
      EnumerableStringKeyOf<Record<number, unknown> | Record<`prefix_${string}`, unknown>>
```

src/internal/types.test.ts
```diff
@@ -156,3 +161,144 @@ describe("IfBoundedRecord", () => {
...
+    >().toEqualTypeOf<string>();
+  });
+
+  test("union keys", () => {
+    expectTypeOf<
+      EnumerableStringKeyOf<Record<number | string, unknown>>
```
eranhirsch, line 186: similarly.

src/internal/types.test.ts
```diff
@@ -156,3 +161,144 @@ describe("IfBoundedRecord", () => {
...
+    expectTypeOf<
+      EnumerableStringKeyOf<Record<Branded<string, symbol>, unknown>>
+    >().toEqualTypeOf<`${Branded<string, symbol>}`>();
+  });
+});
+
+describe("EnumerableStringKeyedValueOf", () => {
+  test("string values", () => {
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<Record<number, string>>
```
eranhirsch, line 220: when you don't care about the key use PropertyKey and not number/string/etc...
```suggestion
      EnumerableStringKeyedValueOf<Record<PropertyKey, string>>
```

src/internal/types.test.ts
```diff
@@ -156,3 +161,144 @@ describe("IfBoundedRecord", () => {
...
+  test("number values", () => {
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<Record<string, number>>
+    >().toEqualTypeOf<number>();
+  });
+
+  test("union of records", () => {
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<
+        Record<number, string> | Record<string, string>
```
eranhirsch, line 233: Similar to my comment in one of the specific tests, this doesn't test the union logic very well because both records use the same value type. Instead, we should test two disjoint value types.
```suggestion
        Record<number, "cat"> | Record<string, "dog">
```

src/internal/types.test.ts
```diff
@@ -156,3 +161,144 @@ describe("IfBoundedRecord", () => {
...
+      EnumerableStringKeyedValueOf<{ [SymbolFoo]: string; b: "1" }>
+    >().toEqualTypeOf<"1">();
+
+    expectTypeOf<
+      EnumerableStringKeyedValueOf<Record<string | typeof SymbolFoo, string>>
+    >().toEqualTypeOf<string>();
+  });
+
+  test("empty object", () => {
+    const emptyObj = {} as const;
```
eranhirsch, line 299: Doesn't type-fest already provide a type for this?

eranhirsch (changes requested):
Some tests don't test the core of the issue very well. I left some comments to address that.

src/internal/types.ts
grigorischristainas: I see, I changed that to also support a string, as in the related test in `mapValues`. Maybe I could just change the type in the test from:
```ts
type UserID = Branded<string, "UserId">;
```
to
```ts
type UserID = Branded<string, symbol>;
```
and revert that change in `Branded` type.

src/internal/types.test.ts
grigorischristainas: Are you referring to the `EmptyObject` type? Sure, I could use that one, but it is also used in the `ValuesOf` type. So I thought I might utilize it this way, as I've seen it used in other tests. I could replace it with the `EmptyObject`.

grigorischristainas:
Thank you @eranhirsch, I rebased and integrated the changes for types tests and the `Branded` type in the previous commits [eab3445](https://github.com/remeda/remeda/pull/728/commits/eab344508aaf421b1c097e40eccc26a3d84b4111) & [4ccfddf](https://github.com/remeda/remeda/pull/728/commits/4ccfddfb798de1b236b6bcda4957afb6dadfcd4c). I added the rest of the changes in new commits. Please let me know if any additional adjustments are needed.

eranhirsch (approved):
This is really good work! thanks for doing it

PR #733, opened 2024-06-18
chore(readme): Migration Guides Links + New Badges

eranhirsch:


cjquines (approved):
do you think that's too many badges? idk but yeah

eranhirsch:
I don't know how people use the badges 🙃 so I don't know how to optimize this experience.

I reviewed several similar projects, and this is the result of taking the best out of each. I looked for stuff that communicates our main principles and the quality of the project at a glance.

The only one I'm not 100% behind is the number of files in the project, which, IMO does not translate into any actionable insight from viewers. wdyt?

PR #746, opened 2024-06-21
chore: add `CONTRIBUTING.md`

cjquines:
quite opinionated (especially in the philosophy section), so push back if you disagree on anything or if you want something added

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,8 +5,8 @@ We're always looking for new contributors and are happy to help you get started!
 - Add a test to a function, testing something that we haven't tested yet.
 - Improve a function's documentation.
 - Fix a bug or an edge case.
-- Work on [![issues labeled good first issue](https://img.shields.io/github/issues/remeda/remeda/good%20first%20issue?style=flat-square)](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%good+first+issue%22)
-- Work on [![issues labeled help wanted](https://img.shields.io/github/issues/remeda/remeda/help%20wanted?style=flat-square)](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%help+wanted%22)
+- Work on [![issues labeled good first issue](https://img.shields.io/github/issues/remeda/remeda/good%20first%20issue?style=flat-square)](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22)
+- Work on [![issues labeled help wanted](https://img.shields.io/github/issues/remeda/remeda/help%20wanted?style=flat-square)](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22)
 
 ## Getting started
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,10 @@ npm test -i src/file.test.ts
 
 We have a pre-commit script that will format and lint the project.
 
+## Documentation
+
+Help us make the documentation better! See [`docs/README.md`](https://github.com/remeda/remeda/blob/main/docs/README.md).
+
 ## Guidelines
 
 ### Adding a new function
@@ -98,20 +102,14 @@ Most of these types will be irrelevant! Use your judgment to pick the ones that 
 - `readonly` versions of the above
 - `null` and `undefined`
 
-### Philosophy
+### Design philosophy
 
-**Type annotations.** Functions shouldn't require type annotations to have good types. This gives the best developer experience. The exception is `piped`.
+**No type annotations.** Functions shouldn't require type annotations to have good types. This gives the best developer experience. The exception is `piped`.
 
 **Have good output types.** One reason Remeda provide "one-liner" functions is if we can have better output types than the defaults. While `data.filter(item => typeof item === 'number')` works, `filter(data, isNumber)` gives a better output type. This is why we have type guards, `hasAtLeast`, `prop`, etc.
 
-This 
-
 **Readability counts.** The other reason Remeda provides one-liner functions is if they make things more readable. Compare `map((item) => item + 3)` to `map(add(3))`, where we don't need to come up with a name for `item`. This is why we have `constant`, `doNothing`, `identity`, etc.
 
-**Prefer composition.** We don't provide simple compositions of functions, because we prefer users 
+**Prefer composition.** We don't provide simple compositions of functions. Having a smaller surface area makes it easier for users to remember Remeda's functions, and easier for us to maintain. We don't have `reject` because it's the same as `filter(isNot)`. An exception is when the composition would be less readable; this is why we have `sum`, even if it's the same as `sumBy(identity)`.
 
-An exception is when the composition would be less readable. This is why we have `sum`, even if it's the same as `sumBy(identity)`.
-
-## Documentation
-
-Help us make the documentation better! See [`docs/README.md`](https://github.com/remeda/remeda/blob/main/docs/README.md).
+**Minimal implementations.** Implementations should, as much as possible, be small and self-contained. Importing an individual function should add less than 500 B to the bundle size in most cases.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,117 @@
+# Contributing guide
+
+We're always looking for new contributors and are happy to help you get started! Some ideas:
+
+- Add a test for a function.
+- Improve a function's documentation.
+- Work on any issue tagged [good first issue](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22).
+- Work on any issue tagged [help wanted](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22).
+
+## Getting started
+
+Requirements:
+
+- Node.js `>=18.20.0`
+- npm `>=10.7.0`
+
+Install dependencies:
+
+```bash
+git clone git@github.com:remeda/remeda.git
+cd remeda
+npm install
+```
+
+Run a test, and watch for changes:
+
+```bash
+npm test -i src/file.test.ts
+```
+
+We have a pre-commit script that will format and lint the project.
+
+## Guidelines
+
+### Adding a new function
+
+When adding a new function, remember to:
+
+- Add a Typedoc comment with a description, parameters, signature, an example, and tags.
+- Add tests for runtime, typing, data-first, and data-last forms.
+  - For data-last tests, prefer tests using `pipe`, as this better matches real-world usage.
+- Add exports to `src/index.ts`.
+- Add Lodash and Ramda equivalents to `docs/src/content/mapping`.
+
+### Input types to consider
+
+Check which of these types are relevant and write tests for them.
+
+Most of these types will be irrelevant! Use your judgment to pick the ones that are likely to come up in a real use case.
+
+- Numbers
+  - Number type (`number`)
+  - Single literal (`1`)
+  - Union of literals (`1 | 2`)
+  - Bigint type (`bigint`)
+  - Single bigint literal (`1n`)
+  - Union of bigint literals (`1n | 2n`)
+  - Unions of numbers and bigints (`1 | 2n`)
+- Strings
+  - String type (`string`)
+  - Single literal (`"cat"`)
+  - Union of literals (`"cat" | "dog"`)
+  - Template with a type slot (`` `id_${number}` ``)
+  - Template with a literal slot (`` `id_${1 | 2}` ``)
+  - Template with multiple slots (`` `id_${1 | 2}_${3 | 4}` ``)
+- Arrays
+  - Array of a single type (`Array<number>`)
+  - Array of a union type (`Array<string | number>`)
+  - Array of literal types (`Array<"cat" | "dog">`)
+  - Union of similar arrays (`Array<1 | 2> | Array<2 | 3>`)
+  - Nested arrays (`Array<Array<number>>`)
+- Tuples
+  - Tuple of a single type (`[number, number, number]`)
+  - Tuple of different types (`[number, string, boolean]`)
+  - Tuple with optional type (`[number, string?]`)
+  - Tuple with spreads (`[...Array<number>, number, number]`)
+  - Tuple with optional type and spread (`[number?, ...Array<string>]`)
+  - Tuple with union values (`[number, string | undefined, boolean]`)
+  - Tuple with literal values (`[number, "cat" | "dog", true]`)
+- Records
+  - Record with string keys (`Record<string, number>`)
+  - Record with number keys (`Record<number, string>`)
+  - Record with union keys (`Record<string | number, unknown>`)
+  - Record with union values (`Record<string, string | number>`)
+  - Record with undefined values (`Record<string, string | undefined>`)
+  - Record with literal keys (`Record<"cat" | "dog", number>`)
+  - Record with literal values (`Record<string, 1 | 2>`)
+  - Record with template keys (``Record<`id_${number}`, string>``)
+  - Union of records (`Record<string, unknown> | Record<number, unknown>`)
+- Objects
+  - Object with named keys (`{ a: number }`)
+  - Object with union values (`{ a: string | number }`)
+  - Object with literal union values (`{ a: "cat" | 1 }`)
+  - Object with optional keys (`{ a?: number }`)
+  - Object with symbol keys (`{ [Symbol("a")]: number }`)
+  - Union of objects (`{ a: 1, b: 2 } | { b: 2, c: 3 }`)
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
+# Contributing guide
+
+We're always looking for new contributors and are happy to help you get started! Some ideas:
+
+- Add a test for a function.
```
eranhirsch, line 5: Maybe rephrase it as "find an interesting test case we haven't tested yet" (I trust you to find better wording) to prevent low-effort copy-paste tests from people looking to add open-source contributions to their profiles. We have never had any issues in the past, but I'm afraid of giving people ideas.

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+- Improve a function's documentation.
+- Work on any issue tagged [good first issue](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22).
+- Work on any issue tagged [help wanted](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22).
+
```
eranhirsch, line 9: Should we add "fix a bug or edge-case"?

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
+# Contributing guide
+
+We're always looking for new contributors and are happy to help you get started! Some ideas:
+
+- Add a test for a function.
+- Improve a function's documentation.
+- Work on any issue tagged [good first issue](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22).
+- Work on any issue tagged [help wanted](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22).
```
eranhirsch, line 8: These could use badges that would also count the number of issues. https://shields.io/badges/git-hub-labels

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Getting started
+
+Requirements:
+
+- Node.js `>=18.20.0`
+- npm `>=10.7.0`
```
eranhirsch, line 15: This isn't accurate, we use nvm and should guide people to rely on nvm to use the version in nvmrc so that we can "lock" versions if we need to (we currently use the latest 22 version I think)

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+```bash
+git clone git@github.com:remeda/remeda.git
+cd remeda
+npm install
+```
+
+Run a test, and watch for changes:
+
+```bash
+npm test -i src/file.test.ts
```
eranhirsch, line 28: :) I always just run `npm t` as it watches everything (after an admittedly long first run of everything).

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+We have a pre-commit script that will format and lint the project.
```
eranhirsch, line 31: We should mention the commands too, in case someone wants to run them before committing.

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Documentation
+
+Help us make the documentation better! See [`docs/README.md`](https://github.com/remeda/remeda/blob/main/docs/README.md).
```
eranhirsch, line 35: Maybe add something about the fact that for most use-cases the documentation is generated from the JSDocs so effectively most documentation PRs wouldn't actually be part of the docs project.

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Guidelines
+
+### Adding a new function
+
+When adding a new function, remember to:
+
+- Add a Typedoc comment with a description, parameters, signature, an example, and tags.
```
eranhirsch, line 43:
```suggestion
- Add a JSDoc comment with a description, parameters, signature, an example, and tags.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+- Add tests for runtime, typing, data-first, and data-last forms.
```
eranhirsch, line 44: Maybe add some more guidelines for testing, like preferring one assert per test, splitting typing and runtime tests (which you did mention). Using toEqualTypeOf and ts-expect-error for type tests?

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+  - For data-last tests, prefer tests using `pipe`, as this better matches real-world usage.
+- Add exports to `src/index.ts`.
+- Add Lodash and Ramda equivalents to `docs/src/content/mapping`.
```
eranhirsch, line 47: Maybe rephrase this to mention that people should actively look for equivalent functions in lodash and ramda when adding something new even if they don't know those libraries.

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+
```
eranhirsch, line 48: We should mention that we depend on type-fest so people should take advantage of types from that library when coming up with new types and not re-invent the wheel.

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+### Input types to consider
```
eranhirsch, line 49: This section will probably be confusing to people as a lot of these things are the result of us maintaining the library, and most people don't think like this about a utility for "arrays". What I would mention is examples, like mentioning that people should prefer `<T extends IterableContainer>(data: T)` over `<T>(data: ReadonlyArray<T>)`, and `<T extends object>(data: T)` over `<T>(data: Readonly<Record<string, T>>)`, and then mention in the testing section that testing types should cover as many use-cases of the supported type as possible.

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+  - Object with optional keys (`{ a?: number }`)
+  - Object with symbol keys (`{ [Symbol("a")]: number }`)
+  - Union of objects (`{ a: 1, b: 2 } | { b: 2, c: 3 }`)
+  - Nested objects (`{ a: { b: { c: 1 } } }`)
+- `readonly` versions of the above
+- `null` and `undefined`
+
+### Design philosophy
+
+**No type annotations.** Functions shouldn't require type annotations to have good types. This gives the best developer experience. The exception is `piped`.
```
eranhirsch, line 107: Let's skip the exception, it's not entirely true because:
```ts
map(piped(filter(isString), only()));
```

Doesn't require type annotations.

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+
+**Have good output types.** One reason Remeda provide "one-liner" functions is if we can have better output types than the defaults. While `data.filter(item => typeof item === 'number')` works, `filter(data, isNumber)` gives a better output type. This is why we have type guards, `hasAtLeast`, `prop`, etc.
```
eranhirsch, line 109: Everything in Remeda is "one-liner" so this sentence wasn't clear to me. We can mention that we provide alternative types for built-ins and trivial implementations for the sake of better typing, but even that doesn't cover the whole thing, we also provide implementations if we want to offer a lazy variant or sometimes just so that it could be used dataLast...

I think it's good to mention that output types are a major concern for the library, just think the phrasing and reasoning is different. We want to move errors from runtime to compile time, so better typing means more compile errors and less runtime errors.

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+
+**Readability counts.** The other reason Remeda provides one-liner functions is if they make things more readable. Compare `map((item) => item + 3)` to `map(add(3))`, where we don't need to come up with a name for `item`. This is why we have `constant`, `doNothing`, `identity`, etc.
+
+**Prefer composition.** We don't provide simple compositions of functions. Having a smaller surface area makes it easier for users to remember Remeda's functions, and easier for us to maintain. We don't have `reject` because it's the same as `filter(isNot)`. An exception is when the composition would be less readable; this is why we have `sum`, even if it's the same as `sumBy(identity)`.
```
eranhirsch, line 113: `sum` isn't the best example because I still think it's redundant (but I agree that right now, the alternative isn't great). Maybe use `compact` as a second example, or `zipObj`

CONTRIBUTING.md
```diff
@@ -0,0 +1,115 @@
...
+
+**Minimal implementations.** Implementations should, as much as possible, be small and self-contained. Importing an individual function should add less than 500 B to the bundle size in most cases.
```
eranhirsch, line 115: Maybe also add that we want to provide the best possible runtime performance where it counts. We consider both CPU and memory, and we consider both proper complexity, but also JS VM optimizations.

eranhirsch (approved):
This is great, I appreciate your point of view on this. I added my thoughts and ideas, but as you mentioned, it's opinionated and my opinion is just one, so it's up to you.

CONTRIBUTING.md
cjquines: my idea behind this was that i wanted to write down our reasoning for why we have certain functions. i guess typing, data-last, and pipes are all reasons, which i might break into a different point (something around working well with the rest of remeda?)

CONTRIBUTING.md
cjquines: not quite sure how to phrase this?

CONTRIBUTING.md
```diff
@@ -0,0 +1,171 @@
+# Contributing guide
+
+We're always looking for new contributors and are happy to help you get started! Some ideas:
+
+- Add a test to a function, testing something that we haven't tested yet.
+- Improve a function's documentation.
+- Fix a bug or an edge case.
+- Work on [![issues labeled good first issue](https://img.shields.io/github/issues/remeda/remeda/good%20first%20issue?style=flat-square)](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%good+first+issue%22).
+- Work on [![issues labeled help wanted](https://img.shields.io/github/issues/remeda/remeda/help%20wanted?style=flat-square)](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%help+wanted%22).
```
eranhirsch, line 9: nit: no final period, it renders badly, and the badges don't look like part of the text anyway. We can move the period before the badge instead.
```suggestion
- Work on [![issues labeled good first issue](https://img.shields.io/github/issues/remeda/remeda/good%20first%20issue?style=flat-square)](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%good+first+issue%22)
- Work on [![issues labeled help wanted](https://img.shields.io/github/issues/remeda/remeda/help%20wanted?style=flat-square)](https://github.com/remeda/remeda/issues?q=is%3Aopen+is%3Aissue+label%3A%help+wanted%22)
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Getting started
+
+We use `npm` as the package manager. We recommend using [`nvm`](https://github.com/nvm-sh/nvm) to manage `Node.js` and `npm` versions.
```
eranhirsch, line 13: maybe add that our *recommended* node version is written in the .nvmrc file, for those installing it manually.

CONTRIBUTING.md
```diff
@@ -0,0 +1,171 @@
...
+
+Install dependencies:
```
eranhirsch, line 15: should these be ### headers?

CONTRIBUTING.md
```diff
@@ -0,0 +1,171 @@
...
+### Adding a new function
+
+When adding a new function, remember to:
+
+- Add a JSDoc comment with a description, parameters, signature, an example, and tags.
+- Add tests for runtime, typing, data-first, and data-last forms.
+  - For data-last tests, prefer tests using `pipe`, as this better matches real-world usage.
+- Add exports to `src/index.ts`.
+- Check for equivalent functions in [Lodash](https://lodash.com/docs/4.17.15) and [Ramda](https://ramdajs.com/docs/), and add them to `docs/src/content/mapping`.
+
```
eranhirsch, line 52: Should we add that we expect all PRs to result in 100% coverage and 0 lint warnings?

CONTRIBUTING.md
```diff
@@ -0,0 +1,171 @@
...
+### Writing types
+
+TODO: using `type-fest`, prefer `<T extends IterableContainer>(data: T)` over `<T>(data: ReadonlyArray<T>)`, and `<T extends object>(data: T)` over `<T>(data: Readonly<Record<string, T>>)`
```
eranhirsch, line 55: I don't understand this TODO, and the sentence that follows, it's unrelated to type-fest...

CONTRIBUTING.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Writing tests
+
+The tests for `file.ts` live in `file.test.ts`. Some guidelines:
```
eranhirsch, line 59:
```suggestion
For each file, the tests for that file are located in a file of the same name with the extension `.test.ts` (e.g. `myFunction.ts` -> `myFunction.test.ts`).

Some guidelines:
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,171 @@
...
+  - Object with literal union values (`{ a: "cat" | 1 }`)
+  - Object with optional keys (`{ a?: number }`)
+  - Object with symbol keys (`{ [Symbol("a")]: number }`)
+  - Union of objects (`{ a: 1, b: 2 } | { b: 2, c: 3 }`)
+  - Nested objects (`{ a: { b: { c: 1 } } }`)
+- `readonly` versions of the above
+- `null` and `undefined`
+
+</details>
+
```
eranhirsch, line 160: Maybe we should add something about not needing to test impossible cases that are prevented by typescript, like testing what a function that doesn't take `undefined` will do if we give it an `undefined` value, I've seen people suggest tests like:

```ts
// @ts-expect-error
map(undefined).toThrowError();
```

These aren't useful tests.

eranhirsch (dismissed):

TkDodo (commented):
great work 👏 

I'm missing:

- conventional commits, which we need for releasing. Granted, we can overwrite it with the merge commit, but still would be good to know as a contributor.
- mention that the jsdocs actually wind up resulting in the API docs in the documentation.
- note that we shouldn't forget to export new files from `index.ts`. We've forgotten that in the past 😅 . Some static analysis for this would be great. As an idea, we could have tests import from the root bundle instead of the src file, so in order to write a test, it must be in the `index.ts` file.

CONTRIBUTING.md
cjquines: i think these sections are too small to get h3 headers

CONTRIBUTING.md
cjquines: i'll add a note on coverage. for lint, we already have `--max-warnings 0` in `npm run lint` so it won't pass the precommit script?

CONTRIBUTING.md
cjquines: i filled in the TODO, does it make more sense now?

cjquines:
> * conventional commits, which we need for releasing. Granted, we can overwrite it with the merge commit, but still would be good to know as a contributor.

this is in the pull request template, so i chose to omit this here, but i guess we can also put it in here?

> * note that we shouldn't forget to export new files from `index.ts`. We've forgotten that in the past 😅 . Some static analysis for this would be great. As an idea, we could have tests import from the root bundle instead of the src file, so in order to write a test, it must be in the `index.ts` file.

this is already mentioned oops, and it is also in the checklist in the pull request template

CONTRIBUTING.md
eranhirsch: I guess it would come up in code review

CONTRIBUTING.md
```diff
@@ -0,0 +1,221 @@
+# Contributing guide
+
+We're always looking for new contributors and are happy to help you get started! Some ideas:
+
+- Add a test to a function, testing something that we haven't tested yet.
```
eranhirsch, line 5: The word test repeats itself, and we don't expect trivial tests to be missing (for most functions) so maybe:

```suggestion
- Add a test to a function, like an interesting edge-case we didn't think about.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,221 @@
...
+// ❌ `T` can be too wide:
+function functionName<T>(data: ReadonlyArray<T>) {}
+
+functionName(data); // inside functionName, data[0] has type string | number
+
+// ✅ `T` is more specific:
+function functionName<T extends IterableContainer>(data: T) {}
+
+functionName(data); // inside functionName, data[0] has type string
+```
```
eranhirsch, line 79: The example is great, and I don't mind if we don't expand on this, but the main reason for using T as the whole array and not just the item type is that we maintain the _shape_ of the array in the output via mapping, so if the array is non-empty, we can return a non-empty output.

eranhirsch (approved):
Looks great!

CONTRIBUTING.md
cjquines: i'll add a note for this. i was trying to think of more guidelines we can put in the writing types section but i haven't thought of any yet. i was thinking of reviewing past PR reviews and looking for patterns

PR #748, opened 2024-06-23
docs(search): Add Algolia DocSearch

eranhirsch:
Use DocSearch globally to search in our docs.

This PR also includes changes to the site to make the crawler more efficient and useful at detecting our different content types.


eranhirsch:
This PR affects how the crawler is able to map our site to records. I need to ship it in order to run the crawler on the new version of the site because I can't run it locally.

Issue #675, opened 2024-05-08
conditional: unexpected types when using isNullish/isNonNullish


JeongJuhyeon:
TS Version: 5.4.5

Example:
```ts
pipe(
      [{ x: 10 }, { x: 20 }], 
      firstBy(prop("x")), 
      conditional(
        [isNullish, (o) => o], // got "o: never", expected "o: undefined"
        [isNonNullish, (o) => o] // got: "o: {}", expected: "o: { x: number; }"
      )
    );
```

I checked separately if it had to do with the FirstBy type but that doesn't seem to be the case:

```ts    
const k1 = pipe([{ x: 10 }, { x: 20 }], firstBy(prop("x"))); // FirstBy<{x: number; }[]>, as expected
type t1 = Extract<typeof k1, null | undefined>; // undefined, as expected
type t2 = NonNullable<typeof k1>; // {x: number; }, as expected
```

It looks like the conditional's `When`s _must_ be a type guard of the form `is A` where `A` is a primitive type, so not even "NonNullable<something>". Is this correct and intended? If so, it would be good to reflect this in the documentation.

eranhirsch:
We've got other reports about conditional acting different than expected for people. It's most likely a bug in our typing for conditional.

grigorischristainas:
@eranhirsch I tried to modify the guards as a workaround and I think `isNullish` could be modified as follows:
```ts
import type { NarrowedTo } from "./internal/types";

export function isNullish<T>(
  data: T | null | undefined,
): data is NarrowedTo<T, null | undefined> {
  return data === null || data === undefined;
}
```
This narrowing allows for `o` in the example above being inferred as `null | undefined`, which I think is correct. 

However, I am not sure what can be done about `isNonNullish`. Best I could do by modifying the guard is end up with `o` being inferred as `unknown` in `conditional`.

eranhirsch:
@grigorischristainas, good catch, `NarrowedTo` does improve the typing for isNullish, but still doesn't fix the problem with conditional

PR #750, opened 2024-06-25
fix(conditional): improve type narrowing

lesha1201:
Fixes https://github.com/remeda/remeda/issues/675

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/conditional.test.ts
+++ b/src/conditional.test.ts
@@ -1,8 +1,12 @@
 import { conditional } from "./conditional";
+import { firstBy } from "./firstBy";
 import { isDeepEqual } from "./isDeepEqual";
+import { isNonNullish } from "./isNonNullish";
+import { isNullish } from "./isNullish";
 import { isNumber } from "./isNumber";
 import { isString } from "./isString";
 import { pipe } from "./pipe";
+import { prop } from "./prop";
 
 describe("runtime (dataFirst)", () => {
   it("falls back to trivial default", () => {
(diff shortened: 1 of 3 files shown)


src/conditional.test.ts
```diff
@@ -68,42 +72,177 @@ describe("runtime (dataLast)", () => {
...
 describe("typing", () => {
-  it("can narrow types in the transformers", () => {
-    const data = 3 as number | string;
-    conditional(
-      data,
-      [
-        isString,
-        (str) => {
-          expectTypeOf(str).toEqualTypeOf<string>();
-          expectTypeOf(str).not.toEqualTypeOf<number>();
```
lesha1201, line 79: I removed `expectTypeOf(str).not.toEqualTypeOf` since there is no need to check both statements because `toEqualTypeOf` checks for the exact type.

cjquines: **praise:** agreed!

src/conditional.ts
```diff
@@ -1,11 +1,13 @@
...
+import { type GuardType } from "./internal/types";
 
-type Case<In, Out, Thru extends In = In> = readonly [
-  when: ((data: In) => boolean) | ((data: In) => data is Thru),
-  then: (data: Thru) => Out,
-];
+type Case<
+  In,
+  Out,
+  CaseFn extends (x: In) => boolean = (x: In) => boolean,
```
cjquines, line 9: **thought:** i'm not sure how i feel about the name `CaseFn`

maybe `When` is better? :shrug:

src/internal/types.ts
```diff
@@ -218,3 +218,19 @@ type InternalIsUnion<T, U = T> = (
...
+ *
+ * @example
+ * type TypeGuardFn = (x: unknown) => x is string;
+ * type Result = GuardType<TypeGuardFn>; // `string`
+ */
+export type GuardType<T, Fallback = never> = T extends (
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- we don't care about arguments types here
+  x: any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- we don't care about arguments types here
+  ...rest: any
```
cjquines, line 233: **question:** is there a reason to use `any` here rather than `unknown`?

cjquines (approved):
a few thoughts, but this looks great to me, thanks for the contribution!

src/internal/types.ts
cjquines: strange! i think this is fine, but i was just curious...

src/conditional.ts
lesha1201: @cjquines Done

src/internal/types.ts
lesha1201: @cjquines 

https://stackoverflow.com/questions/62136028/why-dont-parametersfunc-extend-unknown-arrayunknown-in-ts-strict-mode

image

lesha1201: It seems we can use `never` but it's probably the same as `any` in this case.

cjquines:
thanks again!

Issue #661, opened 2024-05-01
Add "R" or "Remeda" globally to the docs website


illiaChaban:
When you go to lodash docs, you can open dev tools > console, type "_" and access all lodash methods. This helps with trying things out and expeirmenting. 
It would be nice if Remeda docs had the same feature

eranhirsch:
Sounds like a cool idea

eranhirsch:
I'm not sure this is a good first issue, I don't have a good idea how this is achievable

cjquines:
let me outline what i think the steps are

* in https://github.com/remeda/remeda/blob/master/docs/src/layouts/layout.astro , add the following

```html
<script type="module">
  import * as R from "remeda";
  window.R = R;
</script>
```

* check to see if it works (it should [according to astro docs](https://docs.astro.build/en/guides/client-side-scripts/#client-side-scripts))
* (potentially) change the package.json in docs to include remeda from the top-level directory so the version's always updated

i'll be happy to help anyone who runs into snags

eranhirsch:
It looks cool. I don't want the docs project to import outside of NPM because it creates a lot of weird stuff when using Remeda as a utility library in the docs site. We can work around this by bumping the dependency in the docs package.json every time we publish a version of Remeda using another step in the Github workflow that releases it.

PR #761, opened 2024-07-03
docs(layout): add global R for docs

dartess:
Closed https://github.com/remeda/remeda/issues/661

---

Make sure that you:

- [ ] New methods added to `docs/src/content/mapping` for both Lodash and Ramda.

---

### Available Types:

existing).

</details>

### DIFF
--- a/docs/public/sandbox.js
+++ b/docs/public/sandbox.js
@@ -1,0 +1,8 @@
+import * as R from "https://esm.sh/remeda";
+
+window.R = R;
+
+console.log(
+  "%cTry out Remeda right here in your browser console; use the global R to access all the functions.",
+  "font-family: sans-serif; font-size: 16px",
+);
--- a/docs/public/sandbox.js
+++ b/docs/public/sandbox.js
@@ -1,14 +1,18 @@
 import * as R from "https://esm.sh/remeda";
 
-const NOTICE = "font-family: sans-serif; font-size: 16px";
-const CODE = "font-family: monospace; font-size: 16px; font-weight: 500";
+const NORMAL_STYLE = "font-family: sans-serif; font-size: 1rem";
+const BOLD_STYLE =
+  "font-family: sans-serif; font-size: 1.25rem; font-weight: 600";
+// The console's default style is mono-spaced font...
+const CODE_STYLE = "";
 
 window.R = R;
 
 console.log(
-  "%cYou can try out Remeda right here with the %cR%c global.",
-  NOTICE,
-  CODE,
-  NOTICE,
+  "%cYou can try out %cRemeda%c right here via the global const %cR",
+  NORMAL_STYLE,
+  BOLD_STYLE,
+  NORMAL_STYLE,
+  BOLD_STYLE,
 );
-console.log("%ce.g. %cR.add(5, 10)%c", NOTICE, CODE, NOTICE);
+console.log("%ce.g. %cR.add(5, 10)", NORMAL_STYLE, CODE_STYLE);
--- a/docs/public/sandbox.js
+++ b/docs/public/sandbox.js
@@ -1,14 +1,14 @@
 import * as R from "https://esm.sh/remeda";
 
-const SANS_SERIF = "font-family: sans-serif; font-size: 16px";
-const MONOSPACE = "font-family: monospace; font-size: 16px";
+const NOTICE = "font-family: sans-serif; font-size: 16px";
+const CODE = "font-family: monospace; font-size: 16px; font-weight: 500";
 
 window.R = R;
 
 console.log(
   "%cYou can try out Remeda right here with the %cR%c global.",
-  SANS_SERIF,
-  MONOSPACE,
-  SANS_SERIF,
+  NOTICE,
+  CODE,
+  NOTICE,
 );
-console.log("%ce.g. %cR.add(5, 10)%c", SANS_SERIF, MONOSPACE, SANS_SERIF);
+console.log("%ce.g. %cR.add(5, 10)%c", NOTICE, CODE, NOTICE);
--- a/docs/public/sandbox.js
+++ b/docs/public/sandbox.js
@@ -2,7 +2,18 @@ import * as R from "https://esm.sh/remeda";
 
 window.R = R;
 
+const sansSerif = "font-family: sans-serif; font-size: 16px";
+const monospace = "font-family: monospace; font-size: 16px";
+
 console.log(
-  "%cTry out Remeda right here in your browser console; use the global R to access all the functions.",
-  "font-family: sans-serif; font-size: 16px",
+  "%cYou can use Remeda with the %cR%c global.",
+  sansSerif,
+  monospace,
+  sansSerif,
+);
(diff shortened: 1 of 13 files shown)


docs/src/layouts/base.astro
```diff
@@ -26,6 +26,13 @@ const { title } = Astro.props;
     />
     <title>{`${title === undefined ? "" : `${title} - `}Remeda`}</title>
     <script is:inline src="/dist/scripts/init-theme.js"></script>
+    <script type="module">
+      import * as R from "https://esm.sh/remeda";
+      window.R = R;
+      console.log(
+        "Use variable R right here to access the latest version of the library",
+      );
+    </script>
```
eranhirsch, line 35: Let's make this a little cleaner:

1. Put the script under the public folder (but not under dist/scripts which is for autogenerated scripts).
2. Then, put `<script is:inline fetchpriority="low" src="<location>" type="module"></script>` in the **body** section of the layout, so that the script is only processed *after* the full page loads (as it can wait).

docs/src/layouts/base.astro
```diff
@@ -26,6 +26,13 @@ const { title } = Astro.props;
     />
     <title>{`${title === undefined ? "" : `${title} - `}Remeda`}</title>
     <script is:inline src="/dist/scripts/init-theme.js"></script>
+    <script type="module">
+      import * as R from "https://esm.sh/remeda";
+      window.R = R;
+      console.log(
+        "Use variable R right here to access the latest version of the library",
```
eranhirsch, line 33: I'm not sure about the wording here. I think it needs more context to make it obvious that this is an intended message and is here to guide the user. Without it, it looks like an error message. e.g., "Try out Remeda right here in your browser console; use the global `R` to access all the functions."

We can also utilize the fact that console.log can be styled to make the content bigger and more pronounced, at least using a header.

```js
console.log("%c<CONTENT>", "font-size: 24px; color: blue;")
```

@cjquines, wdyt?

eranhirsch (changes requested):

docs/src/layouts/base.astro
dartess: I'd appreciate better wording.

Actually the main purpose of this post for me is to highlight that the _latest_ version is being used, since R will also be available on the documentation page of the first version, but the examples will not work as expected.

dartess: > console.log can be styled

Nice thought. I'd add (maybe) _italics_ and remove the color setting since the console background may vary.

eranhirsch: I'm OK with anything you find that looks nice and clear. I couldn't find examples of sites that do something similar, so we can't just copy. Please attach a screenshot once you update the PR so it's easy to review.

eranhirsch: > Actually the main purpose of this note for me is to highlight that the latest version is being used, since R will also be available on the documentation page of the first version, but the examples will not work as expected.

This is not a real concern of ours. We don't need to treat this feature as a sandbox for the current page, it should be seen as a site-wide feature. The v1 migration page is not an API, it's just a migration guide.

dartess: image
it's just that the non-monospace and increased font size seem to look neat.

docs/src/layouts/base.astro
dartess: I think I did as you suggested. It works but I'm not familiar with astro, please make sure this is what you meant.

docs/src/layouts/base.astro
cjquines: what about something like

```ts
console.log(
  "%cYou can use Remeda with the %cR%c global.",
  "font-family: sans-serif; font-size: 16px",
  "font-family: monospace; font-size: 16px",
  "font-family: sans-serif; font-size: 16px",
);
console.log(
  "%cTry typing %cR.add(5, 10)%c here:",
  "font-family: sans-serif; font-size: 16px",
  "font-family: monospace; font-size: 16px",
  "font-family: sans-serif; font-size: 16px",
);
```

cjquines: ("here in your browser console" is, i think, redundant... if you're seeing this, you're already looking at the browser console?)

dartess: I think this is enough to close the goal :)

cjquines (dismissed):

docs/public/sandbox.js
```diff
@@ -0,0 +1,19 @@
+import * as R from "https://esm.sh/remeda";
+
+window.R = R;
+
+const sansSerif = "font-family: sans-serif; font-size: 16px";
+const monospace = "font-family: monospace; font-size: 16px";
```
eranhirsch, line 6: global consts should be UPPER_CASE
```suggestion
const SANS_SERIF = "font-family: sans-serif; font-size: 16px";
const MONOSPACE = "font-family: monospace; font-size: 16px";
```

eranhirsch, line 6: Place consts *above* logic
```suggestion
const sansSerif = "font-family: sans-serif; font-size: 16px";
const monospace = "font-family: monospace; font-size: 16px";

window.R = R;
```

docs/public/sandbox.js
```diff
@@ -0,0 +1,19 @@
...
+
+console.log(
+  "%cYou can use Remeda with the %cR%c global.",
```
eranhirsch, line 9: instead of use, let's use "try out", the goal here is to provide a way to play around with Remeda, not a way to use it.
```suggestion
  "%cYou can try out Remeda right here with the %cR%c global.",
```

docs/public/sandbox.js
```diff
@@ -0,0 +1,19 @@
...
+  sansSerif,
+  monospace,
+  sansSerif,
+);
+console.log(
+  "%cTry typing %cR.add(5, 10)%c here:",
```
eranhirsch, line 15:
```suggestion
  "%ce.g. %cR.add(5, 10)%c",
```

docs/public/sandbox.js
```diff
@@ -0,0 +1,19 @@
...
+  sansSerif,
+  monospace,
+  sansSerif,
```
eranhirsch, line 18: Perhaps italics for the example?

eranhirsch (approved):
LGTM. I have some code nits and some content suggestions

docs/public/sandbox.js
dartess: @eranhirsch 
image
I thought so too, but monospace Italic looks a little weird, imho

docs/public/sandbox.js
```diff
@@ -0,0 +1,14 @@
+import * as R from "https://esm.sh/remeda";
+
+const SANS_SERIF = "font-family: sans-serif; font-size: 16px";
```
eranhirsch, line 3: Let's add 
```suggestion
const SANS_SERIF = "font-family: sans-serif; font-size: 16px; font-weight: 500";
```

eranhirsch (approved):

eranhirsch (approved):

PR #769, opened 2024-07-04
docs: shorten menu item label for v1 migration guide

eranhirsch:
The current label repeats the "migration" part which is already part of the menu title, and repeats the v2 part which isn't relevant for the viewer.


eranhirsch:
image

vs

image

cjquines (approved):

PR #772, opened 2024-07-07
feat(times): Refined typing and improved perf

eranhirsch:
The current typing implementation doesn't use the provided size param to build the output, this means that even for cases like times(5, ...) our returned array type is not non-empty.

I also noticed that we are throwing on "invalid" inputs in a way which doesn't conform with how we handle these kind of cases in other functions. Instead, we should almost always fallback to a runtime behaviour that makes sense, especially when a semanticly correct option exists, and even more so when we can make the types work for it too.

Looking into this fix, I also wanted to try sparse arrays again, we have a lint rule which might not make sense for us driving us away from sparse arrays and it might not always be the right call. For this function we can get performance benefits of about 2x-3x on large arrays using sparse arrays.
I added a bunch of benchmarks to see what implementation is the best. I will remove them before the merge because they won't make sense once this PR is approved.

### DIFF
--- a/src/times.ts
+++ b/src/times.ts
@@ -26,8 +26,15 @@ type TimesArray<
  * `fn` is passed one argument: The current value of `n`, which begins at `0`
  * and is gradually incremented to `n - 1`.
  *
- * @param count - A value between `0` and `n - 1`. Increments after each function call.
- * @param fn - The function to invoke. Passed one argument, the current value of `n`.
+ * NOTE: The function returns a strict tuple with exactly N elements in it, but
+ * for very large literals TypeScript might not be able to compute it, for those
+ * cases cast the number to a primitive `number` (e.g. `1000 as number`) to
+ * bypass the refined typing.
+ *
+ * @param count - A value between `0` and `n - 1`. Increments after each
+ * function call.
+ * @param fn - The function to invoke. Passed one argument, the current value of
+ * `n`.
  * @returns An array containing the return values of all calls to `fn`.
  * @signature
  *    R.times(count, fn)
@@ -48,7 +55,13 @@ export function times<T, N extends number>(
  * `fn` is passed one argument: The current value of `n`, which begins at `0`
  * and is gradually incremented to `n - 1`.
  *
- * @param fn - The function to invoke. Passed one argument, the current value of `n`.
+ * NOTE: The function returns a strict tuple with exactly N elements in it, but
+ * for very large literals TypeScript might not be able to compute it, for those
+ * cases cast the number to a primitive `number` (e.g. `1000 as number`) to
+ * bypass the refined typing.
+ *
+ * @param fn - The function to invoke. Passed one argument, the current value of
+ * `n`.
  * @returns An array containing the return values of all calls to `fn`.
  * @signature
  *    R.times(fn)(count)
--- a/src/times.ts
+++ b/src/times.ts
@@ -42,13 +42,22 @@ export function times(...args: ReadonlyArray<unknown>): unknown {
 }
 
 function timesImplementation<T>(count: number, fn: (n: number) => T): Array<T> {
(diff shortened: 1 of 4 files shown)


eranhirsch:
Benchmark results comparing the existing implementation with the one introduced here:

```bash
  final impl - src/times.bench.ts > length 10
    1.44x faster than current implementation

  final impl - src/times.bench.ts > length 100
    1.89x faster than current implementation

  final impl - src/times.bench.ts > length 1000
    2.59x faster than current implementation

  final impl - src/times.bench.ts > length 10000
    2.65x faster than current implementation

  final impl - src/times.bench.ts > length -10
    52.03x faster than current implementation

  final impl - src/times.bench.ts > length 10.5
    1.49x faster than current implementation
```

cjquines:
will also look at this closer later—but how much do we care about performance in general? a 2x-to-3x constant speedup does not look super significant to me (compared to, say, a non-constant speedup)

eranhirsch:
> how much do we care about performance in general? a 2x-to-3x constant speedup does not look super significant to me (compared to, say, a non-constant speedup)

We don't care about performance, but if we can offer a strictly better alternative (that is, it's more efficient for most practical usages) at no down cost, we should offer it.

I used this PR to _test out_ sparse arrays again, as we have played around with them in the past and used them incorrectly and opted for alternative implementations instead. I started out simply removing the check that threw an error and went from there.

src/times.ts
```diff
@@ -1,14 +1,40 @@
...
  * `fn` is passed one argument: The current value of `n`, which begins at `0`
  * and is gradually incremented to `n - 1`.
  *
- * @param count - A value between `0` and `n - 1`. Increments after each function call.
- * @param fn - The function to invoke. Passed one argument, the current value of `n`.
+ * NOTE: The function returns a strict tuple with exactly N elements in it, but
+ * for very large literals TypeScript might not be able to compute it, for those
+ * cases cast the number to a primitive `number` (e.g. `1000 as number`) to
+ * bypass the refined typing.
+ *
```
cjquines, line 33: **suggestion:** we can prevent this on the type-level with `type-fest`'s [`GreaterThan`](https://github.com/sindresorhus/type-fest/blob/main/source/greater-than.d.ts). then we can remove this note?

cjquines (dismissed):

cjquines:
> if we can offer a strictly better alternative (that is, it's more efficient for most practical usages) at no down cost, we should offer it.

what's a down cost, then? i'd argue that there's a (minimal) cost here because we have to put a comment to explain why we're suppressing a lint rule. it's a small cost, and in this case i'm more than happy to take it. but there's a line we gotta take eventually right

consider the diff function from [microdiff](https://github.com/AsyncBanana/microdiff/tree/master). it's got some small optimizations to make it faster, see [the explanation in this blog post](https://byteofdev.com/posts/microdiff/#performance-focused-architecture). for example, there's this part that checks two values

```ts
isNaN(objKey) ? objKey + "" === newObjKey + "" : +objKey === +newObjKey
```

which could be confusing, but only a little bit? idk

eranhirsch:
There's no single right answer here; it's subjective. I think that if the change:
- It doesn't require changes to the types of inputs or outputs.
- has a ~+15% runtime improvement over the current implementation for _all_ use cases.
- Its code is as readable as the existing one.
- It doesn't rely on JSisms or known quirks of v8 for its gains (i.e., it makes sense from a computer science perspective).
- The PR comes with a benchmark that can show the gains reproducibly.

I'd be fine with accepting it. I think Remeda is used as a black box; most users would want to get the best available algorithms and not have to reimplement functions we already export just because their specific use-case requires the extra squeeze of resources the better algorithm would provide them.

wdyt?

cjquines (approved):

PR #773, opened 2024-07-07
feat(isStrictEqual, isShallowEqual): add more equality tools

eranhirsch:
A non-deep version of an equality guard for cases where you don't need the deepness but still would appreciate the typing and dataLast invocation.

---

Make sure that you:

---

### Available Types:

existing).

</details>


cjquines:
will read closer later, but my first thought is that the PR's current `isEqual` is what i think of when i read `isShallowEqual`, and i don't see a use for the PR's current `isShallowEqual`?

cjquines:
in particular, the type-level information i am getting from `isShallowEqual(a, b)` when `!isEqual(a, b)` is that `a` and `b` have the same shape, which is not what i think of when i read `isShallowEqual`. i'm not sure why we'd have an equals function that only goes "one-level deep". it's weird to me that `isShallowEqual(data, { a: 1 })` is sometimes true but `isShallowEqual(data, { a: [1] })` is never true

eranhirsch:
This was a suggestion from @TkDodo; the implementation of `isShallowEqual` comes from Zustand's `shallow` function. It's useful when you want to replicate behavior similar to React's dependency arrays.

src/isDeepEqual.ts
```diff
@@ -1,11 +1,11 @@
 import { purry } from "./purry";
 
 /**
- * Performs a deep *semantic* comparison between two values to determine if they
- * are equivalent. For primitive values this is equivalent to `===`, for arrays
- * the check would be performed on every item recursively, in order, and for
- * objects all props will be compared recursively. The built-in Date and RegExp
- * are special-cased and will be compared by their values.
+ * Performs a _deep_ *semantic* comparison between two values to determine if
```
cjquines, line 4: **nit:** both `_` and `*` are italics in markdown, so maybe this should be `*deep semantic*`?

i think we can also drop the word "semantic" here. i don't think it's helpful

src/isDeepEqual.ts
```diff
@@ -1,11 +1,11 @@
...
+ * they are equivalent. For primitive values this is equivalent to `===`, for
+ * arrays the check would be performed on every item recursively, in order, and
+ * for objects all props will be compared recursively. The built-in Date and
+ * RegExp are special-cased and will be compared by their values.
```
cjquines, line 8: **question:** should we mention `Set` and `Map` here?

i now realize that we added that in some PR but didn't update this description

src/isEqual.ts
```diff
@@ -0,0 +1,76 @@
+import { purry } from "./purry";
+
+/**
+ * Performs a _native_ equality check between two values to. For primitive
```
cjquines, line 4: **nit:** this first sentence doesn't make sense, did you mean to write "to determine if they are equivalent"?

maybe we can even write "to determine if they are identical"?

src/isDeepEqual.ts
cjquines: if we want to keep a word like semantic, mahybe "structural" would be it, but idk, i think "deep" alone conveys what's needed?

src/isShallowEqual.ts
```diff
@@ -0,0 +1,163 @@
...
+ * Performs a _shallow_ *semantic* comparison between two values to determine if
+ * they are equivalent. For primitive values this is equivalent to `===`, for
+ * arrays, an identity check would be performed on every item, in order, and for
+ * objects props will be matched and checked for identity.
+ *
+ * !IMPORTANT: symbol properties of objects are not supported right now and
+ * might result in unexpected behavior. Please open an issue in the Remeda
+ * github project if you need support for these types.
+ *
+ * !IMPORTANT: All built-in objects (Promise, Date, RegExp) are shallowly equal,
```
cjquines, line 13: **nit:** isn't this technically incorrect, as `Set` and `Map` have comparators?

src/isDeepEqual.ts
eranhirsch: I agree, it was confusing to me too

src/isDeepEqual.ts
eranhirsch: I think it's assumed that they work. We had them in the important block below as stuff that doesn't work because it would be implied that they do if we didn't.

src/isDeepEqual.ts
```diff
@@ -32,11 +38,11 @@ export function isDeepEqual<T, S extends T>(
 export function isDeepEqual<T, S extends T = T>(data: T, other: S): boolean;
 
 /**
- * Performs a deep *semantic* comparison between two values to determine if they
- * are equivalent. For primitive values this is equivalent to `===`, for arrays
- * the check would be performed on every item recursively, in order, and for
- * objects all props will be compared recursively. The built-in Date and RegExp
- * are special-cased and will be compared by their values.
+ * Performs a _deep_ *semantic* comparison between two values to determine if
```
cjquines, line 41: **nit:** match the other description

cjquines (approved):

src/isStrictEqual.ts
```diff
@@ -0,0 +1,82 @@
...
+ * See:
+ * - `isDeepEqual` for a semantic comparison that allows comparing arrays and
+ * objects "by-value", and recurses for every item.
+ * - `isShallowEqual` if you need to compare arrays and objects "by-value" but
+ * don't want to recurse into their values.
+ *
+ * @param data - The first value to compare.
+ * @param other - The second value to compare.
+ * @signature
+ *    R.isEqual(data, other)
```
felipecrs, line 25: Wasn't this meant to be R.isStrictEqual?

eranhirsch: It should be, the rename was last moment so I missed this... i'll send a fix.

PR #777, opened 2024-07-11
feat(when): add the when utility

eranhirsch:
A simplified version of the `conditional` function that provides better typing, and with better ergonomics: it doesn't require an array wrapper, it has an implicit default case that makes sense for a lot of use-cases, and it passes arguments through.

---

Make sure that you:

---

### Available Types:

existing).

</details>


src/branch.ts
```diff
@@ -0,0 +1,183 @@
...
+  OnFalse extends (
+    data: Exclude<T, GuardType<Predicate>>,
+    ...args: Args
+  ) => unknown,
+>(
+  data: T,
+  predicate: Predicate,
+  onTrue: OnTrue,
+  onFalse: OnFalse,
+  ...args: Args
```
cjquines, line 117: **thought:** the `...args` reading here is confusing to me, especially since you can only do it if you have the `onFalse` argument. i would be much happier if both `onTrue` and `onFalse` only took in `data`, or we only had the `dataLast` version and not the `dataFirst` version

src/branch.ts
```diff
@@ -0,0 +1,183 @@
...
+ * @param onTrue - The function that would run when the predicate returns
+ * `true`.
+ * @signature
+ *   branch(data, predicate, onTrue)
+ *   branch(data, predicate, onTrue, onFalse, ...extraArgs)
+ * @example
+ *   branch(data, isNullish, constant(42));
+ *   branch(data, (x) => x > 3, add(1), multiply(2));
+ *   branch(data, isString, (x, radix) => parseInt(x, radix), identity(), 10);
+ * @dataLast
```
cjquines, line 91: **nit:** dataFirst

src/branch.ts
cjquines: now that i think about it, the optional `onFalse` here is also weird. in particular it feels like this function is trying to do two different things, which should be two different functions. i think the ramda design of having a separate `defaultTo` and `ifElse` feels right to me

maybe we should split this into

```ts
// defaultValue, not a function
defaultIf(predicate, defaultValue)(data)

ifElse(predicate, onTrue, onFalse)(data)
```

eranhirsch:
Hey, thanks for your review and questions, they make sense, and I'll try to convey my thought process around this:

I wrote this utility in one of my projects where I needed to fix something about an object I was constructing, but only in certain cases. I wanted a way to run code conditionally, but only in some cases, and the ergo of conditional (and the typing issues) weren't a good fit.

```ts
const result = pipe(
  ...
  branch(
    // This is a type-predicate
    (x) => x.mode === "double",
    (x) => ({ ...x, data: [...x.dataPartA, ...x.dataPartB] }),
  ),
  ...
);
```

This isn't easily solvable with our typing of `conditional`, so I had to build a new utility.

2. While building this, I looked for an equivalent in Lodash and Ramda and came across ifElse. It made sense to add an else clause because the implementation was already using the ternary operator, which supports it. This way, we provide a de facto wrapper around the ternary for use in pipes, which is cool.

3. But I didn't want to always call it with an else clause because sometimes you just don't need one, and having a bunch of `identity()` calls just makes the code less readable and confusing. So, I opted for an optional argument instead.

4. The variadic args feature came by while writing the tests; I wanted to show a cool usage using the map to fill in gaps in an array, but I wanted to use the `index` param to create unique items. [Ramda's ifElse supports this too](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wEpwGZQiOAIimRABNlCBuAKBoGMIA7VebCCOAXkwDphsAUQA2qAKYAKGnDgB6WXAB6AfmlwJADwBccVlGBMA5gBoUOgEYdhY5E1PmLVm3bj1HEa7YCU3AHxqZDW4uHkIACzFhYQhCOAAyOJR4xPNk12M1CR8uXzgAZgyZLL84ABYMr1ogA).

5. But it doesn't make sense in a data-first call as there's only one invocation, so the arguments come from the same place and could be used via closure instead of pass-through. I wanted the function to have the same signature for both data-first and data-last, so I added it to the data-first. Still, the ambiguity with the optional param means we either need to make more heuristic choices in the currying function or *drop support*. I'm fine with doing any of the 3 options in this regard: 1. dropping it from both data-first signatures, adding it to both, and having some more heuristics in the currying function, or 3. keeping the current impl.

6. `defaultTo` is a completely different thing; it only takes a **value** and not a *mapping function*. With the availability of `constant` and `branch`, this function isn't required as it's easy to compose. In other words, it's a subset of functionality that is only good for a specific case and could be composed using other more general functions, so following similar decisions in the past (e.g. `compact`), it shouldn't be included in the lib.

eranhirsch:
Another way to solve the issues you raised that we can consider is creating a hybrid type for the mapping function that would either be a single function (for the no-else path) or an **object** with two mapper functions (for the with-else path):

```ts
branch(data, predicate, onTrue, ...args);
branch(data, predicate, { onTrue, onFalse }, ...args);
branch(predicate, onTrue): (data, ...args) => 
branch(predicate, { onTrue, onFalse })): (data, ...args) =>
```

The typing is a little harder, but not by much, and we can use `purry` and skip the heuristic problems.

cjquines:
hey, some responses:

agreed with points 1 and 2

> 3. But I didn't want to always call it with an else clause because sometimes you don't need one, and having a bunch of `identity()` calls makes the code less readable and confusing. So, I opted for an optional argument instead. We can drop the optionality of `onElse`, but I think there are a lot of use cases where it isn't needed (defaultTo, my example above, etc...) that become more cumbersome when it's required.

hm, my thought is that in this case it should be named a different thing. having one `branch` function for both is elegant, but i find it hard to read. in my head i see `branch`, `onTrue`, what happens when it's false? oh, it's not there, so it passes through. i think there could be a better name in this case, mostly _because_ there are a lot of use cases where it isn't needed, and in this case seeing `branch` doesn't quite make sense to me?

> 6. But it doesn't make sense in a data-first call as there's only one invocation, so the arguments come from the same place and could be used via closure instead of pass-through. I wanted the function to have the same signature for both data-first and data-last, so I added it to the data-first. Still, the ambiguity with the optional param means we either need to make more heuristic choices in the currying function or _drop support_. I'm okay with doing any of the three options: 1. dropping it from both data-first signatures, adding it to both, and having some more heuristics in the currying function, or 3. keeping the current impl.

it's true that if we don't support this now and add this later, then it'll be a breaking change. i'm still not sure how i feel about the ergonomics of this, but this is a smaller thought than the one i have under 3

> 7. `defaultTo` is a completely different thing; it only takes a **value** and not a _mapping function_. With the availability of `constant` and `branch`, this function isn't required as it's easy to compose. In other words, it's a subset of functionality that is only good for a specific case and could be composed using other more general functions, so following similar decisions in the past (e.g., `compact`), it shouldn't be included in the lib.

so we didn't include `compact` because it's composable, but we've included other functions that are easily written for the sake of readability, like `add`. in this case i would also make an argument for readability, which is the same argument i have under 3

eranhirsch:
I've sent a new version with my suggestion to use an object to label the if and else paths when both are used. This addresses your concern about ambiguity and makes the currying simpler and more correct by not excluding functions as data.

A simple `if` statement and an `if-else` construct are both considered branches from a CS perspective, so I would consider both of them equally valid semantically for a function called branch; I don't see the value in splitting this into two functions with almost identical code.

I feel that the example with `add` is not symmetric to the example with `compact`. Compact would have the same implementation as a filter; so it forces us to handle laziness and testing although it is nearly identical `filter` already. `add`, on the other hand, doesn't have an equivalent function. It's not complex to write a function that does it, but that's not the point. the better example would be if someone wanted to add an `increment` function when we already have `add`; `add` is more valuable than `increment` because it is more flexible, there are more use-cases where it's valuable, as-is, and `increment` on the other hand, could be done using `add(1)`.

cjquines:
i like the idea about the `onTrue` and `onFalse` case, i think it helps. i still have some the same thoughts about the case without `onFalse`. sorry if i'm being too tough about this, but let me try to explain? when i read

```ts
map(
  data,
  branch(
    (x) => x === undefined,
    (_, index) => `item_${index}`,
  ),
);
```

my instinct is that `branch` takes two arguments, that the data is being branched into the `(x) => x === undefined` and the `` (_, index) => `item_${index}` `` case. but then i have to realize that, no, this is actually an `if` statement, the first one is a condition, the second one is the output, and there's a third, implicit output, if the condition is false.

compare something like

```ts
map(
  data,
  changeIf(
    (x) => x === undefined,
    (_, index) => `item_${index}`,
  ),
);
```

maybe it'd help if we got feedback from outside the two of us?

also, re

> splitting this into two functions with almost identical code

i dont actually think the code is almost identical, the purrying would be much simpler for example

eranhirsch:
Don't worry about making extra rounds of reviewing. I don't mind it and think it results in a better-quality product.

I understand your concern about the naming and readability, and the name `branch` was just something I came up with because I didn't like `ifElse` for obvious reasons 😉 and couldn't use `if`. Looking at Lodash and Ramda also makes it clear that `branch` isn't a name they thought suitable either. One alterantive that I have in mind is calling it `when` instead, so that:

```ts
map(data, when(isString, (x) => parseInt(x)));
map(data, isNullish, (_, index) => `item_${index}`);

const result = pipe(
  ...
  when(isSomthing, {
    onTrue: doSomthing,
    onFalse: doSomethingElse,
  }),
  ...
);

myApi({ onFoo: when(isError, handleError, handleSuccess) });
```

It looks like Ramda uses that name similarly (and I missed it when adding the mapping docs): https://ramdajs.com/docs/#when

Otherwise, it would be helpful for the discussion so that we can open this to a broader audience to help decide if you can come up with your own suggestions for what changes you see would suffice to approve this PR. It would make it easier to see both options side-by-side when making the call.

eranhirsch:
@cjquines - Do you have any objections to the current name? I need this utility in a project of mine but am reluctant to ship it without a review.

cjquines:
i started a discussion in https://github.com/remeda/remeda/discussions/824

eranhirsch:
@cjquines, let's get this PR shipped

src/when.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    it("passes extra args to the functions", () => {
+      expect(
+        when(
+          "123",
+          isString,
+          (x, radix) => Number.parseInt(x, radix),
+          // Extra args:
+          10 /* radix */,
+        ),
+      ).toBe(123);
```
cjquines, line 34: **suggestion:** this wouldn't catch if the `10` wasn't being passed

```suggestion
          4 /* radix */,
        ),
      ).toBe(27);
```

cjquines (dismissed):
i think my remaining gripe is how we pass the extra arguments to `onTrue`, `onFalse`, but i don't have any better ideas and i think it's fine. it's strange but the vast majority of cases it's not needed

src/when.test.ts
eranhirsch: Good catch

PR #785, opened 2024-07-20
feat(chunk): literal chunk sizes

### LINKED ISSUE #784: `chunk` can cause nodejs to throw memory exception
In `chunk<T[]>(array: T[], size: number): Chunked<T[]>`, if `size = 0`, then nodejs will crash with a very cryptic error:
```
#
# Fatal error in , line 0
# Fatal JavaScript invalid size error 169220804 (see crbug.com/1201626)
#
#
#
#FailureMessage Object: 0x16bbd90f8
```

This was extremely hard to debug. A precondition check that throws a normal JS `Error` would be much appreciated.

eranhirsch:
1. Fixes #784
2. Adds typing for literal chunk sizes.
3. Slight optimizations to the runtime to handle cases where the chunk size is larger than the array, when it is 1, and to use sparse arrays to handle very large arrays more efficiently.


src/internal/types.ts
```diff
@@ -234,3 +234,46 @@ export type GuardType<T, Fallback = never> = T extends (
 ) => x is infer U
   ? U
   : Fallback;
+
+/**
+ * Creates an array with exactly N elements in it, with a possible Prefix.
```
cjquines, line 239: **question:** does `Prefix` here refer to the thing called `Result`?

cjquines (dismissed):

Issue #782, opened 2024-07-16
`clone` does not work with `File`s


jamesst20:
![Capture d’écran, le 2024-07-16 à 15 23 23](https://github.com/user-attachments/assets/20c1aaa3-6376-4223-92d9-1ac48cb7088f)

Version: Remeda 2.5.0

cjquines:
hm, as written `clone` only supports the javascript built in objects, and `File` isn't one of those

the correct way to clone a `File` is to [use its constructor again](https://developer.mozilla.org/en-US/docs/Web/API/File/File)

jamesst20:
> hm, as written `clone` only supports the javascript built in objects, and `File` isn't one of those
> 
> the correct way to clone a `File` is to [use its constructor again](https://developer.mozilla.org/en-US/docs/Web/API/File/File)

`structuredClone` works though, should this be fixed?

eranhirsch:
I don't have enough experience with File's, why would you need to clone them? do they hold state? does the structuredClone of a file makes sense?

eranhirsch:
Looking at the PR test failures, it looks like File isn't supported in Node 18. We still support that version, so we can't ship code that wouldn't run on it. We'll have to wait until we drop support for Node 18 in mid-2025.

jamesst20:
> I don't have enough experience with File's, why would you need to clone them? do they hold state? does the structuredClone of a file makes sense?

I don't actually need to clone the file, it's simply that I have in my Svelte app a state for my form. When I update my form (i.e an input, a checkbox, ...) I don't want my original object that comes from API to be updated so I clone it prior.

Sometimes when I have validation errors or stuff like that when I submit the form, that object gets updated so I need to rebuild the form object so I clone again. This is where I lose the file instance because Remeda drops it. I guess I could manually loop through all my attributes and reassign the same file instance, however it seems unexpected that I clone my object and I lose data. Like I said this is not happening with structuredClone so I have switched to that in the mean time.

eranhirsch:
Actually we can ship this in v2, see the attached PR (#1139)

PR #788, opened 2024-07-24
fix(clone): Support File instances

eranhirsch:
Fixes: #782

Exempting File instances like we with the other built-in types Date and RegExp...

I don't know exactly what the use-case is for cloning a File, so can't say if using structuredClone is the right approach here...


eranhirsch:
Fails on Node 18

PR #790, opened 2024-07-24
feat: Add toUpperCase, toLowerCase, capitalize, uncapitalize, toCamelCase, startsWith, endsWith

eranhirsch:
Closes: #509

Added the following ECMAScript built-ins:
* `toUpperCase`
* `toLowerCase`
* `startsWith`
* `endsWith`

The TypeScript utilities:
* `capitalize`
* `uncapitalize`

And the type-fest-based:
* `toCamelCase`

---

Make sure that you:


src/uncapitalize.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Makes first character of a string upper-case. Uses the built-in
+ * [`String.prototype.toLowerCase`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLowerCase)
+ * for the runtime, and the built-in [`Uncapitalize`](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html#Uncapitalizestringtype)
+ * utility type for typing.
+ *
+ * For other case manipulations see: `toUpperCase`, `toLowerCase`, `capitalize`,
+ * and `toCamelCase`.
+ *
```
cjquines, line 30: **question:** why not a similar note about non-ASCII characters?

src/toCamelCase.ts
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * For *PascalCase* use `capitalize(toCamelCase(data))`.
+ *
+ * @param data - A string.
+ * @param options - Used to disable the default behavior of preserving
+ * consecutive uppercase characters. This is optional.
+ * @signature
+ *   R.toCamelCase(data);
+ * @example
+ *   R.toCamelCase("hello world"); // "helloWorld"
```
cjquines, line 31: **question:** can we add more examples?

something like `R.toCamelCase("__HELLO_WORLD__")` or something

cjquines (dismissed):

src/uncapitalize.ts
eranhirsch: There's no "text-transform" value that would uncapitalize, so there was no alternative to offer to users.

PR #793, opened 2024-07-27
feat(randomInt): add randomInt

### LINKED ISSUE #783: [Request] Get random number between two values
Equivalent to https://lodash.com/docs/4.17.15#random
or https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random#getting_a_random_integer_between_two_values

StyleShit:
Closes #783

---

Make sure that you:

- [X] New methods added to `docs/src/content/mapping` for both Lodash and Ramda.

---

### Available Types:

existing).

</details>

### DIFF
--- a/docs/src/content/mapping/lodash/random.md
+++ b/docs/src/content/mapping/lodash/random.md
@@ -1,0 +1,14 @@
+---
+category: Number
+remeda: randomInt
+---
+
+- In Lodash, the `random` function supports generating either a random floating-point number or a random integer. In Remeda, only random integers are supported.
+
+```ts
+// Lodash
+random(1, 10);
+
+// Remeda
+randomInt(1, 10);
+```
--- a/src/randomInt.test-d.ts
+++ b/src/randomInt.test-d.ts
@@ -1,0 +1,14 @@
+import { randomInt } from "./randomInt";
+
+test("Both params are of the same type", () => {
+  // @ts-expect-error [ts2345] - Both should be number.
+  randomInt(0, 10n);
+
+  // @ts-expect-error [ts2345] - Both should be bigint.
+  randomInt(0n, 10);
+});
+
+test("Returns the same type as the params", () => {
+  expectTypeOf(randomInt(0, 10)).toEqualTypeOf<number>();
+  expectTypeOf(randomInt(0n, 10n)).toEqualTypeOf<bigint>();
+});
--- a/src/randomInt.test-d.ts
+++ b/src/randomInt.test-d.ts
@@ -12,3 +12,8 @@ test("Returns the same type as the params", () => {
   expectTypeOf(randomInt(0, 10)).toEqualTypeOf<number>();
   expectTypeOf(randomInt(0n, 10n)).toEqualTypeOf<bigint>();
 });
+
+test("Accepts one argument", () => {
+  expectTypeOf(randomInt(10)).toEqualTypeOf<number>();
+  expectTypeOf(randomInt(10n)).toEqualTypeOf<bigint>();
+});
--- a/src/randomInt.test.ts
+++ b/src/randomInt.test.ts
@@ -1,0 +1,47 @@
+import { randomInt } from "./randomInt";
+
+test("Infinity", () => {
+  expect(() => randomInt(Infinity, Infinity + 10)).toThrow(
+    new TypeError("from(Infinity) is not a finite number"),
+  );
+
+  expect(() => randomInt(10, Infinity)).toThrow(
+    new TypeError("to(Infinity) is not a finite number"),
+  );
+});
+
+test("NaN", () => {
+  expect(() => randomInt(Number.NaN, 10)).toThrow(
+    new TypeError("from(NaN) is not a finite number"),
+  );
+
+  expect(() => randomInt(10, Number.NaN)).toThrow(
+    new TypeError("to(NaN) is not a finite number"),
+  );
+});
+
+test("Non-integer", () => {
+  expect(() => randomInt(0.5, 10)).toThrow(
+    new TypeError("from(0.5) is not an integer"),
+  );
+
+  expect(() => randomInt(0, 10.5)).toThrow(
+    new TypeError("to(10.5) is not an integer"),
+  );
+});
+
+test("Invalid range", () => {
+  expect(() => randomInt(10, 0)).toThrow(
+    new RangeError("to(0) should be greater than from(10)"),
+  );
+});
+
+test("BigInt", () => {
+  expect(randomInt(0n, 10n)).toBeGreaterThanOrEqual(0);
+  expect(randomInt(0n, 10n)).toBeLessThanOrEqual(10);
+});
+
+test("Random integer", () => {
+  expect(randomInt(0, 10)).toBeGreaterThanOrEqual(0);
+  expect(randomInt(0, 10)).toBeLessThanOrEqual(10);
+});
--- a/src/randomInt.test.ts
+++ b/src/randomInt.test.ts
@@ -37,11 +37,19 @@ test("Invalid range", () => {
 });
 
 test("BigInt", () => {
-  expect(randomInt(0n, 10n)).toBeGreaterThanOrEqual(0);
-  expect(randomInt(0n, 10n)).toBeLessThanOrEqual(10);
+  expect(randomInt(0n, 10n)).toBeGreaterThanOrEqual(0n);
+  expect(randomInt(0n, 10n)).toBeLessThanOrEqual(10n);
 });
 
-test("Random integer", () => {
+test("Integer", () => {
   expect(randomInt(0, 10)).toBeGreaterThanOrEqual(0);
   expect(randomInt(0, 10)).toBeLessThanOrEqual(10);
+
+  expect(randomInt(10)).toBeGreaterThanOrEqual(0);
+  expect(randomInt(10)).toBeLessThanOrEqual(10);
+});
+
+test("Negative integers", () => {
+  expect(randomInt(-10, 10)).toBeGreaterThanOrEqual(-10);
+  expect(randomInt(-10, 10)).toBeLessThanOrEqual(10);
 });
--- a/src/randomInt.ts
+++ b/src/randomInt.ts
@@ -1,0 +1,52 @@
+/**
+ * Generate a random integer between `from` and `to`.
+ *
+ * @param from - The minimum value.
+ * @param to - The maximum value.
+ * @returns The random integer.
+ * @signature
+ *   R.randomInt(from, to)
+ * @example
+ *   R.randomInt(1, 10) // => 5
+ *   R.randomInt(1n, 10n) // => 7n
+ * @dataFirst
+ * @category Number
+ */
+export function randomInt<T extends bigint | number>(
+  from: T,
+  to: Widen<T>,
+): Widen<T> {
+  if (typeof from === "bigint" && typeof to === "bigint") {
+    return BigInt(
+      Math.floor(Number(to - from + BigInt(1)) * Math.random() + Number(from)),
+    ) as never;
+  }
+
+  if (!Number.isFinite(from)) {
+    throw new TypeError(`from(${from}) is not a finite number`);
+  }
+
+  if (!Number.isFinite(to)) {
+    throw new TypeError(`to(${to}) is not a finite number`);
+  }
+
+  if (!Number.isInteger(from)) {
+    throw new TypeError(`from(${from}) is not an integer`);
+  }
+
+  if (!Number.isInteger(to)) {
+    throw new TypeError(`to(${to}) is not an integer`);
+  }
+
+  if (to < from) {
+    throw new RangeError(`to(${to}) should be greater than from(${from})`);
+  }
+
+  // Cast the parameters to numbers since TypeScript can't infer it properly.
+  const _from = from as number;
+  const _to = to as number;
+
+  return Math.floor(Math.random() * (_to - _from + 1) + _from) as never;
+}
+
+type Widen<T> = T extends bigint ? bigint : number;
--- a/src/randomInt.ts
+++ b/src/randomInt.ts
@@ -6,45 +6,61 @@
  * @returns The random integer.
  * @signature
  *   R.randomInt(from, to)
+ *   R.randomInt(to)
  * @example
  *   R.randomInt(1, 10) // => 5
  *   R.randomInt(1n, 10n) // => 7n
+ *   R.randomInt(10) // => 5
+ *   R.randomInt(10N) // => 7n
  * @dataFirst
  * @category Number
  */
 export function randomInt<T extends bigint | number>(
   from: T,
-  to: Widen<T>,
+  to?: Widen<T>,
 ): Widen<T> {
-  if (typeof from === "bigint" && typeof to === "bigint") {
+  let lower: bigint | number;
+  let upper: bigint | number;
+
+  if (to === undefined) {
+    lower = typeof from === "bigint" ? 0n : 0;
+    upper = from;
+  } else {
+    lower = from;
+    upper = to;
+  }
+
+  if (typeof lower === "bigint" && typeof upper === "bigint") {
     return BigInt(
-      Math.floor(Number(to - from + BigInt(1)) * Math.random() + Number(from)),
+      Math.floor(
+        Number(upper - lower + BigInt(1)) * Math.random() + Number(lower),
+      ),
     ) as never;
   }
 
-  if (!Number.isFinite(from)) {
-    throw new TypeError(`from(${from}) is not a finite number`);
+  if (!Number.isFinite(lower)) {
+    throw new TypeError(`from(${lower}) is not a finite number`);
   }
 
-  if (!Number.isFinite(to)) {
-    throw new TypeError(`to(${to}) is not a finite number`);
+  if (!Number.isFinite(upper)) {
+    throw new TypeError(`to(${upper!}) is not a finite number`);
   }
 
-  if (!Number.isInteger(from)) {
-    throw new TypeError(`from(${from}) is not an integer`);
+  if (!Number.isInteger(lower)) {
+    throw new TypeError(`from(${lower}) is not an integer`);
   }
 
-  if (!Number.isInteger(to)) {
-    throw new TypeError(`to(${to}) is not an integer`);
+  if (!Number.isInteger(upper)) {
+    throw new TypeError(`to(${upper}) is not an integer`);
   }
 
-  if (to < from) {
-    throw new RangeError(`to(${to}) should be greater than from(${from})`);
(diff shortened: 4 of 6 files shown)


src/randomInt.ts
```diff
@@ -0,0 +1,68 @@
...
+ *   R.randomInt(to)
+ * @example
+ *   R.randomInt(1, 10) // => 5
+ *   R.randomInt(1n, 10n) // => 7n
+ *   R.randomInt(10) // => 5
+ *   R.randomInt(10n) // => 7n
+ * @dataFirst
+ * @category Number
+ */
+export function randomInt<T extends bigint | number>(
```
StyleShit, line 18: Wasn't sure whether I should `purry` this function or not, what's the convention?

cjquines: i would say this function shouldn't be purried, and in fact always accept two arguments

src/randomInt.ts
```diff
@@ -0,0 +1,68 @@
...
+    throw new TypeError(`to(${upper!}) is not a finite number`);
+  }
+
+  if (!Number.isInteger(lower)) {
+    throw new TypeError(`from(${lower}) is not an integer`);
+  }
+
+  if (!Number.isInteger(upper)) {
+    throw new TypeError(`to(${upper}) is not an integer`);
+  }
```
cjquines, line 55: **question:** is this necessary here? if the intent is to choose a random integer `n` such that `from <= n <= to`, then i think it's fine we allow this

src/randomInt.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * Generate a random integer between `from` and `to`.
```
cjquines, line 2: **nit:** let's add an "inclusive" here

src/randomInt.ts
StyleShit: I wasn't sure whether to support the single argument option, it's just what lodash has

removed

src/randomInt.ts
StyleShit: hmm, what happens when you try to `randomInt(1.1, 1.2)`?

src/randomInt.ts
StyleShit: Good idea, done :)

src/randomInt.ts
cjquines: good question! i think this should get the same error as `randomInt(2, 1)`, so we should word that differently? maybe something like "range has no integers" but phrased better lol

StyleShit: haha

added something, tell me what you think :)

eranhirsch:
when I played around with this issue, I also took a look at utilizing type-fest's [`IntRange`](https://github.com/sindresorhus/type-fest/blob/main/source/int-range.d.ts) to improve the typing. We can't just use the type as-is, as there are some interesting edge-cases, but it might be useful to make the type stricter for the few cases where it matters.

src/randomInt.ts
```diff
@@ -0,0 +1,50 @@
...
+): Widen<T> {
+  if (to <= from) {
+    throw new RangeError(`to(${to}) should be greater than from(${from})`);
+  }
+
+  if (typeof from === "bigint" && typeof to === "bigint") {
+    return BigInt(
+      Math.floor(Number(to - from + BigInt(1)) * Math.random() + Number(from)),
+    ) as never;
+  }
```
eranhirsch, line 27: This is not an accurate random number generator for `bigint`s because `Math.random` generates just shy of 2^62 different values. If our range of `bigint` is bigger than that (which it might be if the user is already using `bigint`) then this calculation will not result in an unbiased distribution of the values, as some of them will never be returned.

To make things simple I suggest removing the support for `bigint` from this function. We can consider creating a separate `randomBigInt` utility, but i'm not sure the built-in libraries in ECMAScript provide a way to generate it (it's solvable using the crypto module, but I don't think that's standard enough).

src/randomInt.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  if (!Number.isFinite(from)) {
+    throw new TypeError(`from(${from}) is not a finite number`);
+  }
+
+  if (!Number.isFinite(to)) {
+    throw new TypeError(`to(${to}) is not a finite number`);
+  }
```
eranhirsch, line 35: 1. These should be RangeError's, as they consider the range of values; they aren't the wrong type (infinity is a number).
2. As above, add the function name to the message.
3. I'm not sure this is worthwhile, as they add a check at runtime for *everyone* just to catch a really obscure edge case that would result in a semi-erroneous (`Infinity`) value that the user can handle downstream instead.
```suggestion
```

src/randomInt.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  // Cast the values to numbers since TypeScript can't infer it properly.
+  const _from = from as number;
+  const _to = to as number;
```
eranhirsch, line 39: once you drop the bigint support this should be fine.
```suggestion
```

src/randomInt.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  if (Math.floor(_from) === Math.floor(_to)) {
+    throw new RangeError(
+      `there are no integers between from(${from}) and to(${to})`,
+    );
+  }
```
eranhirsch, line 45: Why does this matter? if someone has some code that fetches two numbers (maybe a UI) and it results in calling `randomInt(2, 2)`, why shouldn't we just return `2` for them consistently, throwing here might break people's code.

We prefer not throwing unless the algorithm cannot work without preemptively throwing.

```suggestion
```

src/randomInt.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  return Math.floor(Math.random() * (_to - _from + 1) + _from) as never;
+}
+
+type Widen<T> = T extends bigint ? bigint : number;
```
eranhirsch, line 50: Define all types at the top of the file

src/randomInt.ts
```diff
@@ -0,0 +1,50 @@
...
+  const _from = from as number;
+  const _to = to as number;
+
+  if (Math.floor(_from) === Math.floor(_to)) {
+    throw new RangeError(
+      `there are no integers between from(${from}) and to(${to})`,
+    );
+  }
+
+  return Math.floor(Math.random() * (_to - _from + 1) + _from) as never;
```
eranhirsch, line 47: 1. why are we casting as `never`?
2. Let's use the MDN implementation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random#getting_a_random_integer_between_two_values_inclusive

src/randomInt.ts
```diff
@@ -0,0 +1,50 @@
...
+ *   R.randomInt(1, 10) // => 5
+ *   R.randomInt(1n, 10n) // => 7n
+ * @dataFirst
+ * @category Number
+ */
+export function randomInt<T extends bigint | number>(
+  from: T,
+  to: Widen<T>,
+): Widen<T> {
+  if (to <= from) {
```
eranhirsch, line 19: Because our range is inclusive of the edges, it's okay to call this function with the same value for `to` and `from` (`randomInt(2, 2)`), as it would result in a consistent value which makes sense. This might happen when the numbers are not fixed and are generated by other calculations.
```suggestion
  if (to < from) {
```

src/randomInt.ts
```diff
@@ -0,0 +1,50 @@
...
+    throw new RangeError(`to(${to}) should be greater than from(${from})`);
```
eranhirsch, line 20: add the function name to the error so it's easier to debug
```suggestion
    throw new RangeError(`randomInt: to(${to}) should be greater than from(${from})`);
```

docs/src/content/mapping/lodash/random.md
```diff
@@ -0,0 +1,15 @@
...
+- In Lodash, the `random` function supports generating either a random floating-point number or a random integer. In Remeda, only random integers are supported.
+- In Lodash, the `random` function supports accepting a single argument, which is the upper bound of the random number. In Remeda, the `randomInt` function requires both the lower and upper bounds as arguments.
+
+```ts
+// Lodash
+random(1, 10);
+
+// Remeda
+randomInt(1, 10);
+```
```
eranhirsch, line 15: We should add examples of each unique way our function differs from Lodash, this means we need to give an example where: `_.random()` and `_.random(max)`, not just `_.random(min, max)`, and give a solution for those that need the floating-point variant. You can link to this for that case: https://youmightnotneed.com/lodash#random

eranhirsch (changes requested):
* This is over-complicating what should be a very simple implementation, and has too many places where it might break user's code in production. We aim to throw as little as possible.

* The handling of bigint is not accurate and would lead users to bad assumptions. I prefer to remove it.

* We can play around with type-fest types (`IntRange`, `GreaterThan`, etc...) to make the typing even more useful.

src/randomInt.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Generate an inclusive random integer between `from` and `to`.
```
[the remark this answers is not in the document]
StyleShit: Thanks, done

src/randomInt.ts
StyleShit: Hmm, makes sense

Removed

src/randomInt.ts
StyleShit: Removed

src/randomInt.ts
StyleShit: Removed

src/randomInt.ts
StyleShit: it was mostly for cases like `randomInt(1.5, 1.6)`
what should be returned in this case?
the MDN implementation returns 2 (which is not what I would've expected TBH, I think it should _probably_ be 1)

src/randomInt.ts
StyleShit: Done

src/randomInt.ts
StyleShit: See https://github.com/remeda/remeda/pull/793#discussion_r1704397718

src/randomInt.ts
StyleShit: Done

src/randomInt.ts
StyleShit: Done

docs/src/content/mapping/lodash/random.md
StyleShit: Good point

Are we gonna have a `randomFloat` at some point?
because for now I'm referencing to this link as a solution

StyleShit:
> We can play around with type-fest types (IntRange, GreaterThan, etc...) to make the typing even more useful.

Sounds great to me
Added those types, but I wasn't sure whether I should implement the floats cases, so I just returned `number` to make things simple

eranhirsch (changes requested):
Please make sure all the checks are green before asking for review, you can put the PR in draft while you are working on it.

StyleShit:
@eranhirsch 
It was intentional, see https://github.com/remeda/remeda/pull/793#discussion_r1704397718

src/randomInt.ts
eranhirsch: If the result of `Math.ceil(from)` is smaller than the result of `Math.floor(to)` you should throw a RangeError, and shouldn't throw an error for any other case.

src/randomInt.ts
```diff
@@ -0,0 +1,51 @@
...
+export function randomInt<From extends number, To extends number>(
+  from: From,
+  to: To,
+): RandomInt<From, To> {
+  const fromCeiled = Math.ceil(from);
+  const toFloored = Math.floor(to);
+
+  if (toFloored < fromCeiled) {
+    throw new RangeError(
+      `randomInt: to(${toFloored}) should be greater than from(${fromCeiled})`,
```
eranhirsch, line 44: Don't provide the rounded parameters; the user might not be aware that this is done and might not be able to find the bug. Provide the parameters as they were provided by the user.
```suggestion
      `randomInt: The range [${to},${from}] contains no integers`,
```

eranhirsch (changes requested):
I'm missing some tests, specifically for usage with simple numbers (and not constants).

src/randomInt.ts
StyleShit: Sounds better
done

src/randomInt.ts
```diff
@@ -0,0 +1,51 @@
...
+        : GreaterThan<From, 0> extends true
+          ? LessThan<To, 1000> extends true
+            ? // type-fest's `InRange` supports only numbers between 0 and 1000 (exclusive)
+              IntRange<From, To> | To
+            : number
+          : number;
+
+type IsFloat<T extends number> = `${T}` extends `${number}.${number}`
+  ? true
+  : false;
```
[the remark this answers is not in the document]
StyleShit: Done

src/randomInt.ts
```diff
@@ -0,0 +1,51 @@
...
+    : IsFloat<To> extends true
+      ? number
+      : GreaterThan<From, To> extends true
+        ? never
+        : GreaterThan<From, 0> extends true
+          ? LessThan<To, 1000> extends true
+            ? // type-fest's `InRange` supports only numbers between 0 and 1000 (exclusive)
+              IntRange<From, To> | To
+            : number
+          : number;
```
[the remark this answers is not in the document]
StyleShit: Done

src/randomInt.test.ts
```diff
@@ -0,0 +1,28 @@
+import { randomInt } from "./randomInt";
+
+test("Invalid range", () => {
+  expect(() => randomInt(10, 0)).toThrow(
+    new RangeError("randomInt: to(0) should be greater than from(10)"),
+  );
+
+  expect(() => expect(randomInt(1.5, 1.6))).toThrow(
+    new RangeError("randomInt: to(1) should be greater than from(2)"),
+  );
```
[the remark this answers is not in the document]
StyleShit: Done

src/randomInt.test-d.ts
```diff
@@ -0,0 +1,20 @@
+import { randomInt } from "./randomInt";
+
+test("Returns number for floats", () => {
+  expectTypeOf(randomInt(0.5, 1)).toEqualTypeOf<number>();
+  expectTypeOf(randomInt(1, 1.5)).toEqualTypeOf<number>();
+});
+
```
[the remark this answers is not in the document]
StyleShit: Done

src/randomInt.test-d.ts
```diff
@@ -0,0 +1,20 @@
...
+test("Returns never for invalid range", () => {
+  expectTypeOf(randomInt(10, 1)).toEqualTypeOf<never>();
+});
+
+test("Returns range for valid range", () => {
+  expectTypeOf(randomInt(1, 5)).toEqualTypeOf<1 | 2 | 3 | 4 | 5>();
+  expectTypeOf(randomInt(1, 1)).toEqualTypeOf<1>();
```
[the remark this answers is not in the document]
StyleShit: Done

src/randomInt.test-d.ts
```diff
@@ -0,0 +1,20 @@
...
+});
+
+test("Returns number for invalid range", () => {
+  expectTypeOf(randomInt(0, 5)).toEqualTypeOf<number>();
```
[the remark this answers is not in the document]
StyleShit: Oops
I misread the documentation
fixed

src/randomInt.test-d.ts
```diff
@@ -0,0 +1,20 @@
...
+  expectTypeOf(randomInt(1, 1000)).toEqualTypeOf<number>();
```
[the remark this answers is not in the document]
StyleShit: Done

src/randomInt.test-d.ts
```diff
@@ -0,0 +1,20 @@
+import { randomInt } from "./randomInt";
+
```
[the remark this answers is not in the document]
StyleShit: great test cases
done

src/randomInt.test.ts
```diff
@@ -0,0 +1,61 @@
...
+test("integers with same value", () => {
+  const randoms = createRandomsArray(10, 10);
+
+  for (const v of randoms) {
+    expect(v).toBe(10);
+  }
+});
+
+function createRandomsArray(from: number, to: number): Array<number> {
+  return Array.from({ length: 10 }).map(() => randomInt(from, to));
```
eranhirsch, line 60: We can use a higher count here to increase the likelihood of catching a bug.
```suggestion
  return Array.from({ length: 100 }).map(() => randomInt(from, to));
```

src/randomInt.test.ts
```diff
@@ -0,0 +1,61 @@
+import { randomInt } from "./randomInt";
```
eranhirsch, line 1: Great tests!

src/randomInt.test-d.ts
```diff
@@ -0,0 +1,64 @@
...
+test("returns the same value when to equals from", () => {
+  expectTypeOf(randomInt(1, 1)).toEqualTypeOf<1>();
+});
+
+test("returns the same value when to equals from and out of range", () => {
+  expectTypeOf(randomInt(10_000, 10_000)).toEqualTypeOf<10_000>();
+});
+
+test("Returns number for invalid range (to >= 1000)", () => {
```
eranhirsch, line 28: the range isn't invalid, just out of range for our more refined type.
```suggestion
test("Returns number for large ranges (to >= 1000)", () => {
```

docs/src/content/mapping/lodash/random.md
```diff
@@ -0,0 +1,22 @@
...
+
+// Remeda
+randomInt(1, 10);
+
+// Lodash
+random(10);
+
+// Remeda
+randomInt(0, 10);
+```
```
eranhirsch, line 22: 1. Split the cases and give each one a header (check the docs for `forEach` for an example).
2. Add another example for `random()` (`randomInt(0, 1)`).

eranhirsch (approved):
Looks great! thanks for helping out!

src/randomInt.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Generate a random integer between `from` and `to` (inclusive).
+ *
+ * @param from - The minimum value.
+ * @param to - The maximum value.
+ * @returns The random integer.
+ * @signature
+ *   R.randomInt(from, to)
+ * @example
+ *   R.randomInt(1, 10) // => 5
+ *   R.randomInt(1.5, 2.6) // => 1
```
cjquines, line 35: **todo:** this should be `// => 2`

cjquines (dismissed):
thanks! i'll merge once comments are responded to

gustavopch:
So in the end it was named `randomInteger` instead of `randomInt`?

cjquines:
ah, you're right. i'll open an issue

Issue #794, opened 2024-07-28
`drop` types are not specific


StyleShit:
### Expected Behavior:

```TS
const a = drop([1, 2, 3, 4], 2);
   // ^? [3, 4]
```

### Actual Behavior:

```TS
const a = drop([1, 2, 3, 4], 2);
   // ^? number[]
```

PR #795, opened 2024-07-28
feat(drop): refined return type for literal Ns

StyleShit:
Partially handles #794 

I couldn't figure out why it doesn't work properly when using `pipe`
Is it an issue in the `pipe` function typing? IDK

---

Make sure that you:

- [X] New methods added to `docs/src/content/mapping` for both Lodash and Ramda.

---

### Available Types:

existing).

</details>


src/drop.test-d.ts
```diff
@@ -0,0 +1,48 @@
+import { drop } from "./drop";
+import { pipe } from "./pipe";
+
+describe("data first", () => {
```
cjquines, line 4: **suggestion:** can we add a test for `drop([1, 2, 3, 4, 5], 2 as number)`?

src/drop.ts
```diff
@@ -52,3 +59,17 @@ function lazyImplementation<T>(n: number): LazyEvaluator<T> {
     return { done: false, hasNext: true, next: value };
   };
 }
+
+type Drop<
+  Arr extends Array<unknown>,
+  Count extends number,
+  Removed extends Array<unknown> = [],
+> = `${Count}` extends `-${number}`
```
cjquines, line 67: **question:** can we try using the `ArraySlice` type from `type-fest`? i'd rather we reuse types from them unless we have a good reason not to

cjquines (commented):
i don't think there's a way to make it work for pipes nicely?

do you think we can do something similar for `take`?

src/drop.test-d.ts
StyleShit: added

src/drop.ts
StyleShit: Changed the code to use it, but seems like some cases resolve in a weird way
See all the pipe tests, and the `as number` tests

StyleShit:
> i don't think there's a way to make it work for pipes nicely?

Doesn't seem so 😓 
Using `ArraySlice` from `type-fest` returns an empty array for pipes

> do you think we can do something similar for take?

Sure, will open a separate PR when I'll have some time 😄

src/drop.test-d.ts
```diff
@@ -0,0 +1,46 @@
...
+  it("works when dropping more than the length of the array", () => {
+    expectTypeOf(drop([1, 2, 3, 4, 5], 10)).toEqualTypeOf<[]>();
+  });
+
+  it("works with generalized typed arrays", () => {
+    expectTypeOf(drop([] as Array<number>, 2)).toEqualTypeOf<[]>();
+  });
+
+  it("works with generalized typed count", () => {
+    expectTypeOf(drop([1, 2, 3, 4, 5], 2 as number)).toEqualTypeOf<never>();
```
cjquines, line 26: **question:** can we change the return type to avoid this case?

we can still base our type off `ArraySlice<T, N, T["length"]>`, maybe something like check if `N` is literal, use the arrayslice, otherwise return generic type

StyleShit: Good idea, done :)

The whole `pipe` issue is solvable BTW
we'll need to add `pipe<const A, B, ...>` to the signature, but it breaks other functions that use it, so I'm not sure what's the best way to handle this

src/drop.ts
```diff
@@ -52,3 +58,9 @@ function lazyImplementation<T>(n: number): LazyEvaluator<T> {
     return { done: false, hasNext: true, next: value };
   };
 }
+
+// `ArraySlice<[1, 2, 3], number>` returns `never`, so we handle
+// this case by returning `Array<1 | 2 | 3>` instead.
+type Drop<T extends Array<unknown>, N extends number> = number extends N
+  ? Array<T[number]>
+  : ArraySlice<T, N, T["length"]>;
```
eranhirsch, line 66: Place types at the top of the file.

src/drop.ts
```diff
@@ -52,3 +58,9 @@ function lazyImplementation<T>(n: number): LazyEvaluator<T> {
     return { done: false, hasNext: true, next: value };
   };
 }
+
+// `ArraySlice<[1, 2, 3], number>` returns `never`, so we handle
+// this case by returning `Array<1 | 2 | 3>` instead.
+type Drop<T extends Array<unknown>, N extends number> = number extends N
```
eranhirsch, line 64: Use our IterableContainer
```suggestion
type Drop<T extends IterableContainer, N extends number> = number extends N
```

src/drop.ts
```diff
@@ -15,7 +16,10 @@ import { purry } from "./purry";
  * @lazy
  * @category Array
  */
-export function drop<T>(array: ReadonlyArray<T>, n: number): Array<T>;
+export function drop<const T extends Array<unknown>, N extends number>(
```
eranhirsch, line 19: We currently don't use `const` in our functions, and for consistency we shouldn't here either.
```suggestion
export function drop<T extends Array<unknown>, N extends number>(
```

src/drop.ts
```diff
@@ -35,8 +39,10 @@ export function drop(...args: ReadonlyArray<unknown>): unknown {
   return purry(dropImplementation, args, lazyImplementation);
 }
 
-const dropImplementation = <T>(array: ReadonlyArray<T>, n: number): Array<T> =>
-  n < 0 ? [...array] : array.slice(n);
+const dropImplementation = <const T extends Array<unknown>, N extends number>(
+  array: T,
+  n: N,
+): Drop<T, N> => (n < 0 ? [...array] : array.slice(n)) as never;
```
eranhirsch, line 45: Don't use `as never`. It's not better than disabling the typescript error via `// @ts-expect error`, which makes the typing issue more obvious.

anyway, because this is the implementation and it's not always easy to make typescript aware of complex types return the trivial type.
```suggestion
const dropImplementation = <T extends IterableContainer>(
  array: T,
  n: number,
): Array<T[number]> => (n < 0 ? [...array] : array.slice(n));
```

src/drop.test-d.ts
```diff
@@ -0,0 +1,54 @@
...
+  it("works on regular inputs", () => {
+    expectTypeOf(drop([1, 2, 3, 4, 5], 2)).toEqualTypeOf<[3, 4, 5]>();
+  });
+
+  it("works on empty arrays", () => {
+    expectTypeOf(drop([], 2)).toEqualTypeOf<[]>();
+  });
+
+  it("works with negative numbers", () => {
+    expectTypeOf(drop([1, 2, 3, 4, 5], -1)).toEqualTypeOf<[5]>();
```
eranhirsch, line 14: This is an incorrect test! The runtime doesn't do this, it would return the array as is, e.g. [1,2,3,4,5]

src/drop.test-d.ts
```diff
@@ -0,0 +1,54 @@
+import { drop } from "./drop";
+import { pipe } from "./pipe";
+
+describe("data first", () => {
+  it("works on regular inputs", () => {
+    expectTypeOf(drop([1, 2, 3, 4, 5], 2)).toEqualTypeOf<[3, 4, 5]>();
```
eranhirsch, line 6: when dropping the "const" from the function generics this would become
```suggestion
    expectTypeOf(drop([1, 2, 3, 4, 5], 2)).toEqualTypeOf<[number, number, number]>();
```

src/drop.test-d.ts
```diff
@@ -0,0 +1,54 @@
...
+  it("works with negative numbers", () => {
+    expectTypeOf(drop([1, 2, 3, 4, 5], -1)).toEqualTypeOf<[5]>();
+  });
+
+  it("works when dropping more than the length of the array", () => {
+    expectTypeOf(drop([1, 2, 3, 4, 5], 10)).toEqualTypeOf<[]>();
+  });
+
+  it("works with generalized typed arrays", () => {
+    expectTypeOf(drop([] as Array<number>, 2)).toEqualTypeOf<[]>();
```
eranhirsch, line 22: this is incorrect too, the return type for dropping from a generic array should be that array, and not an empty array...

eranhirsch (changes requested):
The tests here assert the wrong return values, that means that the type is broken for a lot of cases. Please go over the tests and make sure that the **runtime** behavior matches the typing.

Also, I don't see why this shouldn't work on dataLast, we have many similar types that work the same way.

```ts
function drop<N extends number>(n: N): <T extends IterableContainer>(data: T) => Drop<T, N>;
```

src/drop.ts
StyleShit: Done

src/drop.ts
StyleShit: Done

src/drop.ts
StyleShit: it means we lose some of the typings (will be typed as `[number, number]` instead of `[1, 2]`)

but sure, let's go for it, for the sake of consistency 😄

src/drop.ts
StyleShit: Sure
Done

src/drop.test-d.ts
StyleShit: Ooh, how did I miss that
good catch, thanks!

src/drop.test-d.ts
StyleShit: Yup
Done

src/drop.test-d.ts
StyleShit: Fixed

StyleShit:
> Also, I don't see why this shouldn't work on dataLast, we have many similar types that work the same way.

That's exactly what I did earlier and it didn't work 😞 
It seems to _kinda_ work when patching the `pipe` function, but I'm assuming we don't want to...

Any references to similar places where it works properly?

src/drop.ts
```diff
@@ -1,6 +1,16 @@
...
 import { purry } from "./purry";
+import { type IterableContainer } from "./internal/types";
+
+type Drop<T extends IterableContainer, N extends number> = number extends N
+  ? Array<T[number]>
+  : `${N}` extends `-${number}`
+    ? T
+    : number extends T["length"]
+      ? Array<T[number]>
+      : ArraySlice<T, N, T["length"]>;
```
eranhirsch, line 13: This solution doesn't solve the most useful case for a better type: when an array has a variable length but still has some prefix we can maintain; for example: `[string, string, string, ...Array<string>]`. In this case, I would expect `drop` with an `N` of 1 to return the type `[string, string, ...Array<string>]`, which your solution doesn't do.

Some limitations with type-fest's `ArraySlice` (and `ArraySplice`) don't make them an easy tool to solve this. I know we suggested using them in previous review comments, but I now think that was a mistake.

Solving `drop's return type has been requested in the past, and I pushed against it because it was difficult.

Fortunately, we recently added some utilities that make the type for drop easier. The solution is still complex, so I prefer just providing the solution, but if you have questions to better understand it, feel free.

```suggestion
    : [
        ...(TupleParts<T>["prefix"] extends readonly [
          ...NTuple<unknown, N>,
          ...infer V,
        ]
          ? V
          : []),
        ...IfNever<
          TupleParts<T>["item"],
          [],
          [...Array<TupleParts<T>["item"]>, ...TupleParts<T>["suffix"]]
        >,
      ];
```

src/drop.test-d.ts
```diff
@@ -0,0 +1,58 @@
...
+      Array<number | string>
+    >();
+  });
+
+  it("works with generalized typed count", () => {
+    expectTypeOf(drop([1, "2", true], 2 as number)).toEqualTypeOf<
+      Array<boolean | number | string>
+    >();
+  });
+});
```
eranhirsch, line 34: please also add a test for "prefix" tuples
```suggestion
  it("works with prefix arrays", () => {
    const result = drop(["hello", "world"] as [string, string, ...Array<string>], 1);
    expectTypeOf(result).toEqualTypeOf<[string, ...Array<string>]>();
  });
});
```

src/drop.test-d.ts
```diff
@@ -0,0 +1,58 @@
...
+      Array<number>
+    >();
+  });
+
+  it("works when dropping more than the length of the array", () => {
+    expectTypeOf(pipe([1, 2, 3, 4, 5], drop(10))).toEqualTypeOf<
+      Array<number>
+    >();
+  });
+});
```
eranhirsch, line 58: Your dataLast tests don't test what you think they test, you'd need to use "as ..." in the type (like you did with one of them) to make sure that the type passed to the function is of the correct type. `pipe` would generalize the type when a const is provided.

```suggestion
describe("data last", () => {
  it("works on regular inputs", () => {
    expectTypeOf(pipe([1, 2, 3, 4, 5], drop(2))).toEqualTypeOf<Array<number>>();
  });

  it("works on empty arrays", () => {
    expectTypeOf(pipe([] as Array<string>, drop(2))).toEqualTypeOf<
      Array<string>
    >();
  });

  it("works with negative numbers", () => {
    expectTypeOf(pipe([1, 2, 3, 4, 5], drop(-1))).toEqualTypeOf<
      Array<number>
    >();
  });

  it("works when dropping more than the length of the array", () => {
    expectTypeOf(pipe([1, 2, 3, 4, 5], drop(10))).toEqualTypeOf<
      Array<number>
    >();
  });
});
```

eranhirsch (changes requested):
Typing `drop` is hard, that's why it hasn't been done yet. Your implementation still has some cases where it doesn't work well. I've provided my own solution so we can move forward.

Regarding using `const T` in function generics, and the data-last implementation "not working", this stems from a misunderstanding of how types are inferred by typescript. when someone calls `foo([1, 2, 3])`, if the param is typed with a generic typescript needs to infer what that generic type is, and it can do it differently when the generic is defined with a `const T` vs `T`, mainly whether it needs to be the most specific, or the most wide. In Remeda (atm) we use the wide approach, this also works for our users and I haven't seen any issues with it.

It looks like a bigger issue to you because you are writing very simple tests that don't define an explicit type for the params you provide to the function, in most real-life cases the params are well defined, and thus typescript infers them correctly based on that.

src/drop.test-d.ts
StyleShit: Not sure that I understand what you mean
but done

src/drop.test-d.ts
StyleShit: Done

src/drop.ts
StyleShit: Done

src/drop.ts
```diff
@@ -29,14 +42,18 @@ export function drop<T>(array: ReadonlyArray<T>, n: number): Array<T>;
...
+  n: N,
+): <T extends IterableContainer>(data: T) => Drop<T, N>;
 
 export function drop(...args: ReadonlyArray<unknown>): unknown {
   return purry(dropImplementation, args, lazyImplementation);
 }
 
-const dropImplementation = <T>(array: ReadonlyArray<T>, n: number): Array<T> =>
-  n < 0 ? [...array] : array.slice(n);
+const dropImplementation = <T extends Array<unknown>, N extends number>(
```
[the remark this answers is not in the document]
StyleShit: Done

eranhirsch (approved):
Thanks for adding this, I know a lot of people have asked for it

PR #811, opened 2024-08-03
docs: Updates type hints in functional data last example

dougflip:
Update the type hints (displayed as comments) to reflect that TypeScript does properly infer the types when using Remeda in this example.

### DIFF
--- a/docs/src/pages/index.mdx
+++ b/docs/src/pages/index.mdx
@@ -110,9 +110,9 @@ const DATA = [
 R.pipe(
   DATA,
   R.filter((x) => x.gender === "f"),
-  //              ^? 😔 'x' is of type 'unknown'
+  //              ^? (parameter) x: { name: string, age: number; gender: string }
   R.groupBy((x) => x.age),
-  //                 ^? 😔 Property 'age' does not exist on type 'P'
+  //                 ^? (property) age: number
 );
 
 // Ramda
--- a/docs/src/pages/index.mdx
+++ b/docs/src/pages/index.mdx
@@ -132,7 +132,7 @@ _.flow(
   _.filter((x) => x.gender === "f"),
   //              ^? 😔 'x' is of type 'unknown'
   _.groupBy((x) => x.age),
-  //                 ^? 😔 Property 'age' does not exist on type 'P'
+  //               ^? 😔 'x' is of type 'unknown'
 )(DATA);
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/pages/index.mdx
```diff
@@ -110,9 +110,9 @@ const DATA = [
 R.pipe(
   DATA,
   R.filter((x) => x.gender === "f"),
-  //              ^? 😔 'x' is of type 'unknown'
+  //              ^? (parameter) x: { name: string, age: number; gender: string }
   R.groupBy((x) => x.age),
-  //                 ^? 😔 Property 'age' does not exist on type 'P'
+  //                 ^? (property) age: number
```
dougflip, line 115: I was reading through the docs and the remeda section seemed to be a
copy/paste mistake from the lodash/fp section below. I copied the
example over to TS Playground to verify that TS does in fact properly
infer the types.

cjquines: ah, whoops. i think this is actually supposed to be for Ramda, can you fix?

docs/src/pages/index.mdx
```diff
@@ -110,15 +110,15 @@ const DATA = [
...
   R.groupBy((x) => x.age),
-  //                 ^? 😔 Property 'age' does not exist on type 'P'
 );
 
 // Ramda
 R.pipe(
   R.filter((x) => x.gender === "f"),
+  //              ^? 😔 'x' is of type 'unknown'
   R.groupBy((x) => x.age),
+  //                 ^? 😔 Property 'age' does not exist on type 'P'
```
dougflip, line 121: @cjquines I moved this to the Ramda section - good call :+1:

docs/src/pages/index.mdx
```diff
@@ -132,7 +132,7 @@ _.flow(
   _.filter((x) => x.gender === "f"),
   //              ^? 😔 'x' is of type 'unknown'
   _.groupBy((x) => x.age),
-  //                 ^? 😔 Property 'age' does not exist on type 'P'
+  //               ^? 😔 'x' is of type 'unknown'
```
dougflip, line 135: I noticed this was slightly off from what I observed. Probably doesn't make a big difference, but while I was in there I thought I would tweak it. Totally fine if you want me to revert or even dump this whole line if you don't think it adds much (since it is the same error as above).

TS Playground for reference if you want to see the errors: https://tinyurl.com/569s526w

cjquines (approved):

cjquines:
thanks!

dougflip:
> thanks!

np - glad to help! This is such a great lib, thank you and team for all the hard work :muscle:

Issue #796, opened 2024-07-31
takeLast behaves irrationally when trying to take more than array.length


kkurola:
In cases where the array does not have sufficient number of elements to take, the slice will "overflow" and returns unexpected results. Intuitively I would expect it to behave similar to Lodash takeRight where you get the whole array back.

Implementation for quick reference:
```ts
const takeLastImplementation = <T>(
  array: ReadonlyArray<T>,
  n: number,
): Array<T> => (n >= 0 ? array.slice(array.length - n) : []);
```

Compare Remeda:
`R.takeLast([1, 2, 3, 4], 5)` -> `[4]`
`R.takeLast([1, 2, 3, 4], 6)`-> `[3, 4]`
`R.takeLast([1, 2, 3, 4], 7)`-> `[2, 3, 4]`

Versus Lodash:
`_.takeRight([1, 2, 3, 4], 5)` -> `[1, 2, 3, 4]`

Edit:

Seems `dropLast` behaves similarly. Instead of empty array return you get starting values
`R.dropLast([1, 2, 3, 4], 5)` -> `[1, 2, 3]`

ajuvonen:
I was coming here to report the same issue but you beat me to it. I would also expect to get the whole array back in the described scenario, and this is also how takeLast behaves in Ramda.

eranhirsch:
Your analysis sounds right, I'll check this out over the weekend and offer a solution, thanks for reporting this.

PR #812, opened 2024-08-05
fix(dropLast, takeLast): handle overflows

eranhirsch:
Closes: #796


kkurola:
Thanks for the quick fix, looks solid 🙇

cjquines (approved):

Issue #457, opened 2024-01-10
Add string manipulation utilities like camelCase, startCase, etc


viveleroi:
We're hoping to replace lodash and ramda entirely with remeda. There are a few functions remeda doesn't have that I believe it could benefit from.

Common string manipulation methods are one of the reasons we still use lodash in addition to ramda - ramda decided those kinds of utility functions weren't appropriate for the library. I'm hoping remeda is willing to add them.

The below links are for camelCase but these would easily apply to all similar functions.

[Lodash docs](https://lodash.com/docs/4.17.15#camelCase)
[Lodash source](https://github.com/lodash/lodash/blob/main/src/camelCase.ts)

eranhirsch:
I'm supportive of this, can you send a PR?

cjquines:
related: #525

Brodzko:
Does this mean that the stance taken in [this issue](https://github.com/remeda/remeda/issues/207#issuecomment-1213081826) has changed now? If that is the case, I'm happy to help

eranhirsch:
We have more capacity to support these requests now and can accept more additions to the library. Take a look at the open PRs before you send something out, there might be a PR open already for some of these cases

AndreaPontrandolfo:
I think only a subset of Lodash string utilities should be prioritized to make it into Remeda. Since Lodash was created, some string utilities were made part of Ecmascript and so made Lodash alternatives obsoletes.

`startWith`: not needed. There is already a vanilla counterpart.
`endsWith`: not needed. There is already a vanilla counterpart.
`split`: not needed. There is already a vanilla counterpart.
`replace`: not needed. There is already a vanilla counterpart.
`parseInt `: not needed. There is already a vanilla counterpart.
`trim`: not needed. There is already a vanilla counterpart.
`trimStart`: not needed. There is already a vanilla counterpart.
`trimEnd`: not needed. There is already a vanilla counterpart.
`camelCase`: yes.
`snakeCase `: yes.
`kebabCase `: yes.
`truncate `: yes.
`startCase `: yes. Note: In some other libs this is also called "titleCase".
`capitalize `: yes. [PR open](https://github.com/remeda/remeda/pull/525)

Im not gonna mention the other string methods as they are not really common or useful.

But in general, yes, i strongly believe these types of utilities should be part of Remeda, they are extremely handy in Lodash.

eranhirsch:
All functions we were considering for this are now part of the library

PR #818, opened 2024-08-07
feat(toSnakeCase): add toSnakeCase

filipw01:
Added toSnakeCase function

Related issue: https://github.com/remeda/remeda/issues/457

The js implementation is fairly simple, but I struggle with types a bit. Type-fest has `SnakeCase` but as visible in the tests it has some issues. Using `SnakeCase<CamelCase<T, { preserveConsecutiveUppercase: true }>>` was yielding much better effects, but it is not ideal. We'll probably need to make a PR to `type-fest` but I'm optimistic that it can be done with an implementation more similar to CamelCase but tailored for SnakeCase

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/docs/src/content/mapping/lodash/__MISSING.md
+++ b/docs/src/content/mapping/lodash/__MISSING.md
@@ -142,7 +142,6 @@ TODO: Go over: https://you-dont-need.github.io/You-Dont-Need-Lodash-Underscore/#
 - parseInt
 - repeat
 - replace
-- snakeCase
 - startCase
 - template
 - trim
--- a/docs/src/content/mapping/lodash/snakeCase.md
+++ b/docs/src/content/mapping/lodash/snakeCase.md
@@ -1,0 +1,4 @@
+---
+category: String
+remeda: toSnakeCase
+---
--- a/src/index.ts
+++ b/src/index.ts
@@ -136,6 +136,7 @@ export * from "./tap";
 export * from "./times";
 export * from "./toCamelCase";
 export * from "./toLowerCase";
+export * from "./toSnakeCase";
 export * from "./toUpperCase";
 export * from "./uncapitalize";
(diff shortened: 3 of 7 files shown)


filipw01:
https://github.com/sindresorhus/type-fest/pull/930

There's also an open question what should happen when there are numbers in the string, I guess we can follow whatever is decided in type-fest

eranhirsch:
I'm not sure your issue with type-fest covers the main problem with this type, that it returns a union of possible definitions of snake_case, and not a single value, e.g. for the input "parseHTML" the type-fest type accepts more than one possible interpretation of snake_case, which is not good for a return type:

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgZQHYEMDWBTAwugZ2zgF84AzKCEOAIhgE8xsBac7AmWgbgChfGzOAEEAjHAC8KDDnxEAPLTDooRABIAVALIAZWgD4+AeiNwzAPQD8QA

filipw01:
This particular case is covered 😉  So as long as my PR is merged in type-fest we can progress with this one

eranhirsch:
With #924 now providing better types for snake and kebab cases, is there anything this PR aims to do that still needs to be provided? It looks like the type-fest PRs are stuck, and I'm not sure it's still worthwhile pursuing.

filipw01:
I think this PR is no longer needed. The only addition I can think of adding is to support not splitting on numbers like here in the test cases at the bottom https://github.com/sindresorhus/type-fest/pull/930/files#diff-1dfa4b7476785567bfe17d478e30cf2b76fffc23fe49ba264f08b95473fe073d

We could revisit this after type-fest PRs are merged around this, but I'll close this one

eranhirsch:
Thanks for your involvement!

PR #826, opened 2024-08-12
docs(readme): add contributors

eranhirsch:
@cjquines, wdyt?


cjquines (approved):
think we can give it an h2 :)

Issue #822, opened 2024-08-09
`debounce` does not call on trailing edge in "leaky bucket" case


filipw01:
I noticed there's already a debounce function (which I think in lodash can be used to simulate throttle) but it seems that remeda has a different behavior for the maxWait

See how I was able to get the throttle behavior of lodash using debounce https://stackblitz.com/edit/vitejs-vite-egpikb?file=src%2Fmain.ts

What do you think about adding equivalent of [docs-lodash.com/v4/throttle](https://docs-lodash.com/v4/throttle/) or changing debounce to closer align with the one in lodash (so that we could replicate throttle behavior)?

eranhirsch:
Hey, I don't remember the specifics when I added debounce, but I *think* I made it compatible with the lodash impl, can you say in what way doesn't it work to implement throttle the same way lodash's debounce can be used to implement it?

filipw01:
Did you take a look at the stackblitz I shared? I was able to replicate throttle behavior with lodash debounce, but not remeda debounce. If there's a way to do it do you mind sharing it? We could add it to the list of equivalent functions for migration from lodash page
<img width="425" alt="Screenshot 2024-08-12 at 21 15 03" src="https://github.com/user-attachments/assets/fba7ddea-efed-4504-97a8-a19ebc89703d">

cjquines:
oh, i see why. let's say our throttle time is 100 and we're calling the function every 90 ms. then:

* the function is called at times 90, 180, 270, 360, 450
* lodash invokes at times 90, 190, 360, 460
* remeda invokes at times 90, 280, 460

i think the remeda implementation is in fact bugged when `maxWaitMs` is set and `timing` is `"both"`

aretrace:
@cjquines, @SirJayEarn had `throttle` equivalent behavior in a [roadmap](https://github.com/remeda/remeda/issues/37) some time ago, any timeline for when this might be resolved?

eranhirsch:
This is my repro. It sheds more light on which invocations are being delayed and suppressed. I just started looking at it, so I don't have a suggestion yet.

[...]

PR #848, opened 2024-09-06
feat: add `funnel`, an execution timing utility

eranhirsch:
A replacement and extension of the existing `debounce` utility that provides more control over timing parameters, fixes throttling bugs, allows collecting parameters to allow more flexibility in processing, and removes direct management of return value state.

Fixes: #822

!Important: This massive PR requires a lot of attention. Despite its size, most of this PR was done manually (not via copy-paste or generated code), test-by-test. Reviewing this test by test doesn't make sense, but at least all the content blocks (comments, docs) and the implementation should be reviewed.

---

Make sure that you:

---

### Available Types:

existing).

</details>


cjquines:
wait, why replace?

eranhirsch:
It does more, debounce is a use-case of this. It's like compact vs filter(isTruthy)

cjquines:
two thoughts:

1. when i've used something like `batcher` in the past, i've wanted to split up the return values back to the individual calls, like what e.g. [batshit does](https://www.npmjs.com/package/@yornaath/batshit). with this we only get one big value for the whole batch

2. i think it's conceptually different enough, even if `debounce` is a special case. the name `batcher` suggests we do something on the whole batch of calls in between invokes, but with `debounce` and `throttle`, we're only doing something using the very last call.

    with `compact`, you can read `filter(isTruthy)` and it's clear that these are the same; but `debounce` and `batcher({ prepare: (prev, ...params) => params })` read quite different to me. (compare the other ones we've removed: `countBy` with `filter.length` or `flatMapToObj` with `fromEntries(flatMap)`, which are simpler compositions.)

eranhirsch:
Your concerns are on point. The PR isn't done yet and even the name is still not final. 

The aim is to create a doc with recipes on top of this with all the combos you are thinking about.

This is utility is very different from others we offer so I'm still considering how best to handle it. The last thing i want is to maintain several runtime "utility that has timers", i want one and have it be the definitive one. If we feel that a recipe doc is not enough we should consider helper utils on top of it, but even then, that would the first in the library and thus a slippery slope...

cjquines:
decent number of our problems would be solved if we had a `remeda-contrib` 😅

cjquines:
hey, probably won't get to reviewing this weekend, but wanted to raise a question real quick—i know we can derive `debounce` or `throttle` from `funnel`, but do we really want to not include it because it is derivable? the number of settings you have to pass in are quite big

if we don't include `debounce` or `throttle`:

* have to make decisions about which implementation of `debounce` and `throttle` to use
    * if we envision remeda to eventually replace lodash (which is part of my personal vision) then we might as well pick the lodash impls
* we can get rid of some of the text in the migrations pages
* much simpler for library consumers
    * by far the most common use case won't need batching
    * also, do i really want to think about which flavor of `debounce` i care about

(i know these arguments also apply for some other functions, but _particularly_ the complexity of `debounce` and `throttle` make these tradeoffs apparent, especially since it's not just a one-line replacement)

non-arguments (imo):

* would increase the number of functions we export, which means increased surface area / harder to learn stuff
    * we already export a lot of functions anyway. plus, `debounce` and `throttle` are much easier to grok than `funnel`
    * it also wouldn't increase the bundle size compared to consumers copying our implementation, because of strong tree-shaking
* no separate function = less explicit tests needed
    * i don't think we actually need to test derived functions so thoroughly. like, we don't need to have type tests for them i think? because they'd be redundant
    * and for `debounce` and `throttle` we're already including a bunch of tests anyway

eranhirsch:
Concretely, the only difference between funnel and debounce is **the reducer callback**, as the other params are also part of debounce:

```ts
_.debounce(callback, burstCoolDownMs);
_.throttle(callback, delayMs);

R.funnel(reducer, callback, { burstCoolDownMs });
R.funnel(reducer, callback, { delayMs });
```

I agree this is a real issue; there's no easy way to "plug" the reducer without knowing precisely what you are doing. For the two most common use cases, the resulting code isn't as trivial as I would want it to be:

```ts
// 1. No args.
R.funnel(constant(anything except `undefined`), ...);

// 2. Last args (lodash-style)
R.funnel((_, ...args) => args, ...); 

// 3. First args
R.funnel((prev, ...args) => prev ?? args, ...);

// 4. All args
R.funnel((prev, ...args) => [...(prev ?? []), args], ...);
``` 

The first case is, at the same time, the most common and least obvious to do in Remeda; we can consider allowing an `undefined` reducer to fall back to it. For the rest, I couldn't come up with a way where a remeda utility makes it better (e.g., `identity`, `concat`, etc...). I'll be happy to get ideas for other solutions, too.

---

## But more broadly speaking...

I think Lodash's implementation is kinda... bad. That brings me to something I've already said in the past, which is that we don't aim to be a drop-in replacement for Lodash (e.g., `isEmpty(undefined)`) but want to make migrating as easy as possible (so missing functionality is something we care about, but not different APIs).

People have this wrong assumption about Lodash being this very high-quality, well designed and planned, highly optimized for perf, and well-tested library, but that's just not what Lodash is. Lodash is just an aggregation of loose ends collected from people's "util" libraries over time, mostly around 2012-2016, and at a time where EMCAScript was just not a good standard to rely on, so people had to make guesses on how APIs should behave.

Taking debounce as an example, Lodash put a lot of code to support managing the "cachedValue" spec, but very few examples in the wild ever use that value, and its implementation assumes the lack of modern async/promise constructs that just work differently. Additionally, Lodash makes assumptions about what argument you want to use with your callback, which leaves anyone with a different requirement stuck without an implementation; similarly, examples and blog posts rarely use the argument, so they hide the problems.

So what should we offer? A debounce with a last-arg spec? A debounce with a cached-value spec? Both? None? Aren't we doing people a disservice by bundling logic they don't need just because someone might have asked for it 8 years ago?

The more I worked on this PR, the more I liked your idea of an adjacent library (remedax? rex?) or even just an adjacent import path, which **does** offer quick composite solutions on top of core utility functions but with a ton of caveats about the amount of support we will offer, how thorough it is tested, and any performance implications by possibly doing redundant stuff or skipping our lazy evaluation mechanism. This would allow us to offer the whole range of mapped utility functions we currently reject to adding: `reject`, `compact`, `pluck`, etc... 

Another approach is to provide an adapter layer for lodash/ramda that would allow you to replace imports from lodash with imports from a "remeda/lodash" path which will provide adapters that makes it possible to migrate to Remeda without any changes to your code (with the caveat that this doesn't provide easy utilities for people who are **not** coming from Lodash).
```ts
import _ from "remeda/lodash";
```

Two final approaches are to provide snippets: either like shadcn, where we have a section of the docs site for them; or as test files, where they are coupled with the code and could show usage examples via test suites.

The latter approach was used in this PR as it required the fewest changes to our docs infra. If this approach wins, we can consider creating a different folder for it (so we can remove it from coverage reports), and we should definitely add more reference implementations!

eranhirsch:
Another way to view this is that it makes sense for Remeda to be a Lodash replacement but not a Lodash **v4** replacement; that is, it's OK for us to make breaking changes to the Lodash APIs as long as people have a mostly trivial way to migrate from one "version" to the next (in other words Remeda should be seen as Lodash v5, not a replacement for Lodash v4).

eranhirsch:
I redesigned the solution so that the reducer is optional and part of the options object, making more straightforward cases with no args easier to configure. This doesn't yet solve the issue of how to make the reducer functions easier to read and write:

```ts
const debounced = _.debounce(callback, burstCoolDownMs);
debounced();

const debounced = R.funnel(callback, { burstCoolDownMs });
debounced.call();
```

```ts
const throttled = _.throttle(callback, delayMs);
throttled();

const throttled = R.funnel(callback, { delayMs });
throttled.call();
```

src/funnel.ts
```diff
@@ -0,0 +1,280 @@
...
+ * the optional `reducer` function accepts. Use **explicit** types for these
+ * to ensure that everything _else_ is well-typed.
+ *
+ * Notice that this function constructs a funnel **object**, and does **not**
+ * execute anything when called. The returned object should be used to execute
+ * the funnel via the its `call` method.
+ *
+ * Debouncing: use `burstCoolDownMs` and any `invokedAt`.
+ * Throttling: use `delayMs` and `invokedAt: "start"` or `"both"`.
+ * Batching: See the reference implementation in [`funnel.reference-batch.test.ts`](https://github.com/remeda/remeda/blob/main/src/funnel.reference-batch.test.ts).
```
cjquines, line 63: **nit:** let's format this as a list maybe

src/funnel.ts
```diff
@@ -0,0 +1,280 @@
...
+ * **not** be enabled, 0 is not used a default fallback.
+ * @param options.reducer - Combines the arguments passed to `call` with the
+ * value computed on the previous call (or `undefined` on the first time). The
+ * goal of the function is to extract and summarize the data needed for
+ * `callback`. It should be fast and simple as it is called often and should
+ * defer heavy operations to the `execute` function. If the final value
+ * is `undefined`, `callback` will not be called.
+ * @param options.invokedAt - At what "edges" of the funnel's activity
+ * window should `execute` be invoked. `start` means The function will be
+ * invoked immediately (withing the **same** execution frame!), and any
```
cjquines, line 81: **typo:** withing -> within

src/funnel.ts
```diff
@@ -0,0 +1,280 @@
...
+ * invoked, relative to the calls of `call`. A timer that isn't defined will
+ * **not** be enabled, 0 is not used a default fallback.
+ * @param options.reducer - Combines the arguments passed to `call` with the
+ * value computed on the previous call (or `undefined` on the first time). The
+ * goal of the function is to extract and summarize the data needed for
+ * `callback`. It should be fast and simple as it is called often and should
+ * defer heavy operations to the `execute` function. If the final value
+ * is `undefined`, `callback` will not be called.
+ * @param options.invokedAt - At what "edges" of the funnel's activity
+ * window should `execute` be invoked. `start` means The function will be
```
cjquines, line 80: **nit:** can this be formatted as a list, with `start`, and `end` etc.?

cjquines: we should really improve how the docs site presents params...

src/funnel.ts
```diff
@@ -0,0 +1,280 @@
...
+ * @param options.burstCoolDownMs - The burst timer prevents subsequent
+ * calls in short succession to cause excessive invocations (aka "debounce").
+ * This duration represents the **minimum** amount of time that needs to pass
+ * between calls (the "quiet" part) in order for the subsequent call to **not**
+ * be considered part of the burst. In other words, as long as calls are faster
+ * than this, they are considered part of the burst.
+ * @param options.maxBurstDurationMs - Bursts are extended every time a
+ * call is made within the burst period. This means that the burst period could
+ * be extended indefinitely. To prevent such cases, a maximum burst duration
+ * could be defined.
```
cjquines, line 98: **note:** i want to bikeshed how to name this and describe this... at minimum, if we're sticking with `burstCooldownMs` and `maxBurstDurationMs`, we should actually give an explicit definition of a burst

src/funnel.ts
```diff
@@ -0,0 +1,280 @@
...
+ * @param options.delayMs - A minimum duration between calls of `execute`.
+ * This is maintained regardless of the shape of the burst and is ensured even
+ * if the `maxBurstDurationMs` is reached before it. (aka "throttle").
```
cjquines, line 101: **note:** i similarly want to bikeshed this name, i dont think `delayMs` is a great name

cjquines (commented):
initial comments

src/funnel.ts
eranhirsch: I'm not sure that would render properly in IDE tooltips, they usually consider this as a single line.

src/funnel.ts
eranhirsch: I don't mind considering other names.

I've said in the past that I don't want remeda to just be a shallow clone of Lodash, and we should consider improving in places where Lodash didn't do a good job; Lodash uses 'wait' and 'maxWait' and I felt that those don't represent the semantics well when we add a third time param (delay) to the mix. I didn't want to call these 'debounceTime' and 'throttleTime' because that requires someone to know what debounce and throttle is. I tried to find names that make the most sense when Remeda is the only library you know.

eranhirsch:
@cjquines, where are we on this, it's still the top missing feature for people coming from Lodash

cjquines:
i'll try to take a look later this week, during thanksgiving

docs/src/content/mapping/lodash/debounce.md
```diff
@@ -1,4 +1,137 @@
...
+
+- A more complete reference implementation that also maintains Lodash's
+  capability to store the callback's return value is available below, and in [`funnel.lodash-debounce-with-cached-value.test.ts`](https://github.com/remeda/remeda/blob/main/src/funnel.lodash-debounce-with-cached-value.test.ts).
+
+- These implementations can be copied as-is into your project, but might contain
+  redundant parts which are not relevant for your specific use cases. By
+  inlining only the parts you need you can take advantage of capabilities not
+  available in Lodash.
+
+### Reference
```
cjquines, line 18: **thought:** maybe we can collect these in the docs under a `remeda-contrib` thing...? hm

src/funnel.test.ts
```diff
@@ -0,0 +1,1049 @@
...
+    foo.call();
+
+    expect(mockFn).toHaveBeenCalledTimes(1);
+    expect(mockFn).toHaveBeenLastCalledWith("hello world");
+  });
+
+  it("reduces call args", async () => {
+    const mockFn = vi.fn((_total: number): void => {
+      /* do nothing */
+    });
```
cjquines, line 37: **nit:** i mildly prefer one of these two forms, what do you think?

```ts
    const mockFn = vi.fn() as (total: number) => void;
    const mockFn = vi.fn<(total: number) => void>();
```

src/funnel.test.ts
```diff
@@ -0,0 +1,1049 @@
...
+
+      expect(mockFn).toHaveBeenCalledTimes(2);
+
+      await sleep(UT);
+
+      expect(mockFn).toHaveBeenCalledTimes(3);
+      expect(mockFn).toHaveBeenLastCalledWith(["d", "e", "f"]);
+    });
+
+    it("never fires when only delay is used and invoked only at end", async () => {
```
cjquines, line 171: **question:** is this behavior desirable?

i know it's technically to-spec, but i think it's counterintuitive. i think we should disallow this at the type-level

src/funnel.ts
cjquines: actually, i'm not sure why we even want `delayMs`. i feel like `maxBurstDurationMs` is enough, and i cannot imagine a case where these would be different

src/funnel.ts
```diff
@@ -0,0 +1,280 @@
...
+ * reduced result would trigger an invocation. When `both` is used The function
+ * will be invoked immediately, and then the funnel would behave as if it was
+ * in the 'end' state. @default 'end'.
+ * @param options.burstCoolDownMs - The burst timer prevents subsequent
+ * calls in short succession to cause excessive invocations (aka "debounce").
+ * This duration represents the **minimum** amount of time that needs to pass
+ * between calls (the "quiet" part) in order for the subsequent call to **not**
+ * be considered part of the burst. In other words, as long as calls are faster
+ * than this, they are considered part of the burst.
+ * @param options.maxBurstDurationMs - Bursts are extended every time a
```
cjquines, line 95: **question:** instead of using `maxBurstDurationMs`, what do you think about something like a `maxCallsPerSecond` or `maxCallsPerMinute`?

the thing is, i'm not sure whether thinking in terms of "minimum time between executions" is better than "maximum number of executions per given time period". for me it depends on the use case:

* when i'm, say, typing on a keyboard and fetching with every keystroke, i think more like "i want to call this API a maximum of 2 times per second"
* when i'm doing animation, i think more like "i want to update the screen once every 30ms"

src/funnel.test.ts
```diff
@@ -0,0 +1,1049 @@
...
+      await sleep(UT);
+
+      expect(mockFn).toHaveBeenCalledTimes(1);
+
+      foo.call("d");
+
+      expect(mockFn).toHaveBeenCalledTimes(2);
+      // "b" and "c" have been ignored because the funnel is only configured to
+      // run the function at the start of the delay window, meaning anything
+      // after that is simply dropped.
```
cjquines, line 248: **question:** do you think this behavior is desirable?

i think it is... but i'm not sure.

src/funnel.test.ts
```diff
@@ -0,0 +1,1049 @@
...
+    expect(mockFn).toHaveBeenLastCalledWith(["a"]);
+
+    await sleep(0);
+
+    expect(mockFn).toHaveBeenCalledTimes(2);
+    expect(mockFn).toHaveBeenLastCalledWith(["b"]);
+  });
+});
+
+describe("disabled timers (no defined timeout durations)", () => {
```
cjquines, line 771: **question:** do you think we should support these cases?

src/funnel.lodash-debounce-with-cached-value.test.ts
```diff
@@ -0,0 +1,252 @@
...
+      // Funnel provides more control over the args, but lodash simply passes
+      // them through, to replicate this behavior we need to spread the args
+      // array maintained via the reducer below.
+      // Also, every time the function is invoked the cached value is updated.
+      cachedValue = func(...args) as ReturnType<F>;
+    },
+    {
+      // Debounce stores the latest args it was called with for the next
+      // invocation of the callback.
+      reducer: (_, ...args: Parameters<F>) => args,
```
cjquines, line 73: **question:** does this work when the callback is called with `undefined`?

src/funnel.ts
```diff
@@ -0,0 +1,280 @@
...
+ * via closure to the outer scope of the funnel.
+ * @param options - An object that defines when `execute` should be
+ * invoked, relative to the calls of `call`. A timer that isn't defined will
+ * **not** be enabled, 0 is not used a default fallback.
+ * @param options.reducer - Combines the arguments passed to `call` with the
+ * value computed on the previous call (or `undefined` on the first time). The
+ * goal of the function is to extract and summarize the data needed for
+ * `callback`. It should be fast and simple as it is called often and should
+ * defer heavy operations to the `execute` function. If the final value
+ * is `undefined`, `callback` will not be called.
```
cjquines, line 78: **thought:** i can think of four common reducer cases:

1. no reducer — when the callback takes no params
2. use the first value — this should be, in theory, `identity()`, but that's not true because of the `undefined` the first time
3. use the last value — this is `(_, ...args) => args`, that's fine
4. array of all values — what the test calls `ARGS_COLLECTOR`.

we have a good solution for 1 now, maybe not the rest

cjquines: i guess the fact that we have an example for 4 is fine

cjquines: maybe we should have a wiki page for `funnel`

src/funnel.test.ts
eranhirsch: The form I used works when you also want to provide an actual implementation and are using the mock as a wrapper to track things like called arguments or the number of calls. e.g. 

```ts
let accumulator = 0;
const mockFn = vi.fn((param: number) => accumulator + param);

...

expect(accumulator).toBe(3);
expect(mockFn).toHaveBeenCalledWith([3]);
```

That can't be solved with either of the other forms, so it would create two different syntaxes for using vi.fn, depending on whether you need an actual implementation. Do you think that's still better?

src/funnel.test.ts
eranhirsch: Yea, that's a good idea to prevent this at the type level. My only concern is that the typing might constraint certain cases where people might build the timing policy object dynamically, e.g.:

```ts
const options = {};

if (...) {
  options.invokedAt = "end";
} else {
  options.invokedAt = "start";
}

if (...) {
  options.delayMs = 500;
} else {
  options.burstCoolDownMs = 500;
}

const x = funnel(..., options);
```

Albeit this is a contrived example, and perhaps preventing accidental bad edge-cases is more valuable.
```

File: src/funnel.ts (hunk 9 above)
Comment (eranhirsch) [reply src/funnel.ts:95]: That's a great idea also to have those, but they are separate concerns. Debouncing (burst) and Throttling (delay) are timing-based execution limitation mechanisms. 

This is one of the reasons I had a hard time designing this utility. At the end of the day, it makes sense to have a standalone library that provides a whole set of execution-shaping functions so that you can match it to whatever spec you have. If Lodash didn't have debounce/throttle, I would have said this function is out of scope, but as it stands right now these two functions are a blocker for migrating from Lodash for a lot of people.

File: src/funnel.test.ts (hunk 10 above)
Comment (eranhirsch) [reply src/funnel.test.ts:248]: There is no other logical way to handle them. If execution is set to "start" it means there is no later timing they could be used at.

File: src/funnel.test.ts (hunk 11 above)
Comment (eranhirsch) [reply src/funnel.test.ts:771]: We can default the burst timer to 0 when nothing is provided; it's probably the most sensible option.

File: src/funnel.lodash-debounce-with-cached-value.test.ts (hunk 12 above)
Comment (eranhirsch) [reply src/funnel.lodash-debounce-with-cached-value.test.ts:73]: args is an array, if a function is *called* with an arg of `undefined` this would be `[undefined]` which is a non-trivial value and should work.

File: src/funnel.ts (hunk 13 above)
Comment (eranhirsch) [reply src/funnel.ts:78]: Following a similar discussion we had around groupBy/countBy (e.g. "reduceBy") we can consider exporting a folder of "reduce utils" that would provide functions like below, 

```ts
const grouper = <T>(accum: readonly T[] | undefined, item: T) => [...(accum ?? []), item];
// e.g. groupBy = reduceBy(grouper);

const first = <T>(accum: T | undefined, item: T) => accum ?? item;

const last = <T>(accum: T | undefined, item: T) => item;

const counter = (accum: number | undefined, item: unknown) => (accum ?? 0) + 1
// e.g. countBy = reduceBy(counter);

const max = (accum: number | undefined, item: number) => Math.max(accum ?? -Infinity, item);

const min = (accum: number | undefined, item: number) => Math.min(accum ?? Infinity, item); 
```

Speaking more broadly, though, a reducer provides users the flexibility to use our functions as a well-tested base, which on top, they can build very custom logic, imagine for example, that you are debouncing mouse-movement events (which fire a lot) to log how far the user moved their mouse after clicking a button. Instead of adding all these event objects into an array so that you can process them when the function is invoked (to extract the max distance AND send the result to the server), you can build a custom reducer that, on every call to the funnel, extracts the distance from a single event, and check if that distance is the max. This means that instead of wasting all that time and memory building a huge array of events, you only maintain the information you need for the actual invocation.

I think that designing the API like this allows us to make both scenarios (the common "I just need the first/last item" and the rare "I need to keep this specific datum") easy to write and read.

eranhirsch:
Your biggest concerns seem to be around the following:

1. Naming, which I agree with and would love to reach an agreement on. I'm OK with using debounceMs and throttleMs as those would be very easy to grok for people coming from Lodash, but fear that people not familiar with Lodash would find them non-descriptive enough to understand.

2. The fact that debounce and throttle live within the same function; causing a mish-mash of terms and flexibilities that could be split into more concise utilities. I understand why the current API might be confusing, but I assumed that people would only use the specific parts of the options object that answer their need, just like react-query provides a huge Options object to useQuery, but people aren't expected to understand all of it to use it. One of the things I liked about having both timers is that it allows you to throttle and debounce at the same time! something that Lodash doesn't provide (you can only "throttle" by using maxWait === wait). e.g., in our implementation, you can build a funnel that:
a. Is never invoked more than once per second (delay)
b. Is prevented from being called too frequently (100ms burst)
c. Is at least called once every 5 seconds, even if during a burst (max burst duration).

src/funnel.test.ts
cjquines: i think it's a matter of preference. i mildly prefer using one form for empty functions and one form for functions with an implementation, but honestly this is a nit

src/funnel.test.ts
cjquines: i can't really think of a case where the timing object is built dynamically. i would suggest we prevent this now, and if we decide to change this in the future we can always make the typing less strict, which is easier than the reverse.

src/funnel.ts
cjquines: yeah, i see the point now. i really don't think a function like this should belong in remeda, but i see the practicality to including it.

src/funnel.test.ts
cjquines: i think that makes sense, we can also forbid it at the type-level?

docs/src/content/mapping/lodash/debounce.md
cjquines: there's libraries called "lodash-contrib" and "underscore-contrib".

i actually don't like my idea anymore

cjquines:
with regards to naming, i'm currently thinking something like

```ts
function funnel(callback: Function, options: {
  reducer: Function;
  callAt: "end" | "start" | "both";
  idleThresholdMs: number;
  groupLimitMs: number;
  callIntervalMs: number;
});
```

rationale:

* `callAt`: while we use the word "invoke" a lot, i don't think we should use it in the param... we use "call" more. the js name is `Function.prototype.call` too

* `idleThresholdMs`: the main use of debounce, at least in the frontend, is to detect when the user is idle. we debounce calls to window resizing and typing, because we want to wait for the user to idle before doing anything; compare this to scroll handlers or animations, where we don't want the user to idle before we do things. i think the "idle" framing is better than the "burst" framing in this case. i don't know how i feel about "threshold", but it feels like the best word

* `groupLimitMs`: this is the one i feel least sure about. i think the advantage of "group" over "burst" is when you write docs. "limit" is short and feels better than e.g. "timeout" or "maxDuration"

* `callIntervalMs`: "interval" is specifically chosen to match `setInterval`

for docs, imagine something like:

> Creates a funnel that groups consecutive calls to `callback` according to a timing policy.
>
> When the funnel is first called, it starts a new group. Subsequent calls are part of this group. The group ends when:
> * `callIntervalMs` has passed since the last `callback` call, _and_
>   * `idleThresholdMs` has passed since the last funnel call, _or_
>   * `groupLimitMs` has passed since the start of the group.
>
> The `callback` can be called at the start of each group, at the end of each group, or both. A `reducer` can combine arguments before passing them to `callback`.

plus some of the explanation that's currently there

eranhirsch:
re *call* vs *invoke*: my thinking is that call refers to the user calling the `.call(...)` function of the funnel; this is where the business logic is **actively** interacting with the funnel. `invoke` on the other hand is when the funnel is triggered, this is an internal **passive** event that the business logic isn't aware of.

cjquines:
hm. right now i don't think the words "call" and "invoke" give that meaning, they're kinda the same word. i'm currently leaning toward "funnel call" and "callback call"

eranhirsch:
Hey @cjquines , I did another take on the things that you surfaced last time:
1. Changed timer param names (minQuietPeriodMs, maxGapMs, and minGapMs), yet the debouncing timer is still referred to as a "burst" in the docs.
2. Prevent certain combinations of options at the type level.
3. Default minQuietPeriodMs to 0 when no timers are configured.

Let's try and get this thing shipped. We can discuss changes in the future in a major version bump (which we can plan for this june 2025 once Node 18 reaches EOL).

eranhirsch:
OK... I still think I can make it even clearer...

eranhirsch:
While working on the previous revision I noticed that it's wasn't intuitive to define the batcher with a non 0 time window. I made 2 changes: 1. rename the option so that it reflects the whole burst window. 2. allow calling the funnel with *just* the `maxBurstDurationMs` param and default the `minQuietPeriodMs` param to it when it isn't defined too (similar to what lodash does).

This now allows many use-cases to be defined with simple params:
```ts
const debounced = funnel(func, { minQuietPeriodMs: 100 });
const throttled = funnel(func, { minGapMs: 100, triggerAt: "both" });
const batched = funnel(func, { reducer: (acc, params) => [...(acc ?? []), params] });
const timeBatched = funnel(func, { reducer: ..., maxBurstDurationMs: 100 });
```

cjquines (approved):
i still somewhat dislike the documentation and param names (i still think we should not be referring to "bursts" at all), but that is easier to change than the other comments i had. approving

PR #852, opened 2024-09-13
feat(sumBy): add bigint support for `sumBy`

homura:
This PR adds `bigint` support for the `sumBy` function

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/sumBy.test.ts
+++ b/src/sumBy.test.ts
@@ -17,6 +17,16 @@ describe("data first", () => {
       ),
     ).toEqual(25);
   });
+
+  test("works with bigint", () => {
+    expect(
+      sumBy([{ a: 1n }, { a: 2n }, { a: 4n }, { a: 5n }, { a: 3n }], prop("a")),
+    ).toEqual(15n);
+  });
+
+  it("should return 0 for an empty array", () => {
+    expect(sumBy([] as Array<{ a: bigint }>, prop("a"))).toBe(0);
+  });
 });
 
 describe("data last", () => {
@@ -27,6 +37,19 @@ describe("data last", () => {
         sumBy(prop("a")),
       ),
     ).toEqual(15);
+  });
+
+  test("works with bigint", () => {
+    expect(
+      pipe(
+        [{ a: 1n }, { a: 2n }, { a: 4n }, { a: 5n }, { a: 3n }],
+        sumBy(prop("a")),
+      ),
+    ).toEqual(15n);
+  });
+
+  it("should return 0 for an empty array", () => {
+    expect(pipe([] as Array<{ a: bigint }>, sumBy(prop("a")))).toBe(0);
   });
 
   test("indexed", () => {
--- a/src/sumBy.ts
+++ b/src/sumBy.ts
@@ -11,6 +11,13 @@ import { purry } from "./purry";
  *      [{a: 5}, {a: 1}, {a: 3}],
  *      R.sumBy(x => x.a)
  *    ) // 9
+ *
+ *    R.pipe(
+ *      [{a: 5}, {a: 1}, {a: 3}],
+ *      R.sumBy(x => x.a)
+ *    ) // 9n
+ *
+ *    R.pipe([] as {a: bigint}[], R.sumBy(x => x.a)) // 0
  * @dataLast
  * @category Array
  */
@@ -18,6 +25,10 @@ import { purry } from "./purry";
 export function sumBy<T>(
   callbackfn: (value: T, index: number, data: ReadonlyArray<T>) => number,
 ): (items: ReadonlyArray<T>) => number;
+
+export function sumBy<T>(
+  callbackfn: (value: T, index: number, data: ReadonlyArray<T>) => bigint,
+): (items: ReadonlyArray<T>) => bigint | 0;
 
 /**
  * Returns the sum of the elements of an array using the provided predicate.
@@ -31,6 +42,13 @@ export function sumBy<T>(
  *      [{a: 5}, {a: 1}, {a: 3}],
  *      x => x.a
  *    ) // 9
+ *
+ *    R.sumBy(
+ *      [{a: 5n}, {a: 1n}, {a: 3n}],
+ *      x => x.a
+ *    ) // 9n
+ *
+ *   R.sumBy([] as {a: bigint}[], x => x.a)
  * @dataFirst
  * @category Array
  */
@@ -40,18 +58,33 @@ export function sumBy<T>(
   callbackfn: (value: T, index: number, data: ReadonlyArray<T>) => number,
 ): number;
 
+export function sumBy<T>(
+  data: ReadonlyArray<T>,
+  callbackfn: (value: T, index: number, data: ReadonlyArray<T>) => bigint,
+): bigint | 0;
+
 export function sumBy(...args: ReadonlyArray<unknown>): unknown {
   return purry(sumByImplementation, args);
 }
 
 const sumByImplementation = <T>(
   array: ReadonlyArray<T>,
-  callbackfn: (value: T, index: number, data: ReadonlyArray<T>) => number,
-): number => {
-  let sum = 0;
+  callbackfn: (
+    value: T,
+    index: number,
+    data: ReadonlyArray<T>,
+  ) => bigint | number,
+): bigint | number => {
+  let sum: bigint | number | undefined;
   for (const [index, item] of array.entries()) {
     const summand = callbackfn(item, index, array);
-    sum += summand;
+    if (sum === undefined) {
+      sum = summand;
+      continue;
+    }
+    // we use an `as number` here to let TypeScript
+    // treats `sum` and `summand` as the same type
+    (sum as number) += summand as number;
   }
-  return sum;
+  return sum ?? 0;
 };


homura:
Hi, @cjquines and @eranhirsch, could you please review or give me feedback on this PR? It simplifies the `sum` scenario when working with `bigint`

```ts
R.sumBy(someFn) == R.piped(R.map(someFn), R.sum)
```

src/sumBy.ts
```diff
@@ -40,18 +58,34 @@ export function sumBy<T>(
...
     const summand = callbackfn(item, index, array);
-    sum += summand;
+    if (sum === undefined) {
+      sum = summand;
+      continue;
+    }
+    // we use 'as number' to ensure TypeScript treats
+    // 'sum' and 'summand' as the same type,
+    // which could be a bigint.
+    (sum as number) += summand as number;
```
cjquines, line 88: **suggestion:** let's use an explicit `ts-expect-error` rather than a typecast; see the similar thing in `sum`

src/sumBy.ts
```diff
@@ -19,6 +26,10 @@ export function sumBy<T>(
   callbackfn: (value: T, index: number, data: ReadonlyArray<T>) => number,
 ): (items: ReadonlyArray<T>) => number;
 
+export function sumBy<T>(
+  callbackfn: (value: T, index: number, data: ReadonlyArray<T>) => bigint,
+): (items: ReadonlyArray<T>) => bigint | 0;
```
cjquines, line 31: **suggestion:** let's have a stricter type here; if the array is non-empty we know that the output type is `bigint`, not just `bigint | 0`

cjquines (changes requested):

src/sumBy.ts
```diff
@@ -40,18 +58,34 @@ export function sumBy<T>(
...
+    value: T,
+    index: number,
+    data: ReadonlyArray<T>,
+  ) => bigint | number,
+): bigint | number => {
+  let sum: bigint | number | undefined;
   for (const [index, item] of array.entries()) {
     const summand = callbackfn(item, index, array);
-    sum += summand;
+    if (sum === undefined) {
```
eranhirsch, line 81: This adds a check on every iteration, but we know it's only relevant on the first iteration. We should run the mapper on the first item before the loop and use it to initialize `sum`.

src/sumBy.ts
homura: thx for your suggestion. I have made the changes, so please review it again

src/sumBy.ts
homura: thx for your suggestion. I have added the type tests

src/sumBy.ts
homura: thx for the suggestion. I've made a change to avoid checking for the undefined in the loop

src/sumBy.ts
```diff
@@ -11,6 +18,13 @@ import { purry } from "./purry";
  *      [{a: 5}, {a: 1}, {a: 3}],
  *      R.sumBy(x => x.a)
  *    ) // 9
+ *
+ *    R.pipe(
+ *      [{a: 5n}, {a: 1n}, {a: 3n}],
+ *      R.sumBy(x => x.a)
+ *    ) // 9n
+ *
+ *    R.pipe([] as {a: bigint}[], R.sumBy(x => x.a)) // 0
```
eranhirsch, line 27: This concept is much easier to understand when mentioned in the function's description instead of as an example. The example doesn't make it clear what it's trying to say. Check out how we did it in `sum`.

src/sumBy.ts
```diff
@@ -1,4 +1,11 @@
 import { purry } from "./purry";
+import { type IterableContainer } from "./internal/types";
+
+type SumByBigInt<T, TS extends IterableContainer<T>> = TS extends []
+  ? 0
```
eranhirsch, line 5: This will not catch readonly arrays (e.g. `[] as const`)
```suggestion
type SumByBigInt<T, TS extends IterableContainer<T>> = TS extends readonly []
  ? 0
```

src/sumBy.ts
```diff
@@ -1,4 +1,11 @@
...
+  : TS extends readonly [T, ...ReadonlyArray<T>]
```
eranhirsch, line 6: We only care about the shape of the array, not it's content, the second generic is redundant here.

```suggestion
type SumByBigInt<T extends IterableContainer> = T extends []
  ? 0
  : T extends readonly [unknown, ...ReadonlyArray<unknown>]
```

src/sumBy.test.ts
```diff
@@ -17,6 +17,16 @@ describe("data first", () => {
...
+
+  test("works with bigint", () => {
+    expect(
+      sumBy([{ a: 1n }, { a: 2n }, { a: 4n }, { a: 5n }, { a: 3n }], prop("a")),
+    ).toEqual(15n);
+  });
+
+  it("should return 0 for an empty array", () => {
+    expect(sumBy([] as Array<{ a: bigint }>, prop("a"))).toBe(0);
+  });
```
eranhirsch, line 29: This is not a runtime test; it shouldn't be in this file.
```suggestion
```

src/sumBy.test.ts
```diff
@@ -29,6 +39,19 @@ describe("data last", () => {
...
+      pipe(
+        [{ a: 1n }, { a: 2n }, { a: 4n }, { a: 5n }, { a: 3n }],
+        sumBy(prop("a")),
+      ),
+    ).toEqual(15n);
+  });
+
+  it("should return 0 for an empty array", () => {
+    expect(pipe([] as Array<{ a: bigint }>, sumBy(prop("a")))).toBe(0);
+  });
```
eranhirsch, line 53: Similarly
```suggestion
```

src/sumBy.test-d.ts
```diff
@@ -0,0 +1,85 @@
...
+  });
+});
+
+function toNumber(_value: unknown): number {
+  return 1;
+}
+
+function toBigInt(_value: unknown): bigint {
+  return 1n;
+}
```
eranhirsch, line 85: We have `constant` for these kinds of things, you don't need to define your own function.
```suggestion
```

src/sumBy.test-d.ts
```diff
@@ -0,0 +1,85 @@
+import { sumBy } from "./sumBy";
+import { expectTypeOf } from "vitest";
+import { pipe } from "./pipe";
+
+test("empty array", () => {
+  const result1 = sumBy([], BigInt);
+  expectTypeOf(result1).toEqualTypeOf<0>();
+
+  const result2 = sumBy([], toNumber);
+  expectTypeOf(result2).toEqualTypeOf<number>();
```
eranhirsch, line 10: To be consistent, the typing of `sumBy` for `number` should also be fixed. This should also return a constant `0`, just like for `bigint`.

src/sumBy.test-d.ts
```diff
@@ -0,0 +1,85 @@
...
+    expectTypeOf(result).toEqualTypeOf<bigint>();
+  });
+
+  test("fixed-size tuples", () => {
+    const result = sumBy([1n, 2n] as [unknown, unknown], toBigInt);
+    expectTypeOf(result).toEqualTypeOf<bigint>();
+  });
+});
+
+describe("dataLast", () => {
```
eranhirsch, line 67: Add tests for the empty case for data-last to test that the types are inferred correctly "through" the pipe.

eranhirsch (changes requested):

src/sumBy.ts
```diff
@@ -40,17 +68,38 @@ export function sumBy<T>(
   callbackfn: (value: T, index: number, data: ReadonlyArray<T>) => number,
 ): number;
 
+export function sumBy<
+  T,
+  TS extends IterableContainer<T> = IterableContainer<T>,
+>(
+  data: TS,
+  callbackfn: (value: T, index: number, data: TS) => bigint,
+): SumByBigInt<T, TS>;
```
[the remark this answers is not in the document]
eranhirsch: When considering changes, we don't consider the generics as part of the function signature. This PR is already marked as a `feat` (and not a `fix`), so it's OK to make slight changes to the typing. Notice that we never write tests that have explicit values for generics for this reason. We assume that generics are inferred.

cjquines (commented):
after eran's suggested changes, i'm fine to ship

src/sumBy.test.ts
homura: the `sumBy` test lacks the empty array case. So I removed the type and kept the test to ensure it returns 0 when the parameter is an empty array

src/sumBy.ts
```diff
@@ -40,17 +68,38 @@ export function sumBy<T>(
...
+): bigint | number => {
+  if (array.length === 0) {
+    return 0;
+  }
+
+  const [head, ...tail] = array;
+  let sum = callbackfn(head!, 0, array);
+
+  for (const [index, item] of tail.entries()) {
+    const summand = callbackfn(item, index + 1, array);
```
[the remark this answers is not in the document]
homura: Refactored using the iterator API

src/sumBy.ts
homura: Moved to the description for easier to understand

src/sumBy.ts
homura: Done

src/sumBy.ts
homura: Done

src/sumBy.test-d.ts
homura: The result of the function `sumBy` has been refactored for stricter, please check it again

src/sumBy.test-d.ts
homura: Two empty cases were added for stricter

src/sumBy.test-d.ts
homura: Moved to the `constant`

src/sumBy.ts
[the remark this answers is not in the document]
eranhirsch: From my experimentation, this typing should work as expected; the `=== true` syntax should be avoided if possible.

https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABMOcA8AVAfACgCYCGUBAXIhgNoC6AlGQG5wx6IDeAUIohAgM5SIYUAKYAnRAF5EhYgDphYKKJjDeOGgG5OiAPQ6uXAHoB+dtp5h+yGKP4BRRaICekwSNGywwgB5R1Wrj0DRBMzLhhgRBxgG3tHJ1k8BGEaNm0uUWEoEFEwAMQAXzDuPgEOYPoCABsQYTIKABprWygANWraqgbtAtcYloclJ3yLXjgq4VkquABzaNi2jpT8oOC10KKgA

src/sumBy.ts
```diff
@@ -1,8 +1,26 @@
...
+    : U | 0;
 
 /**
  * Returns the sum of the elements of an array using the provided predicate.
  *
+ * Works for both `number` and `bigint` predicates, but not predicates that contain both
+ * types.
+ *
+ * IMPORTANT: The result for empty arrays would be 0 (`number`) regardless of
+ * the type of the predicate; to avoid adding this to the return type for cases
```
eranhirsch, line 20:
```suggestion
 * the type of the mapper; to avoid adding this to the return type for cases
```

eranhirsch (approved):
Looks good and thanks for doing this! I'll merge this once you resolve the remaining remarks.

src/sumBy.ts
```diff
@@ -1,8 +1,26 @@
...
+> = T extends readonly []
+  ? 0
+  : T extends readonly [unknown, ...ReadonlyArray<unknown>]
+    ? U
+    : U | 0;
 
 /**
  * Returns the sum of the elements of an array using the provided predicate.
  *
+ * Works for both `number` and `bigint` predicates, but not predicates that contain both
```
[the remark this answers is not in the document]
homura: I use the word `predicate` copied from the following line

https://github.com/remeda/remeda/blob/9b6f4b90dd69fa023a6830e07ea53b68d0ad26ee/src/sumBy.ts#L14

And I changed it BTW, please review again

src/sumBy.test-d.ts
```diff
@@ -0,0 +1,101 @@
...
+test("empty array", () => {
+  const result1 = sumBy([], BigInt);
+  expectTypeOf(result1).toEqualTypeOf<0>();
+
+  const result2 = sumBy([], toNumber);
+  expectTypeOf(result2).toEqualTypeOf<0>();
+});
+
+test("disallow mixed predicate", () => {
+  const toNumberOrBigint = constant<bigint | number>(1);
```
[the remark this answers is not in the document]
homura: refactored

src/sumBy.test-d.ts
```diff
@@ -0,0 +1,101 @@
+import { sumBy } from "./sumBy";
+import { expectTypeOf } from "vitest";
+import { pipe } from "./pipe";
+import { constant } from "./constant";
+
+const toNumber = constant(1);
+const toBigInt = constant(1n);
```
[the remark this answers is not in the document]
homura: Refactored as inline

src/sumBy.test-d.ts
```diff
@@ -0,0 +1,101 @@
+import { sumBy } from "./sumBy";
+import { expectTypeOf } from "vitest";
```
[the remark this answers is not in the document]
homura: The unnecessary import is removed from the type test

src/sumBy.ts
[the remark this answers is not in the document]
cjquines: oh that should be mapper not predicate. thanks for the change

cjquines (approved):

src/sumBy.ts
```diff
@@ -11,13 +11,13 @@ type SumBy<
     : U | 0;
 
 /**
- * Returns the sum of the elements of an array using the provided predicate.
+ * Returns the sum of the elements of an array using the provided mapper.
  *
- * Works for both `number` and `bigint` predicates, but not predicates that contain both
+ * Works for both `number` and `bigint` mappers, but not mappers that contain both
```
eranhirsch, line 16:
```suggestion
 * Works for both `number` and `bigint` mappers, but not mappers that return both
```

src/sumBy.ts
```diff
@@ -11,17 +29,32 @@ import { purry } from "./purry";
...
+  callbackfn: (value: T[number], index: number, data: T) => number,
+): (items: T) => SumBy<T, number>;
+export function sumBy<T extends IterableContainer>(
+  callbackfn: (value: T[number], index: number, data: T) => bigint,
+): (items: T) => SumBy<T, bigint>;
 
 /**
  * Returns the sum of the elements of an array using the provided predicate.
  *
+ * Works for both `number` and `bigint` predicates, but not predicates that can return both
```
eranhirsch, line 50:
```suggestion
 * Works for both `number` and `bigint` mappers, but not mappers that can return both
```

src/sumBy.ts
```diff
@@ -11,17 +29,32 @@ import { purry } from "./purry";
...
-): (items: ReadonlyArray<T>) => number;
+export function sumBy<T extends IterableContainer>(
+  callbackfn: (value: T[number], index: number, data: T) => number,
+): (items: T) => SumBy<T, number>;
+export function sumBy<T extends IterableContainer>(
+  callbackfn: (value: T[number], index: number, data: T) => bigint,
+): (items: T) => SumBy<T, bigint>;
 
 /**
  * Returns the sum of the elements of an array using the provided predicate.
```
eranhirsch, line 48:
```suggestion
 * Returns the sum of the elements of an array using the provided mapper.
```

src/sumBy.ts
```diff
@@ -11,17 +29,32 @@ import { purry } from "./purry";
...
  *
+ * Works for both `number` and `bigint` predicates, but not predicates that can return both
+ * types.
+ *
+ * IMPORTANT: The result for empty arrays would be 0 (`number`) regardless of
+ * the type of the predicate; to avoid adding this to the return type for cases
```
eranhirsch, line 54:
```suggestion
 * the type of the mapper; to avoid adding this to the return type for cases
```

eranhirsch (approved):
🥳

Issue #561, opened 2024-03-07
Add type safe `curry` function


somnicattus:
Related to #560.

I finally found that curry is enough for most users. This worked well with generic-typed functions.

```ts
export type Curried<
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    F extends (...args: readonly any[]) => unknown,
> =
    Parameters<F> extends readonly [infer T, ...infer R extends unknown[]]
        ? (...args: R) => (arg: T) => ReturnType<F>
        : never;

export type NthCurried<
    N extends number,
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    F extends (...args: readonly any[]) => unknown,
    A extends unknown[] = [unknown],
> =
    Parameters<F> extends readonly [unknown, ...unknown[]]
        ? A['length'] extends N
            ? Curried<F>
            : NthCurried<N, Curried<F>, [unknown, ...A]>
        : never;

/**
 * Curry a function's parameter.
 *
 * @example
 * const pow = curry(Math.pow);
 * const result1 = Math.pow(2, 3); // 8
 * const result2 = pow(3)(2); // 8
 *
 * @example
 * const map = <I, O>(
 *     ...args: Parameters<Curried<typeof _map<I, O>>>
 * ): ReturnType<Curried<typeof _map<I, O>>> => curry(_map<I, O>)(...args);
 * const result1 = _map([1, 2], x => x * 2); // [2, 4]
 * const result2 = map((x: number) => x * 2)([1, 2]); // [2, 4]
 */
export function curry<
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    F extends (...args: readonly any[]) => unknown,
    N extends number | undefined,
>(fn: F, nth?: N): NthCurried<N extends number ? N : 1, F> {
    const curried = ((...args) =>
        (a0: Parameters<F>[0]) =>
            fn(a0, ...args)) as Curried<F>;

    return (nth === undefined || nth === 1 ? curried : curry(curried, nth - 1)) as NthCurried<
        N extends number ? N : 1,
        F
    >;
}
```

cjquines:
see also #458

eranhirsch:
This still requires the user to use a lot of manual typing that can't be inferred. I tried using it with this example and got typing issues.

```ts
const mapped = pipe(
  ['a', 'aa', 'aaa'],
  curry(map, (x) => x.length),
);
```

[...]

Issue #560, opened 2024-03-07
Add type safety for `purry`


somnicattus:
How do you think about this?

Lazy is ignored because I don't know much about it. 

```ts
export type PurriedParameters<
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    F extends (...args: readonly [any, any, ...unknown[]]) => unknown,
> =
    Parameters<F> extends readonly [unknown, ...infer R extends unknown[]]
        ? Parameters<F> | R
        : never;

// eslint-disable-next-line @typescript-eslint/no-explicit-any
export type PurriedReturnType<F extends (...args: readonly [any, any, ...unknown[]]) => unknown> =
    Parameters<F> extends readonly [infer I, ...unknown[]]
        ? ReturnType<F> | ((input: I) => ReturnType<F>)
        : never;

function purry<I, O, F extends (...args: readonly [I, any, ...unknown[]]) => O>(
    fn: F,
    args: PurriedParameters<F>,
): PurriedReturnType<F>
```

eranhirsch:
We don't really need a return type for purry because we instead rely on function overloads to define the purried overloaded type (which means `purry` is only used as an "implementation detail").

For external usage, purry as a tool, I'd still recommend people implement helper functions the same way we do.

So this brings me up to the question of can a typed purry *replace* the way we implement our functions substantially, removing the need to declare overloads to begin with, and only typing the dataFirst impl and getting everything else for free. To answer that question I think the main problem would be how do we handle **generic** functions, those that take a <T> for a param. Can you test to see if, for example, this purried type would work with `map` or `filter`? Does it still maintain the generic throughout the type?

e.g. 

```ts
declare function map<T, S>(array: T[], mapper: (item: T) => S): S[];

function purriedMap(...args: PurriedParameters<typeof map>): PurriedReturnType<typeof map> {
  return purry(map, args);
}

// Is x typed properly here?
const result = purriedMap([1,2,3], (x) => x + 1);

// what about here?
const result = pipe([1,2,3], purriedMap((x) => x + 1));
```

somnicattus:
I updated the code on the top. How about this?

eranhirsch:
I appreciate your intent, but it just doesn't work for generic functions (and can't as far as my experience with typescript goes).

[...]

PR #854, opened 2024-09-15
feat(partialBind): Add partialBind, partialRightBind

### LINKED ISSUE #458: Add partial, partialRight

cjquines:
part of #458 

typing data-last was not working well, and i decided it was not super useful in a data-last context anyway, so i omitted it

a partialBindObj function is in #855


src/partialBind.ts
```diff
@@ -0,0 +1,63 @@
...
+  : T["length"] extends 0
+    ? PartialBindError<"Too many args provided to function">
+    : T extends [infer THead, ...infer TRest]
+      ? Prefix extends [infer PrefixHead, ...infer PrefixRest]
+        ? PrefixHead extends THead
+          ? RemovePrefix<TRest, PrefixRest>
+          : PartialBindError<"Argument of the wrong type provided to function">
+        : Prefix extends Array<THead>
+          ? [THead?, ...RemovePrefix<TRest, Prefix>]
+          : never
```
eranhirsch, line 25: Shouldn't this use PatialBindError too?

src/partialBind.ts
```diff
@@ -0,0 +1,63 @@
...
+ *    const partialFn = R.partialBind([1, 2], fn)
+ *    partialFn(3) // => 1, 2, and 3
+ * @dataFirst
+ * @category Function
+ * @see partialRightBind
+ */
+export function partialBind<
+  T extends IterableContainer,
+  F extends (...args: any) => any,
+>(data: T, func: F): PartiallyBound<F, T> {
```
eranhirsch, line 61: I'm confused about the order of arguments here. Shouldn't the data-first order be reversed? The `data` here is the function, and the partial params list is the argument to the function. This is how you defined the type (with F before T), and that's how it's done in Lodash.

src/partialBind.ts
```diff
@@ -0,0 +1,63 @@
...
+            ? RemovePrefix<TRest, PrefixRest>
+            : PartialBindError<"Argument of the wrong type provided to function">
+          : Prefix extends TRest
+            ? TRest
+            : never
+        : never;
+
+type PartiallyBound<
+  F extends (...args: any) => any,
+  T extends IterableContainer,
```
eranhirsch, line 38: The usage of T and F here is confusing. `F` makes sense for the main type here, which is the function, but the partial arguments list should have an explicit name.

src/partialBind.ts
```diff
@@ -0,0 +1,63 @@
...
+          : never
+      : T extends [(infer THead)?, ...infer TRest]
+        ? Prefix extends [infer PrefixHead, ...infer PrefixRest]
+          ? PrefixHead extends THead
+            ? RemovePrefix<TRest, PrefixRest>
+            : PartialBindError<"Argument of the wrong type provided to function">
+          : Prefix extends TRest
+            ? TRest
+            : never
+        : never;
```
eranhirsch, line 34: Why aren't we using PartialBindError here?

src/partialBind.ts
```diff
@@ -0,0 +1,63 @@
...
+];
+
+type RemovePrefix<
+  T extends IterableContainer,
+  Prefix extends IterableContainer,
+> = Prefix["length"] extends 0
+  ? T
+  : T["length"] extends 0
+    ? PartialBindError<"Too many args provided to function">
+    : T extends [infer THead, ...infer TRest]
```
eranhirsch, line 18: Even if effectively T can never be readonly, we should make the type itself as generic as possible, especially considering its name considers it as a generic util. mutable tuples always extend the readonly version with the same shape.
```suggestion
    : T extends readonly [infer THead, ...infer TRest]
```

src/partialRightBind.test-d.ts
```diff
@@ -0,0 +1,100 @@
...
+    (
+      x: "RemedaTypeError(partialRightBind): Argument of the wrong type provided to function.",
+    ) => string
+  >();
+  expectTypeOf(
+    partialRightBind(["hello", "world"] as [string, ...Array<string>], foo),
+  ).toEqualTypeOf<(...parts: ReadonlyArray<string>) => string>();
+});
+
+it("should support readonly variadic params", () => {
```
eranhirsch, line 93: The name of the test should reflect that this doesn't actually work right now...

eranhirsch (commented):
1. Can we use TupleParts to simplify the main types here? 
2. I think the order of the arguments to the function is reversed, the function should come before the list of arguments.
3. We could potentially accept a variadic instead of an array, that seems to be how Lodash does it.

eranhirsch:
I found it a little hard to understand the logic behind the new types. Can you add some inline documentation to explain the different branches?

src/partialBind.test-d.ts
```diff
@@ -0,0 +1,94 @@
+/* eslint-disable unicorn/consistent-function-scoping */
+import { partialBind } from "./partialBind";
+
+const fn = (x: number, y: number, z: number): string => `${x}, ${y}, and ${z}`;
```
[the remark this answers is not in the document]
cjquines: hm yeah, i organized it into describe blocks and put it back in

cjquines:
1. i've looked into it, but couldn't figure out how to make this cleaner with TupleParts
2. done
3. i'd prefer we take an array rather than a variadic, but don't feel too strongly about this

src/partialBind.ts
```diff
@@ -15,49 +15,62 @@ type RemovePrefix<
...
+            : PartialBindError<"Argument of the wrong type provided to function">
+        : // We got passed a parameter list that isn't what we expected; this
+          // is an internal error.
+          PartialBindError<1>;
 
 type PartiallyBound<
   F extends (...args: any) => any,
-  T extends IterableContainer,
-> = (...rest: RemovePrefix<Parameters<F>, T>) => ReturnType<F>;
+  Partial extends IterableContainer,
```
eranhirsch, line 51: Don't use partial as the name because it's also a built-in utility type, which can be confusing (and will break grepping the code for Partial).

src/partialRightBind.ts
```diff
@@ -0,0 +1,76 @@
...
+ *    R.partialRightBind(func, partial)
+ * @example
+ *    const fn = (x, y, z) => `${x}, ${y}, and ${z}`
+ *    const partialFn = R.partialRightBind(fn, [2, 3])
+ *    partialFn(1) // => 1, 2, and 3
+ * @dataFirst
+ * @category Function
+ * @see partialBind
+ */
+export function partialRightBind<
```
eranhirsch, line 71: A thought about naming, we don't consider the right end of anything "right" in our naming, we call it "last", like `takeLast`, `dropLast`, and `last`... this is in contrast with Lodash which has inconsistent naming (they have both `last` and `takeRight`. This makes `partialRight` stand out as the only name using right to refer to the last elements.

Maybe it should be renamed `partialLast`?

src/partialBind.ts
```diff
@@ -0,0 +1,76 @@
...
+          PrefixHead extends THead
+          ? RemovePrefix<TRest, PrefixRest>
+          : PartialBindError<"Argument of the wrong type provided to function">
+        : Prefix extends ReadonlyArray<THead>
+          ? // T has a non-rest head, Prefix is an array.
+            // Prefix could possibly be empty, so this has to be THead?.
+            [THead?, ...RemovePrefix<TRest, Prefix>]
+          : // Prefix is e.g. [...Array<number>, string]. We can't do a
+            // type-level prefix removal, so we return an error.
+            PartialBindError<"Can't infer type of provided args">
```
eranhirsch, line 30: Should the error be that we can't *construct* the signature, not infer it, as it is being inferred properly, just that the types don't match?

src/partialBind.ts
```diff
@@ -0,0 +1,76 @@
...
+        ? Prefix extends readonly [infer PrefixHead, ...infer PrefixRest]
+          ? // Prefix has a non-rest head.
+            PrefixHead extends THead
+            ? RemovePrefix<TRest, PrefixRest>
+            : PartialBindError<"Argument of the wrong type provided to function">
+          : Prefix extends readonly [THead?, ...TRest]
+            ? // Prefix is an array. It must match *both* the optional type
+              // and the rest param.
+              TRest
+            : PartialBindError<"Argument of the wrong type provided to function">
```
eranhirsch, line 44: We shouldn't use the same string twice so when we get reports about it we can tell the two branches apart.

src/partialBind.test-d.ts
```diff
@@ -0,0 +1,242 @@
...
+    >();
+  });
+
+  test("should correctly type all partial args", () => {
+    expectTypeOf(partialBind(fn, [1, 2, 3])).toEqualTypeOf<() => string>();
+  });
+
+  test("should allow refined types", () => {
+    expectTypeOf(partialBind(fn, [1, 2, "a"])).toEqualTypeOf<() => string>();
+  });
```
eranhirsch, line 26: It looks like these tests are the same case, but they have different names. I think I'm missing something...

src/partialBind.test-d.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  test("should correctly type all partial args", () => {
+    expectTypeOf(partialBind(fn, ["hello", undefined, false])).toEqualTypeOf<
+      () => string
+    >();
+  });
+});
+
+describe("simple rest param case", () => {
```
eranhirsch, line 97: The usage with rest params is beautiful!

eranhirsch (commented):
This looks like a really powerful utility, but I'm still a little lost on when its useful so it's hard for me to assess how the API should be designed.

* We need to consider either recommending turning on the [eslint rule `default-param-last`](https://typescript-eslint.io/rules/default-param-last) and mentioning we don't support optional params in the middle of the argument list; or support it proper (if it isn't already, it was hard for me to track this via reading code alone. If it is we should test it explicitly).

* We can make the typing stricter by pulling some of the logic for computing the prefix to restrict the type of the partial function arguments list. This would mean that the errors are surfaced higher up as typescript errors (e.g. ts2322) instead of on the return type, and would allow typeahead and smarter completion in the ide. It will looks something like:

```ts
function partialBind<
  F extends (...args: any) => any,
  Args extends PartialArgumentsOf<F>,
>(func: F, args: Args): PartiallyBound<F, Args>;
```

src/partialBind.ts
```diff
@@ -0,0 +1,76 @@
...
+ *    const fn = (x, y, z) => `${x}, ${y}, and ${z}`
+ *    const partialFn = R.partialBind(fn, [1, 2])
+ *    partialFn(3) // => 1, 2, and 3
+ * @dataFirst
+ * @category Function
+ * @see partialRightBind
+ */
+export function partialBind<
+  F extends (...args: any) => any,
+  T extends IterableContainer,
```
eranhirsch, line 73: We still use T here, we should use the name we use in PartiallyBound

src/partialRightBind.ts
cjquines: hm, `partial` and `partialLast` maybe?

cjquines: or `partialBind` and `partialLastBind` idk

eranhirsch: I didn't mean to suggest partialLast, that is obviously not a good name, i just meant to show an example that uses last instead of left 🙃

eranhirsch:
Nice, this is almost like a sort of typed curry function that's really useful to replace all those non-curried: `($) => myFunc($, ...args)` hacks I end up adding to pipes.

eranhirsch (dismissed):
I wrote in my last review that we should see if we can create a type for the Suffix/Prefix parameters so that the error handling is done directly in the function call, not in the return type. This type would be similar to how `setPath`'s second parameter "inspects" the data argument and only accepts values that make sense. Considering a function F, it should be possible to create a type which is any prefix of the parameters of T:
```ts
type PrefixParameters<F extends (...args: any) => any> = ...
```

This might allow us to move some type-checking errors from RemedaError to a regular typescript one. It will also help with typeahead, where the IDE would show a type hint while typing out `partialBind(myFunc, ["stringParam", |...`.

I don't know how complex this type would be, how much can be reused from the existing solution, or if there are any caveats that make it harder to type. If we can create a hybrid approach where at least some of the nuance is moved from the return type to the argument, that is also a plus. I don't mind shipping this kind of change later on as an improvement, so this isn't a show-stopper right now.

Regardless, to match the naming convention of most of our utils, the name picked by Lodash and Ramda is not ideal, as it refers to the result (kinda?) and not what the function _does_. Like map and filter describe what the utility does, not what the output is (mapped, filtered), the name of this function should be `partiallyBind` and `partiallyBindLast`. Naming is hard and usually inconsequential, so I'm just mentioning it for your discretion.

cjquines:
ooops, i forgot to respond to some of the last review comments

> support optional params in the middle of the argument list

[typescript itself prevents you](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABMAFADwFyLCAtgIwFMAnAGkQE8B+LAZymJjAHNyAvLfOOAG0IEMwASkQBvAL5A) from writing something like

```ts
function f(x: number, y?: string, z: boolean) {}
```

> I wrote in my last review that we should see if we can create a type for the Suffix/Prefix parameters so that the error handling is done directly in the function call, not in the return type.

will do this

> the name of this function should be `partiallyBind` and `partiallyBindLast`

hm, i also think `partialBind` works. the analogy would be like, `flatMap` and `flatlyMap`. consider the sentence "The utility `partialBind` does a partial bind of the function passed in."

cjquines:
okay, i have an idea about how to do this, but it's going to need more code, so i'll write the helpers first

src/internal/types.ts
```diff
@@ -273,12 +295,63 @@ export type TupleParts<
...
+        ? TupleParts<Tail, [...Prefix, MaybeHead?], Suffix>
+        : never
+      : // This is the Array<number | string> case.
+        T extends ReadonlyArray<infer Item>
+        ? {
+            prefix: Prefix;
+            item: Item;
+            suffix: Suffix;
+          }
+        : never;
```
cjquines, line 312: **note:** this does the optional changes for #863

src/internal/types.test-d.ts
cjquines, line 1: **note:** i think we should split this file, but we can do that later

cjquines:
more data re naming:

* pretty much every js/ts util library calls it partial
* python calls it [partial](https://docs.python.org/3/library/functools.html#functools.partial)
* c++ calls it [bind, bind_front, bind_back](https://en.cppreference.com/w/cpp/utility/functional/bind_front)
* java calls it [bindTo](https://docs.oracle.com/javase/7/docs/api/java/lang/invoke/MethodHandle.html#bindTo(java.lang.Object)) and [insertArguments](https://docs.oracle.com/javase/7/docs/api/java/lang/invoke/MethodHandles.html#insertArguments(java.lang.invoke.MethodHandle,%20int,%20java.lang.Object...))

eranhirsch (dismissed):
This is really good! Do you think we can link and close the issues about better typing for purry? do you think we should mention these functions in the documentation of purry?

src/internal/types.ts
```diff
@@ -273,12 +295,63 @@ export type TupleParts<
...
+type FixedTupleSplits<
+  L extends IterableContainer,
+  R extends IterableContainer = [],
+> = L extends readonly []
+  ? { left: L; right: R }
+  : L extends readonly [...infer LHead, infer LTail]
+    ? { left: L; right: R } | FixedTupleSplits<LHead, [LTail, ...R]>
+    : L extends readonly [...infer LHead, (infer LTail)?]
+      ? { left: L; right: R } | FixedTupleSplits<LHead, [LTail?, ...R]>
       : never;
```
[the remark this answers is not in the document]
cjquines: i'm surprised—i remember tests breaking when i tried this, but maybe things changed

src/partialBind.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+import type {
+  IterableContainer,
+  RemedaTypeError,
+  TupleSplits,
+} from "./internal/types";
+
+type PartialBindError<Message extends string | number> = [
+  RemedaTypeError<"partialBind", Message>,
+];
```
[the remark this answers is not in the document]
cjquines: ah, this is a leftover from the old error type

src/partialBind.ts
```diff
@@ -0,0 +1,69 @@
...
+
+type TuplePrefix<T extends IterableContainer> = TupleSplits<T>["left"];
+
+type RemovePrefix<
+  T extends IterableContainer,
+  Prefix extends TuplePrefix<T>,
+> = Prefix["length"] extends 0
+  ? T
+  : T extends readonly [infer THead, ...infer TRest]
+    ? Prefix extends readonly [infer _PrefixHead, ...infer PrefixRest]
```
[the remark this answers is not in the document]
cjquines: we refer to the name in the comments, so i think i'd keep

cjquines:
not sure which issues you're talking about, where this would be relevant?

eranhirsch:
> not sure which issues you're talking about, where this would be relevant?

#560, #561?

eranhirsch (approved):
I feel that the documentation for these functions doesn't cover them well enough for people not coming from Lodash/Ramda. I think that we should mention the concept of binding and what it means more clearly, and perhaps provide examples that show where these functions are useful. There is also a remaining unresolved comment about an edge-case that our typing doesn't currently support.

PR #855, opened 2024-09-15
feat(partialBindObj): Add partialBindObj

cjquines:
part of #458

PR #863, opened 2024-09-19
feat(splice): Make splice typing stricter

cjquines:
i chose `feat` rather than `fix` because this is a large type change


src/splice.ts
```diff
@@ -1,12 +1,104 @@
...
+  Sum,
+} from "type-fest";
+import type {
+  CoercedArray,
+  IterableContainer,
+  TupleParts,
+} from "./internal/types";
 import { purry } from "./purry";
 
+type FixedLengthSplice<
```
eranhirsch, line 15: I'd simplify this with more helper utility types. The logic is pretty straightforward, but it isn't apparent when inlined like this.

This is how I simplified it while reading the code:

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgSQM4DkCmBzAhjYAN0wBo4BlAVxDIHEpN9MoAVAC1wDsyqAjGKLgDGMMgEEoggJ7kwAG2BDMcAL5wAZlAgg4AIhhSwmALTrMqGLoDcAKBsGjKGM1y85mAMIROMXME7MADwscAC8cJScANacEADunAB8YXAMuAAm3nJScCwA2gC6cAA+cIW29obKyOoAMsDOgnKB6HCYAB7OnOmocJzUvMxkAGK4cnK8wlHJ4f0gg1BtnZjdva0A-HCj45NCUXAAXHDoFQ7KYiKUYwDy6uqomDCBNnC5S109Ti5unt6+-kMXsd3itPnMFiQbDMUBgcPgiJgWskOh9euoxg8gZt0ECjuDmCDVrk8rp3JxsDA2LoCli+gNmLiUHUGi5mlQQMESWSKVSCmR0IkyAAGRKnKpwC4wK5yWorHnPV4hFGg3rIRqudxeHx+AJQSGvW73R6EsH0vVA1rKon483QtBYPAEYhIk29ASUTC0oWMm2u4mkuWU6m0m2M+iMRrsLiBdmBQ0PUTHQX+7lBgrI5ZE92e16bGr1dVsyj8QQiTkB8lpsjxx7JkWMk52M5bYDtTDpWWVtiyBRKBVvK2fNXfTV-HWA17kXywP02-VwAAimHczi8kXgg96c6BACVMPJhJgQCsN5mh+qflr-rr5wBRO6YEQIqe4GebunzAnhSXSmtPFg8NOMCClCKQSNIPaKIiQIAUC95mE+xAvrA84-mMnbygBcDwY+TqYMhiZLiuvzriBrx7geSjHj4UK2EAA

src/splice.ts
```diff
@@ -1,12 +1,104 @@
...
+  Start extends number,
+  DeleteCount extends number,
+  Replacement extends IterableContainer,
+> = TupleParts<T>["item"] extends never
+  ? FixedLengthSplice<T, Start, DeleteCount, Replacement>
+  : IsNegative<Start> extends true
+    ? // type-fest Sum returns number when the sum is negative.
+      number extends Sum<TupleParts<T>["suffix"]["length"], Start>
+      ? // Splice cuts into the variable-length part.
+        T
```
eranhirsch, line 60: Do you think there should be a TODO here? This is why we have a known issue in the tests, right?

src/splice.ts
```diff
@@ -1,12 +1,104 @@
...
+            DeleteCount,
+            Replacement
+          >,
+        ]
+    : GreaterThan<
+          Sum<Start, DeleteCount>,
+          TupleParts<T>["prefix"]["length"]
+        > extends true
+      ? // Splice cuts into the variable-length part.
+        T
```
eranhirsch, line 77: And similarly here...

eranhirsch (commented):
Really cool! Do you think we can solve the known issues, or at least some cases within them so that the type is more useful for regular arrays.

If we get the type correctly this would cover the return type for all the slice-related functions: take, drop, takeLast, dropLast.

src/splice.ts
cjquines: i was thinking about this... i think we could just mark the suffix optional and it'd work? let me try

eranhirsch:
@cjquines, this PR still has several comments that aren't marked as resolved. Would it be OK to move it to draft if you are still planning on addressing those issues so that the review queue is clear? If you are waiting for a review on this, my biggest issue is still that the replacement array type isn't tied to the data type, allowing inserting incompatible types into the array.

cjquines:
i am planning on addressing these, but these have been in my backlog. i'll get to these today

src/splice.test-d.ts
```diff
@@ -1,21 +1,119 @@
...
+      [1, 4, 2, 3]
+    >();
+  });
+
+  it("works with start >= length", () => {
+    expectTypeOf(splice([1, 2, 3] as const, 3, 0, [4] as const)).toEqualTypeOf<
+      [1, 2, 3, 4]
+    >();
+  });
+
```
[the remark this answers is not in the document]
cjquines: oh good catch, that was wrong

src/splice.test.ts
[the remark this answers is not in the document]
cjquines: it's already a test.each?

src/internal/types.ts
```diff
@@ -255,15 +255,24 @@ export type NTuple<
...
  * rest parameter (if any), and the item is the type of the rest parameter.
  *
  * The output could be used to reconstruct the input: `[
  *   ...TupleParts<T>["prefix"],
- *   ...Array<TupleParts<T>["item"]>,
+ *   ...CoercedArray<TupleParts<T>["item"]>,
  *   ...TupleParts<T>["suffix"],
  * ]`.
+ *
+ * Optionals in a tuple are always considered to be part of the rest parameter.
```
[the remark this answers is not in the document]
cjquines: pretty sure

cjquines:
oh dear. the type tests are failing in 5.1 and 5.2 but not 5.3. don't know how to fix this

src/splice.test.ts
```diff
@@ -98,7 +98,7 @@ describe("regression test including edge cases", () => {
       { items: [   ], start:  1, deleteCount:  0, replacement: [3], expected: [3] },
       { items: [   ], start:  1, deleteCount:  1, replacement: [ ], expected: [] },
       { items: [   ], start:  1, deleteCount:  1, replacement: [3], expected: [3] },
-    ] as const;
+    ];
```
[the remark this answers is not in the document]
eranhirsch: #889

src/splice.ts
```diff
@@ -15,19 +107,29 @@ import { purry } from "./purry";
...
+  start: Start,
+  deleteCount: DeleteCount,
+  replacement: Replacement,
+): Splice<T, Start, DeleteCount, Replacement>;
 
 /**
  * Removes elements from an array and, inserts new elements in their place.
  *
- * @param start - The index from which to start removing elements.
+ * Equivalent to `items.toSpliced(start, deleteCount, ...replacement)`.
```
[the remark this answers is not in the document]
eranhirsch: https://github.com/search?q=repo%3Aremeda%2Fremeda+developer.mozilla.org+language%3ATypeScript&type=code&l=TypeScript

src/splice.ts
```diff
@@ -1,12 +1,133 @@
...
+  WidenLiteral,
+} from "./internal/types";
 import { purry } from "./purry";
 
+type IfLiteral<N extends number, Fallback> = number extends N ? Fallback : N;
+
+type PositiveIndex<
+  T extends IterableContainer,
+  N extends number,
+> = N extends unknown
```
eranhirsch, line 23: Can you document why we need to check that N extends unknown here?

src/splice.ts
```diff
@@ -1,12 +1,133 @@
...
+  ? IsNegative<N> extends false
```
eranhirsch, line 24: This is just a nit but let's align with https://eslint.org/docs/latest/rules/no-negated-condition for types too and check the extends true branch first.

src/splice.ts
```diff
@@ -1,12 +1,133 @@
...
+    ? GreaterThan<N, T["length"]> extends true
+      ? T["length"]
+      : N
+    : number extends T["length"]
+      ? number
+      : IfLiteral<Sum<T["length"], N>, 0>
```
eranhirsch, line 30: Can you document why 0 is the fallback? It's not clear to me; intuitively, it feels like it should be `never`, but I don't know without an explanation.

src/splice.ts
```diff
@@ -1,12 +1,133 @@
...
+        ? IfLiteral<Subtract<T["length"], Offset>, 0>
+        : Length;
+
+type FixedLengthSplice<
+  T extends IterableContainer,
+  Start extends number,
+  DeleteCount extends number,
+  Replacement extends IterableContainer,
+  PositiveStart extends number = PositiveIndex<T, Start>,
+> = PositiveStart extends unknown
```
eranhirsch, line 52: Similarly, can you document why we check the `extends unknown` here?

src/splice.ts
```diff
@@ -1,12 +1,133 @@
...
+      PositiveStart,
+      ClampedLength<T, PositiveStart, DeleteCount>,
+      Replacement
+    >
+  : never;
+
+/** Widen a tuple into an array type. */
+type WidenTuple<T extends IterableContainer> =
+  T extends ReadonlyArray<infer Item>
+    ? IfNever<Item, ReadonlyArray<unknown>, Array<WidenLiteral<Item>>>
```
eranhirsch, line 64: I see you wrap WidenTuple everywhere with Readonly<>, why not define it as readonly here?

```suggestion
    ? IfNever<Item, ReadonlyArray<unknown>, ReadonlyArray<WidenLiteral<Item>>>
```

src/splice.ts
```diff
@@ -15,19 +136,29 @@ import { purry } from "./purry";
...
-  items: ReadonlyArray<T>,
-  start: number,
-  deleteCount: number,
-  replacement: ReadonlyArray<T>,
-): Array<T>;
+export function splice<
+  T extends IterableContainer,
+  Start extends number,
+  DeleteCount extends number,
+  Replacement extends Readonly<WidenTuple<T>>,
```
eranhirsch, line 143: Wouldn't this typing cause typo's in arrays of enum (literal union) values break silently, e.g.

```ts
type MyEnum = "cat" | "dog";

const DATA = ["cat"] as const satisfies readonly MyEnum[];

splice(DATA, 0, 0, ["bat"]);
```

src/splice.test-d.ts
```diff
@@ -1,21 +1,179 @@
...
+      expectTypeOf(
+        splice([1, 2, 3] as const, 1, 2, [4] as const),
+      ).toEqualTypeOf<[1, 4]>();
+    });
+
+    it("works with deleteCount + start > length", () => {
+      expectTypeOf(
+        splice([1, 2, 3] as const, 1, 4, [4] as const),
+      ).toEqualTypeOf<[1, 4]>();
+    });
```
eranhirsch, line 85: Let's add some values to the input array in the tests so that they expect different outputs. This will help protect against cases where a bug breaks one of them but hides it because the input is too simple.

src/splice.test-d.ts
```diff
@@ -1,21 +1,179 @@
...
+    expectTypeOf(
+      splice([1, 2, 3] as const, 1, 1, [4, 5, 6] as const),
+    ).toEqualTypeOf<[1, 4, 5, 6, 3]>();
+  });
+
+  describe("non-literal params", () => {
+    it("works as number", () => {
+      expectTypeOf(
+        splice([1, 2, 3] as const, 1 as number, 1 as number, [4] as const),
+      ).toEqualTypeOf<Array<number>>();
```
eranhirsch, line 104: This isn't ideal, it widens the output too much, we still know that the output here is `Array<1 | 2 | 3 | 4>`;

src/splice.test-d.ts
```diff
@@ -1,21 +1,179 @@
...
+    >();
+  });
+
+  it("works with big replacement", () => {
+    expectTypeOf(
+      splice([1, 2, 3] as const, 1, 1, [4, 5, 6] as const),
+    ).toEqualTypeOf<[1, 4, 5, 6, 3]>();
+  });
+
+  describe("non-literal params", () => {
```
eranhirsch, line 100: The describe here isn't exact, 1 | 2 is considered a literal (at least by type-fest's defintion).

eranhirsch (commented):
I'm still concerned about the typing for the replacement, I feel that widening the type is not the correct thing to do, as it doesn't protect against inserting typos into arrays of enums, which I feel could be common for use-cases like filters/searches into tables of data.

src/internal/types.ts
```diff
@@ -198,11 +200,31 @@ export type UpsertProp<T, K extends PropertyKey, V> = Simplify<
...
 type IsSingleLiteral<K> =
   IsLiteral<K> extends true ? (IsUnion<K> extends true ? false : true) : false;
 
+/**
+ * Widen a literal type.
+ *
+ * @example WidenLiteral<"cat" | "dog"> //=> string
+ * @example WidenLiteral<"cat" | 1> //=> string | number
+ */
+export type WidenLiteral<K> = K extends unknown
```
eranhirsch, line 216: Now that I read the whole code, the name here is confusing because it doesn't just run on literals. It's also a bit limited, if I had a union of objects, wouldn't we expect to widen it just the same? e.g.

```ts
type Narrow = { a: "cat" } | { a: "dog" };

type Widened = WidenLiteral<Narrow>;
//   ^? { a: string };
```

cjquines:
> I'm still concerned about the typing for the replacement, I feel that widening the type is not the correct thing to do, as it doesn't protect against inserting typos into arrays of enums, which I feel could be common for use-cases like filters/searches into tables of data.

hm, that's true, yeah. i think we can just not widen it, and instead decide that those cases should get explicit type casts

eranhirsch:
@cjquines - are you planning on getting back to this or can this PR be closed?

cjquines:
yes (and same for the other ones). i'll try to get this done soon, and then #855 and #869

src/internal/types.ts
```diff
@@ -273,13 +304,27 @@ export type TupleParts<
...
-          suffix: Suffix;
-        }
-      : never;
+    : // We need to distinguish between e.g. [number? ...Array<string>] and
+      // Array<number | string>.
+      IsTupleRestOnly<T> extends false
+      ? // This is the [number? ...Array<string>] case.
+        T extends readonly [(infer MaybeHead)?, ...infer Tail]
+        ? TupleParts<Tail, [...Prefix, MaybeHead?], Suffix>
+        : never
```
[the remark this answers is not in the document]
cjquines: not clear, i think we should only use it for top-level function types…

src/splice.test-d.ts
```diff
@@ -1,24 +1,173 @@
...
+    expectTypeOf(splice([], 0, 0, [] as Array<number>)).toEqualTypeOf<
+      Array<number>
+    >();
+  });
+
+  it("reflects both `items` and `replacement` in the output type", () => {
+    expectTypeOf(
+      splice([] as Array<number>, 0, 0, [] as Array<string>),
+    ).toEqualTypeOf<Array<number | string>>();
+  });
```
cjquines, line 20: **note:** this could lead to mistakes... but i think i'd rather compromise like this

cjquines: in particular, the output type will be different if there's a mistake, so i'm somewhat fine

cjquines:
not sure why tests are failing, they work locally. we'll see

eranhirsch:
oops, i was too eager deleting old branches of mine, I didn't notice I deleted yours...

eranhirsch:
@cjquines - let's close this one out for now, we can redo it later on from scratch starting with up to date versions of tuple-parts etc...

PR #868, opened 2024-09-27
perf(isEmpty): Use better algorithm to determine object emptiness

C-Higgins:
The prior implementation is O(n). This can be done in O(1).

You may also swap it for `Object.prototype.hasOwnProperty.call(data, key)` if `hasOwn` is not permitted for compatibility.

### DIFF
--- a/src/isEmpty.ts
+++ b/src/isEmpty.ts
@@ -48,5 +48,10 @@ export function isEmpty(data: object | string | undefined): boolean {
     return data.length === 0;
   }
 
-  return Object.keys(data).length === 0;
+  for (const key in data) {
+    if (Object.hasOwn(data, key)) {
+      return false;
+    }
+  }
+  return true;
 }


C-Higgins:
It also seems to me like you could simply write it like this. But I have not seen any other libraries do it this way:

```js
for (const _ in data) { return false };
return true;
```

cjquines:
`hasOwn` is supported by our minimum environment (es2022) so it's fine i think

src/isEmpty.ts
```diff
@@ -48,5 +48,10 @@ export function isEmpty(data: object | string | undefined): boolean {
     return data.length === 0;
   }
 
-  return Object.keys(data).length === 0;
+  for (const key in data) {
+    if (Object.hasOwn(data, key)) {
+      return false;
+    }
+  }
```
cjquines, line 55: **question:** does using `of` and then omitting the `hasOwn` work?

eranhirsch:
Although this looks like an improvement from an algorithmic point of view, I failed to come up with a test plan that shows it has any advantage over the existing implementation. My best guess is that v8 has optimizations for `Object.keys(data).length === 0` as that is such an idiomatic way to test for object emptiness.

I ran some benchmarks that didn't come out in favor of making the change. This ran on an x86 machine, I don't know how an ARM/Mac Mx processor would handle it differently.

Benchmark test:
https://gist.github.com/eranhirsch/a8a554cd325e42c3b41ee7a332193e9b

Results:
https://gist.github.com/eranhirsch/c0fa162cf8523e05fc2ff06d31760101

C-Higgins:
> Although this looks like an improvement from an algorithmic point of view, I failed to come up with a test plan that shows it has any advantage over the existing implementation. My best guess is that v8 has optimizations for `Object.keys(data).length === 0` as that is such an idiomatic way to test for object emptiness.
> 
> I ran some benchmarks that didn't come out in favor of making the change. This ran on an x86 machine, I don't know how an ARM/Mac Mx processor would handle it differently.
> 
> Benchmark test: https://gist.github.com/eranhirsch/a8a554cd325e42c3b41ee7a332193e9b
> 
> Results: https://gist.github.com/eranhirsch/c0fa162cf8523e05fc2ff06d31760101

Interesting. Could you please test again with sizes in the range of ~10 million?
There is also potential GC impact here. Maybe v8 is smart enough to not actually allocate the array on the heap, but if not I don't think the GC lag would appear on this test.

src/isEmpty.ts
C-Higgins: Objects aren't iterable by default, so forof won't work on them. For forin, it will loop over the inherited properties as well, so an empty object with a non-empty prototype chain would return not empty unless the hasOwn check is done.

cjquines: that makes sense

cjquines:
i think that's a v8 specific optimization. [here's a jsperf](https://jsperf.app/wakoje), firefox shows a noticeable difference

C-Higgins:
> i think that's a v8 specific optimization. [here's a jsperf](https://jsperf.app/wakoje), firefox shows a noticeable difference

I see that large difference as well on Firefox. Around 3 orders of magnitude faster with the loop. On Brave (V8) there is no difference.

eranhirsch:
>Could you please test again with sizes in the range of ~10 million?

Is this a relevant real-world scenario where one would need to check if an object has either 10m entries or 0? How often would anyone be handling 10m-entries objects (in JS) anyway? Aren't there better tools for those cases in general?

> Maybe looping each test numerous times would reveal that.

You can see in the benchmark results that each function was run about 6K times on each input size.

---


[I slightly expanded the test](https://jsperf.app/wakoje/4
), ran it on Chrome, Safari, and Firefox on a non-intel Mac, and got this:


Screenshot 2024-09-29 at 10 40 50
Screenshot 2024-09-29 at 10 41 15
Screenshot 2024-09-29 at 10 45 09

The Firefox result looks so out that it seems like the benchmarking tool is broken; I don't know what to make of it.

We know that the Node/v8 results don't justify the change, and Safari would suffer. I don't see a reason to optimize for Firefox, even if the benchmark results seemed reasonable, as it isn't a popular browser.

C-Higgins:
Safari does not suffer; it retains the same performance as the current implementation. It suffers in comparison to an implementation that is not used right now. Every benchmark either improves or stays the same (within error) with the implementation suggested in this PR. I don't see why you wouldn't make this change.

eranhirsch:
We can't compare the number of ops/sec between browsers because that would also include whatever the browser is doing between tests and how it wraps the tested lines. We can only benchmark when we nullify any biases, and comparing between browsers by absolute numbers introduces such biases. It's clear that Firefox isn't effectively **x110K faster** than Chrome, and what we see in the result is more likely due to an optimization Firefox is doing because of how this test is implemented, most likely caching the function partially, or executing the code effectively before the timer is started.

[If we move the object generation into the test](https://jsperf.app/wakoje/5), the results align with expectations (this time run on an Intel x86 machine):

Screenshot 2024-09-29 220640

image

Screenshot 2024-09-29 220133

Also looking at the error bars the test reports (of +-10%) I sense that the benchmarks aren't being run enough times to get a stat-sig result.

eranhirsch:
Here's a stackoverflow from 7 years ago which probably explains the x100k perf benefit we see above: https://stackoverflow.com/questions/41419724/why-does-firefox-execute-javascript-incredibly-faster-than-other-browsers

In short, Firefox likely didn't even run the tested function, noticing that the return value is ignored (so the function call is redundant) and being able to ensure it doesn't have side effects because it's the only implementation without any function calls.

C-Higgins:
Putting the object generation into the test isn't valid, because the object creation is the slow part and it dominates the speed of the emptiness check. The test results just become the speed of creating the object.

[Here](https://jsperf.app/wakoje/8) is a benchmark using console.log to force the output to be evaluated. You can see in the console that `false` is being logged each time. The results:
image


[Here](https://jsperf.app/wakoje/9)'s another test, again with logging, with an object that is randomly empty or full, to attempt to make the JIT ineffective. 
image

What's interesting is that this performs better even on V8
image


I don't know what would make these results incorrect. Also, would it not make sense to merge this *regardless* of any test showing benefit, if only because it's clearly algorithmically better, and this library may run on older hardware lacking the optimizations that give it parity? No tests have shown the loop to be worse in any context.

eranhirsch:
> [...] would it not make sense to merge this regardless of any test showing benefit, if only because it's clearly algorithmically better, and this library may run on older interpreters lacking the optimizations that give it parity? No tests have shown the loop to be worse in any context.

No. I don't know how v8 implements simple objects under the hood, which might impact perf more than from a pure algorithmic point of view. V8 might be maintaining each object's size already, and it might be optimized to access that number in O(1) when it sees code like `Object.keys(data).length`; it might also have the array of keys already in memory as an array, so that `Object.keys(data)` might be O(1), and if the array is also optimized it might have a size property on it so that accessing `.length` might be a second O(1) call. On the other hand, we know nothing about implementing for-in loops and how they are implemented; perhaps v8 needs to generate an array of keys on the fly, effectively calling `Object.keys` under the hood. We don't know anything about the implementation of `Object.hasOwn` - maybe it's expensive? Regardless, iterating over anything in JavaScript code has costs for managing the loop. When you call `Object.keys(data).length === 0`, you allow the engine to go into "engine-land" to implement everything, returning just a single result.

My point is that you don't _know_ a lot about the engines' implementation, so even from an algorithmic point of view, you don't have enough information to say that this implementation is better. That's why I rely on benchmarks; I don't trust my gut feeling. You say this is "clearly algorithmically better" - I'm just saying it's not that clear.

C-Higgins:
Fortunately we do have benchmarks that show V8 does not suffer from this change.

eranhirsch:
I'm entertaining this discussion because it has some value to others who might stumble upon it, but you are not participating in good faith. I don't have an infinite amount of time to perform benchmarks and reviews, and doing this means that that time is not used for other features and requests that are more valuable to other users. I'd be happy to consider any perf improvements as long as they can be shown to be meaningful; the benchmark you provided shows an improvement of x100, which is unlikely - you have been critical to any attempts of mine to further the discussion, but remain uncritical to any results you bring, even when they don't make sense.

I'll be happy to discuss this further once you provide a result that can show there is value in this change. I cannot put more time and effort on this PR at this time.

C-Higgins:
> I'm entertaining this discussion because it has some value to others who might stumble upon it, but you are not participating in good faith. I don't have an infinite amount of time to perform benchmarks, and doing this means that that time is not used for other features and requests that are more valuable to other users. I'd be happy to consider any perf improvements as long as they can be shown to be meaningful; the benchmark you provided shows an improvement of x100, which is unlikely - you have been critical to any attempts of mine to further the discussion, but remain uncritical to any results you bring, even when they don't make sense.
> 
> I'll be happy to discuss this further once you provide a result that can show there is value in this change. I cannot put more time and effort on this PR at this time.

100x speedup is *not* unlikely, because I am using an object of size 100,000 in these tests. It is perfectly plausible that constructing a 100,000 length array and checking the length of it is 100x slower than constructing a loop and returning from it immediately. And that is what the latest two benchmarks show, in Firefox. Let's review the thread:

Pros:
- Loop implementation is, naively, a more performant algorithm
- Benchmarks that we both devised have shown no significant difference in V8
- Benchmarks have shown the loop is much better in Firefox
- Lodash has been using this implementation for ~8 years, and they have far more eyes on their code than this library does

Cons:
- Interpreter optimizations may cause the `keys` implementation to be faster overall. There is no evidence of this.
- The benchmarks for Firefox may be incorrect. Okay, but there is no evidence of this; my last benchmarks with the logger makes this unlikely.
- It is more lines of code

From my perspective, you are spending a lot of effort to deny this change, while there is no evidence against it and some evidence in favor of it. If I appear to be arguing in bad faith, it's because you appear to be unreasonably attached to the existing implementation. As you said, we don't know the interpreter optimizations, so we must rely on benchmarks, which thus far have favored this change. Naive intuition also favors this change. There is no evidence against this change. Neither of us have infinite time to run benchmarks, so *with the information we have* it would make sense to merge this and move on. I don't see any reason for me to participate in this issue further, because it is unclear to me what evidence, if any, of Firefox improvements would justify this change to you, when the current benchmarks have not.

cjquines:
the firefox data is strange. i compiled spidermonkey locally, and in fact, the current implementation is 4x faster. so there's definitely something going on in the browser here

PR #869, opened 2024-09-30
chore(jsdoc): deduplicate jsdocs between overloads

cjquines:
pr is still pretty draft. the idea is:

* one jsdoc in a file is marked as `copyDoc`
* others might be marked with `pasteDoc`
* we have a build step where we copy the jsdoc from the one with `copyDoc` to the one with `pasteDoc`, and apply modifications if needed. right now we support:
    * if `copyDoc` has `dataFirst` and `pasteDoc` has `dataLast`, remove the first param and change the tag
    * if `pasteDoc` has a tag that the `copyDoc` also has, it overrides

there's two ways we can do this:

* there's an ad hoc parsing script in this pr, at `scripts/copyDocs.ts`
* there's also a jscodeshift codemod, at `scripts/copyDocsTransform.ts`. this adds extra dependencies compared to the parsing script, but it's likely more maintainable in the future + opens up more transform possibilities

questions / todos before we merge:

* should we do this at all?
* how should we organize this? right now i have it in a `scripts` folder, but it could feasibly get its own `package.json` inside it like we do for docs
* how do we test this? i think we should test this to avoid breaking build. right now i think comparing a snapshot of the `add.d.ts` files might work; it's a file we expect to change infrequently i think
* audit the rest of the files to see if `copyDoc` / `pasteDoc` fits their needs

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -208,6 +208,8 @@ export default tseslint.config(
             "lazy",
             "strict",
             "category",
+            "copyDoc",
+            "pasteDoc",
           ],
         },
       ],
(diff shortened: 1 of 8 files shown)


eranhirsch:
@cjquines - Are you planning on getting back to this or can this PR be closed?

Issue #874, opened 2024-10-02
Drop support for `bigint` in `mean`


grigorischristainas:
It would be helpful to have these utilities, equivalent to https://ramdajs.com/docs/#mean and https://ramdajs.com/docs/#median. 

If you support their addition, I would be happy to contribute.

grigorischristainas:
@eranhirsch I noticed that `meanBy` returns NaN in case of an empty array. Maybe it would be better to return NaN in the new utilities as well for consistency? This could also be helpful when migrating from Lodash or Ramda, as their corresponding utilities return NaN too.

cjquines:
i dislike it, but i'd vote for `NaN` for consistency. builtins like `parseInt` or `parseFloat` return `NaN` when something isn't parseable, rather than `undefined`

if we _really_ wanted to, we could do something like

```ts
declare const TAG_NAN: unique symbol;

type NaN = Tagged<number, typeof TAG_NAN>;
```

and make everything return `number | NaN` instead. but i also don't like this. it would mean having to change the types of everything that returns a number

eranhirsch:
Lodash and Ramda are effectively untyped, as is the Javascript (ECMAScript) standard library; with Remeda being strictly typed, we should reconsider these decisions. Our main goal with better typing is to move errors from **runtime** to **compile time**; returning `NaN` is an example of a value that is not valuable to users and probably means something went wrong, but because it's typed as `number,` it is hard to detect at compile time and protect against.

Javascript has a lot of legacy decisions that don't make sense in retrospect; for example, `3/0` is `Infinity`, although the math says it should be `undefined` (https://en.wikipedia.org/wiki/Division_by_zero). NaN is an artifact of [double precision floating point](https://en.wikipedia.org/wiki/Double-precision_floating-point_format) where the APU has no other means of "throwing an exception" other than turning on bits of the output that effectively say that the value returned is not the result of the operation. It is not a mathematical concept. Javascript has multiple values that all mean "not a regular value": `undefined`, `null`, `NaN`, and this makes code brittle and confusing to maintain. There is already [a movement to stop using `null`](https://github.com/sindresorhus/meta/discussions/7) and I feel that `NaN` should be put in the same bucket, especially since it isn't typed differently (at least `null` is a concrete type).

[...]

PR #878, opened 2024-10-06
feat(mean): Add mean

grigorischristainas:
Related to #874. Please note that I utilized the `sum` utility (not sure if this is something you would prefer to be avoided). I'd be happy to make any further adjustments as needed.

---

Make sure that you:

---


src/mean.ts
```diff
@@ -0,0 +1,87 @@
...
+  T extends readonly []
+    ? undefined
+    : // Non-empty bigint arrays will always result in bigint.
+      T extends readonly [bigint, ...ReadonlyArray<unknown>]
+      ? bigint
+      : // But an empty bigint array would result in undefined.
+        T[number] extends bigint
+        ? undefined
+        : // Non-bigint arrays are always handled correctly.
+          number;
```
cjquines, line 16: **question:** shouldn't this be `number | undefined`?

src/mean.ts
```diff
@@ -0,0 +1,87 @@
...
+type Mean<T extends IterableContainer<bigint> | IterableContainer<number>> =
+  // Empty arrays would always result in undefined.
+  T extends readonly []
+    ? undefined
+    : // Non-empty bigint arrays will always result in bigint.
+      T extends readonly [bigint, ...ReadonlyArray<unknown>]
+      ? bigint
+      : // But an empty bigint array would result in undefined.
+        T[number] extends bigint
+        ? undefined
```
cjquines, line 14: **nit:** this is always covered by the earlier case anyway right? we can remove this branch

src/mean.ts
cjquines, line 16: **suggestion:** let's add a check for non-empty `number` arrays

src/mean.ts
```diff
@@ -0,0 +1,87 @@
...
+ * @example
+ *   R.mean([1, 2, 3]); // => 2
+ *   R.mean([1n, 2n, 3n]); // => 2n
+ *   R.mean([]); // => undefined
+ * @dataFirst
+ * @category Number
+ */
+export function mean<
+  T extends IterableContainer<bigint> | IterableContainer<number>,
+>(data: T): Mean<T>;
```
cjquines, line 42: **thought:** in theory we could provide a more specific type using the `Positive` or `Negative` or `Finite` types from `type-fest`, but idk if we want to do that shrug

src/mean.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export function mean(...args: ReadonlyArray<unknown>): unknown {
+  return purry(meanImplementation, args);
+}
+
+function meanImplementation<
+  T extends IterableContainer<bigint> | IterableContainer<number>,
+>(data: T): T[number] | undefined {
+  if (data.length === 0) {
+    return;
```
cjquines, line 79: **nit:** let's use `return undefined`

i thought this would be caught by `@typescript-eslint/consistent-return` but apparently we don't have that rule on...

src/mean.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Returns the mean of the elements of an array.
+ *
+ * Works for both `number` and `bigint` arrays, but not arrays that contain both
+ * types.
+ *
+ * IMPORTANT: The result for empty arrays would be `undefined`, regardless of
+ * the type of the array. This approach improves type-checking and ensures that
+ * cases where `NaN` might occur are handled properly. To avoid adding this to
+ * the return type for cases where the array is known to be non-empty you can use
+ * `hasAtLeast` or `isEmpty` to guard against this case.
```
cjquines, line 28: **question:** what do you think about adding a type test for something like

```ts
const data = [] as Array<number>;
if (hasAtLeast(data, 1)) {
  expectTypeOf(mean(data)) // etc
}
```

cjquines (changes requested):

src/mean.ts
grigorischristainas: You are right, thanks, I missed that. I am going to simplify this type, taking into account the rest of your suggestions.

src/mean.ts
grigorischristainas: Nice idea, that would be very useful!

grigorischristainas:
@cjquines thank you for your feedback! I committed the suggested changes and enhanced tests. For the `Mean` type, I utilized the internal type `NonEmptyArray`, as well as the new `NonEmptyReadonlyArray` type I added, which is the equivalent for readonly arrays. Please let me know if any additional changes are needed.

src/mean.ts
```diff
@@ -0,0 +1,87 @@
...
+} from "./internal/types";
+import { purry } from "./purry";
+import { sum } from "./sum";
+
+type Mean<T extends IterableContainer<bigint> | IterableContainer<number>> =
+  T extends NonEmptyArray<bigint> | NonEmptyReadonlyArray<bigint>
+    ? bigint
+    : T extends NonEmptyArray<number> | NonEmptyReadonlyArray<number>
+      ? number
+      : T extends [] | readonly []
```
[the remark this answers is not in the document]
grigorischristainas: I initially omitted it, but it seems this distinction is required, otherwise type results to `number | undefined` and related type tests break.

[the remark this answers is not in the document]
grigorischristainas: I agree, as the codebase increases in size, it is crucial to maintain consistency. While I haven't come across a lint rule specifically for enforcing consistency in type-level empty array checks, it might be worth exploring a custom one for this purpose?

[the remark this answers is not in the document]
cjquines: hmmmmm okay

PR #880, opened 2024-10-06
chore(RFC): enable word wrapping in vscode

eranhirsch:
Better-comments is part of our recommended extensions. It's great, but at least it's default config doesn't handle multi-line TODOs very well. It doesn't group all the lines after the TODO as part of the same comment so it colors the first line differently than the rest.

We can mitigate this globally for all maintainers (using vscode) by adding some settings to the vscode settings file and rewrite the todo omments to be single line.

This matches the already existing issue where `eslint-disable -- <reason>` comments are already single-line and can't be configured in eslint.

Before:

image

After:

image

It actually makes PRs a little easier to read because these walls of text don't render well on my laptop screen which doesn't have 80 characters for the side-by-side mode.


cjquines (approved):
i don't prefer this style, as i think it's inconsistent with the other jsdocs, but consistency is more important to me

PR #884, opened 2024-10-09
feat(median): Add median

grigorischristainas:
Related to #874. Please let me know if any further adjustments are needed.

---

Make sure that you:

---

### DIFF
--- a/src/median.test-d.ts
+++ b/src/median.test-d.ts
@@ -33,42 +33,10 @@ describe("numbers", () => {
   });
 });
 
-describe("bigints", () => {
-  test("arbitrary arrays", () => {
-    const result = median([] as Array<bigint>);
-    expectTypeOf(result).toEqualTypeOf<bigint | undefined>();
-  });
-
-  test("arbitrary readonly arrays", () => {
-    const result = median([] as ReadonlyArray<bigint>);
-    expectTypeOf(result).toEqualTypeOf<bigint | undefined>();
-  });
-
-  test("arbitrary non-empty arrays", () => {
-    const result = median([1n, 2n] as [bigint, ...Array<bigint>]);
-    expectTypeOf(result).toEqualTypeOf<bigint>();
-  });
-
-  test("consts", () => {
-    const result = median([1n, 2n, 3n] as const);
-    expectTypeOf(result).toEqualTypeOf<bigint>();
-  });
-
-  test("fixed-size tuples", () => {
-    const result = median([1n, 2n] as [bigint, bigint]);
-    expectTypeOf(result).toEqualTypeOf<bigint>();
-  });
-});
-
 describe("dataLast", () => {
   test("numbers", () => {
     const result = pipe([1, 2, 3] as const, median());
     expectTypeOf(result).toEqualTypeOf<number>();
-  });
-
-  test("bigints", () => {
-    const result = pipe([1n, 2n, 3n] as const, median());
-    expectTypeOf(result).toEqualTypeOf<bigint>();
   });
 
   test("arbitrary number arrays", () => {
@@ -80,19 +48,4 @@ describe("dataLast", () => {
     const result = pipe([] as ReadonlyArray<number>, median());
     expectTypeOf(result).toEqualTypeOf<number | undefined>();
   });
-
-  test("arbitrary bigint arrays", () => {
-    const result = pipe([] as Array<bigint>, median());
-    expectTypeOf(result).toEqualTypeOf<bigint | undefined>();
-  });
-
-  test("arbitrary readonly bigint arrays", () => {
-    const result = pipe([] as ReadonlyArray<bigint>, median());
-    expectTypeOf(result).toEqualTypeOf<bigint | undefined>();
-  });
-});
-
-it("doesn't allow mixed arrays", () => {
-  // @ts-expect-error [ts2345] - Can't median bigints and numbers...
-  median([1, 2n]);
 });
--- a/src/median.test-d.ts
+++ b/src/median.test-d.ts
@@ -1,5 +1,3 @@
-import { hasAtLeast } from "./hasAtLeast";
-import { isEmpty } from "./isEmpty";
 import { median } from "./median";
 import { pipe } from "./pipe";
 
@@ -91,22 +89,6 @@ describe("dataLast", () => {
   test("arbitrary readonly bigint arrays", () => {
     const result = pipe([] as ReadonlyArray<bigint>, median());
     expectTypeOf(result).toEqualTypeOf<bigint | undefined>();
-  });
-});
-
-describe("type-guards", () => {
-  it("narrows to `number` using `hasAtLeast`", () => {
-    const data = [] as Array<number>;
-    if (hasAtLeast(data, 1)) {
-      expectTypeOf(median(data)).toEqualTypeOf<number>();
-    }
-  });
-
-  it("narrows to `undefined` using `isEmpty`", () => {
-    const data = [] as Array<number>;
-    if (isEmpty(data)) {
-      expectTypeOf(median(data)).toEqualTypeOf<undefined>();
-    }
   });
 });
 
--- a/src/median.test-d.ts
+++ b/src/median.test-d.ts
@@ -38,12 +38,12 @@ describe("numbers", () => {
 describe("bigints", () => {
   test("arbitrary arrays", () => {
     const result = median([] as Array<bigint>);
-    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+    expectTypeOf(result).toEqualTypeOf<bigint | undefined>();
   });
 
   test("arbitrary readonly arrays", () => {
     const result = median([] as ReadonlyArray<bigint>);
-    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+    expectTypeOf(result).toEqualTypeOf<bigint | undefined>();
   });
 
   test("arbitrary non-empty arrays", () => {
@@ -85,12 +85,12 @@ describe("dataLast", () => {
 
   test("arbitrary bigint arrays", () => {
     const result = pipe([] as Array<bigint>, median());
-    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+    expectTypeOf(result).toEqualTypeOf<bigint | undefined>();
   });
 
   test("arbitrary readonly bigint arrays", () => {
     const result = pipe([] as ReadonlyArray<bigint>, median());
-    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+    expectTypeOf(result).toEqualTypeOf<bigint | undefined>();
   });
 });
 
--- a/src/median.test-d.ts
+++ b/src/median.test-d.ts
@@ -1,0 +1,116 @@
+import { hasAtLeast } from "./hasAtLeast";
+import { isEmpty } from "./isEmpty";
+import { median } from "./median";
+import { pipe } from "./pipe";
+
+test("empty arrays", () => {
+  const result = median([] as const);
+  expectTypeOf(result).toEqualTypeOf<undefined>();
+});
+
+describe("numbers", () => {
+  test("arbitrary arrays", () => {
+    const result = median([] as Array<number>);
+    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+  });
+
+  test("arbitrary readonly arrays", () => {
+    const result = median([] as ReadonlyArray<number>);
+    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+  });
+
+  test("arbitrary non-empty arrays", () => {
+    const result = median([1, 2] as [number, ...Array<number>]);
+    expectTypeOf(result).toEqualTypeOf<number>();
+  });
+
+  test("consts", () => {
+    const result = median([1, 2, 3] as const);
+    expectTypeOf(result).toEqualTypeOf<number>();
+  });
+
+  test("fixed-size tuples", () => {
+    const result = median([1, 2] as [number, number]);
+    expectTypeOf(result).toEqualTypeOf<number>();
+  });
+});
+
+describe("bigints", () => {
+  test("arbitrary arrays", () => {
+    const result = median([] as Array<bigint>);
+    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+  });
+
+  test("arbitrary readonly arrays", () => {
+    const result = median([] as ReadonlyArray<bigint>);
+    expectTypeOf(result).toEqualTypeOf<number | undefined>();
+  });
+
+  test("arbitrary non-empty arrays", () => {
+    const result = median([1n, 2n] as [bigint, ...Array<bigint>]);
+    expectTypeOf(result).toEqualTypeOf<bigint>();
+  });
+
+  test("consts", () => {
+    const result = median([1n, 2n, 3n] as const);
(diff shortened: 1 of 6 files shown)


src/median.ts
```diff
@@ -0,0 +1,103 @@
...
+
+function medianImplementation<
+  T extends IterableContainer<bigint> | IterableContainer<number>,
+>(data: T): T[number] | undefined {
+  if (data.length === 0) {
+    return undefined;
+  }
+
+  // Sort taking into account bigint values
+  const sortedData = sort(data, (a, b) => (a > b ? 1 : a < b ? -1 : 0));
```
cjquines, line 86: **suggestion:** import is unnecessary here; let's use the built-in sort

src/median.ts
```diff
@@ -0,0 +1,103 @@
...
+import { sort } from "./sort";
+
+type Median<T extends IterableContainer<bigint> | IterableContainer<number>> =
+  T extends NonEmptyArray<bigint> | NonEmptyReadonlyArray<bigint>
+    ? bigint
+    : T extends NonEmptyArray<number> | NonEmptyReadonlyArray<number>
+      ? number
+      : T extends [] | readonly []
+        ? undefined
+        : number | undefined;
```
cjquines, line 19: **question:** actually, shouldn't this depend on whether `T` is an array of `bigint`s or `number`s? and same with `mean`?

cjquines (commented):
sorry i only just got to this

src/median.ts
```diff
@@ -0,0 +1,103 @@
...
+  // For odd length, return the middle element
+  if (sortedData.length % 2 !== 0) {
+    return sortedData[middleIndex];
+  }
+
+  // For even length, return the mean of the two middle elements
+  const middleElements = [sortedData[middleIndex - 1], sortedData[middleIndex]];
+
+  if (middleElements.every(isBigInt) || middleElements.every(isNumber)) {
+    return mean(middleElements);
```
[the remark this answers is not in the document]
grigorischristainas: I think there are some issues that have already been handled in `mean`, that's why I used it. Would you suggest something similar to the following as an alternative?
```
// For even length, return the mean of the two middle elements
const middleElements = [sortedData[middleIndex - 1], sortedData[middleIndex]] as Array<number> | Array<bigint>;

const dataSum = sum(middleElements);

 return typeof dataSum === "bigint"
    // eslint-disable-next-line @typescript-eslint/no-magic-numbers
    ? dataSum / 2n
    : dataSum / 2;
```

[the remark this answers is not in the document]
cjquines: i'm not sure i see why we have to import `sum` or `mean` here? we can just add the two elements

src/median.ts
grigorischristainas: I think you are right, would you suggest something like the following: 
```
T extends [] | readonly []
    ? undefined
    : T[number] extends bigint
         ? bigint | undefined
         : number | undefined
```
and the related type tests for bigints should be modified as follows:
```
test("arbitrary bigint arrays", () => {
    const result = pipe([] as Array<bigint>, median());
    expectTypeOf(result).toEqualTypeOf<bigint | undefined>(); // instead of <number | undefined>
});
```

cjquines: i think so! and rmb to do the same with `mean`

grigorischristainas: Thank you, should I open a new PR to integrate this change into `mean` ?

src/median.ts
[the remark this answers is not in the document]
grigorischristainas: I used `mean` (and `sum` in this example) mainly to avoid repeating the logic of handling bigint and number separately, as the two elements `sortedData[middleIndex - 1]` and `sortedData[middleIndex]` can be either `number` or `bigint`. 

If we try to add them directly, we will get a type error indicating that `operator '+' cannot be applied to types 'number | bigint' and 'number | bigint'`. This is why I used `sum` in this example (and `mean` in the original implementation), as it handles this logic internally. 

In order to add the two numbers directly in this case without repeating the sum logic, I think we would need to use a conditional.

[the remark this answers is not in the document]
grigorischristainas: I added a related commit without using `mean` or `sum` :)

eranhirsch (changes requested):
This uses a naive algorithm to compute the median which is based on first sorting all the elements in the array (O(nlogn)). [QuickSelect](https://en.wikipedia.org/wiki/Quickselect) allows us to find the median at O(n) time. We already have it implemented in our library as a util, because it is used in nthBy.

grigorischristainas:
@eranhirsch that's cool I didn't know QuickSelect, thanks! Your suggestion about `nthBy` would work fine for odd length arrays. However, for even length arrays we need the average of the two middle elements. Please correct me if I am mistaken but in that case it would look something like this: 
```
const middleIndex = Math.floor(data.length / 2)
const median =
      (nthBy(data, middleIndex, identity()) +
        nthBy(data, middleIndex - 1, identity())) /
      2;
```
In any case, if you think that the `median` is not necessary, we can close this PR.

eranhirsch:
I trust @cjquines's opinion on this as he has a better intuiton than me on how our users will see this

cjquines:
i do think median of even-length arrays is an annoying enough edge case that i want a median function

eranhirsch:
Yesterday, after posting my comments, I was curious to learn more about Quickselect and how different libraries implement their median utility function. I learned that the built-in sort is actually extremely efficient (this is the case for v8, Safari, but not Firefox!, and also Python and Java, amongst others), so the vast majority of libraries implement median by sorting and not by using Quickselect.

With median requiring either an **adapted** quickselect algorithm to find two items (N/2 and N/2+1 when N%2 === 0) to prevent running quickselect twice, your current implementation is more efficient for most practical use cases, and there's no need to change it.

eranhirsch:
https://en.wikipedia.org/wiki/Timsort

src/median.ts
```diff
@@ -0,0 +1,101 @@
...
+    return sortedData[middleIndex];
+  }
+
+  // For even length, return the mean of the two middle elements
+  const firstMiddle = sortedData[middleIndex];
+  const secondMiddle = sortedData[middleIndex - 1];
+
+  return typeof firstMiddle === "bigint" && typeof secondMiddle === "bigint"
+    ? // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+      firstMiddle + secondMiddle / 2n
```
eranhirsch, line 99: Are we OK with median on bigints also rounding down the result when the array has an odd length? this implementation would cause `median([1n, 2n])` to return `1n`, but `median([1, 2])` would return `1.5`.

I'm not sure this is the right thing to do, but if it is, we need to mention it in the docs so people are aware; not all our users would notice this or think about it themselves. We also need to have a test that surfaces this edge case.

src/median.ts
```diff
@@ -0,0 +1,101 @@
...
+
+function medianImplementation<
+  T extends IterableContainer<bigint> | IterableContainer<number>,
+>(data: T): T[number] | undefined {
+  if (data.length === 0) {
+    return undefined;
+  }
+
+  // Sort taking into account bigint values
+  const sortedData = [...data].sort((a, b) => (a > b ? 1 : a < b ? -1 : 0));
```
eranhirsch, line 84: This arrow function isn't dependent on any closures so it should be pulled out so that the engine doesn't need to constantly recreate it if median is called often.

```suggestion
  const sortedData = [...data].sort(numberAndBigIntComparator);
```

```ts
function numberAndBigIntComparator(a: number | bigint, b: number | bigint): number {
  return a > b ? 1 : a < b ? -1 : 0;
}
```

eranhirsch, line 84: Can you add a TODO here that we should replace `sort` with `toSorted` once we drop Node 18 support next year like this: https://github.com/remeda/remeda/blob/main/src/sortBy.ts#L102

src/median.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const middleIndex = Math.floor(sortedData.length / 2);
```
eranhirsch, line 86: This is a bit wasteful because we can *first* check if the array length is odd or even. If the length is odd, we don't need the `Math.floor`: we can compute it using `array.length - 1 / 2`. If the length is even we don't need the floor at all.

src/median.ts
```diff
@@ -0,0 +1,101 @@
...
+  // For odd length, return the middle element
+  if (sortedData.length % 2 !== 0) {
+    return sortedData[middleIndex];
+  }
+
+  // For even length, return the mean of the two middle elements
+  const firstMiddle = sortedData[middleIndex];
+  const secondMiddle = sortedData[middleIndex - 1];
+
+  return typeof firstMiddle === "bigint" && typeof secondMiddle === "bigint"
```
eranhirsch, line 97: This adds an additional runtime check for the type of secondMiddle, although **we** know it would match the type of firstMiddle; in these cases, it's better to suppress the typescript error.

src/median.test-d.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+  });
+
+  it("narrows to `undefined` using `isEmpty`", () => {
+    const data = [] as Array<number>;
+    if (isEmpty(data)) {
+      expectTypeOf(median(data)).toEqualTypeOf<undefined>();
+    }
+  });
+});
```
eranhirsch, line 111: I don't feel we need these; they test has at `hasAtLeast` and `isEmpty` more than they test median.

eranhirsch (commented):
My biggest concern right now is the return value for bigint arrays of even length. Currently we round down so that the result remains a bigint (`median([1n, 2n]) //=> 1n`). We can return a `number` instead, but that would mean that `median([1_000_000_000_000_000_000_000n])` would overflow.

I don't know what the right answer is here...

grigorischristainas:
> My biggest concern right now is the return value for bigint arrays of even length. Currently we round down so that the result remains a bigint (`median([1n, 2n]) //=> 1n`). We can return a `number` instead, but that would mean that `median([1_000_000_000_000_000_000_000n])` would overflow.
> 
> I don't know what the right answer is here...

I agree this is a valid concern and we should mention it in the docs. We could also add a note about the limitations of converting large bigint values to number due to overflow risks. Also, a link to [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#operators) would be helpful for users who want to understand the underlying behavior better.

src/median.ts
cjquines: agreed we should call out this case specifically. we can also punt on this decision by not supporting `bigint`

grigorischristainas: How would you suggest to eventually proceed? Should we drop support for `bigint` or mention these issues in the docs? 

We might need to do the same in `mean` too.

eranhirsch:
Just remember that timsort is optimized for real world cases and might perform worse with random data.

grigorischristainas:
@cjquines @eranhirsch I integrated your suggestions from your previous review comments and also enhanced docs about `bigint` limitations. Could you please let me know if and how we will proceed with the `timsort` implementation?

eranhirsch:
I think we should use sort because it's the more standard approach with less moving parts on our side to maintain, and i think we should drop bigint support as function has a well defined mathematical meaning that doesn't include truncating the results so might lead to misunderstands and errors. Documenting it isn't enough to protect people from making mistakes.

grigorischristainas:
I agree, we should probably do the same for `mean`. I can open a separate PR for that.

src/median.ts
```diff
@@ -0,0 +1,86 @@
...
+  // TODO [2025-05-01]: When node 18 reaches end-of-life bump target lib to ES2023+ and use `Array.prototype.toSorted` here.
+  const sortedData = [...data].sort(numberComparator);
+
+  // For odd length, return the middle element
+  if (sortedData.length % 2 !== 0) {
+    return sortedData[(sortedData.length - 1) / 2];
+  }
+
+  // For even length, return the mean of the two middle elements
+  const middleIndex = Math.floor(sortedData.length / 2);
```
eranhirsch, line 83: You don't need `Math.floor`; the number is even and divided cleanly by 2.

```suggestion
  const middleIndex = sortedData.length / 2;
```

src/median.ts
```diff
@@ -0,0 +1,86 @@
...
+} from "./internal/types";
+import { purry } from "./purry";
+
+type Median<T extends IterableContainer<number>> = T extends
+  | NonEmptyArray<number>
+  | NonEmptyReadonlyArray<number>
+  ? number
+  : T extends [] | readonly []
+    ? undefined
+    : number | undefined;
```
eranhirsch, line 14: I'm not a big fan of the NonEmptyArray usage because it obfuscates the logic behind another reference hop, making local modifications harder when needed. The code is even shorter without it. Additionally, instead of repeating the different components (undefined, number) several times, we can be more concise by conditionally adding each one exactly one.

```suggestion
type Median<T extends IterableContainer<number>> =
  | (T extends readonly [] ? never : number)
  | (T extends readonly [unknown, ...Array<unknown>] ? never : undefined);
```

This reads: "If the array is empty then the result can never be number, and if the array is non-empty the result can never be undefined.

src/median.test.ts
```diff
@@ -0,0 +1,26 @@
+import { median } from "./median";
+import { pipe } from "./pipe";
+
+describe("dataFirst", () => {
+  it("should return the median of numbers in an array", () => {
+    expect(median([6, 10, 11])).toBe(10);
+    expect(median([1, 2, 3, 9])).toBe(2.5);
+    expect(median([-1, 0, 1, 10])).toBe(0.5);
+  });
```
eranhirsch, line 9: Because the logic is different for arrays of even length and arrays of odd length, let's split the tests so that each branch is tested separately. This would make the tests more precise in signaling what broke.
```suggestion
  test("arrays of odd length", () => {
    expect(median([6, 10, 11])).toBe(10);
  });
  
  test("arrays of even length", () => {
    expect(median([1, 2, 3, 9])).toBe(2.5);
    expect(median([-1, 0, 1, 10])).toBe(0.5);
  });
```

src/median.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  it("should return undefined for an empty array", () => {
+    expect(median([])).toBeUndefined();
+  });
+});
+
+describe("dataLast", () => {
+  it("should return the median of numbers in an array", () => {
```
eranhirsch, line 17: Simlarly.

src/median.test-d.ts
```diff
@@ -0,0 +1,51 @@
+import { median } from "./median";
+import { pipe } from "./pipe";
+
+test("empty arrays", () => {
+  const result = median([] as const);
+  expectTypeOf(result).toEqualTypeOf<undefined>();
+});
+
+describe("numbers", () => {
```
eranhirsch, line 9: We don't need this describe anymore.
```suggestion
```

eranhirsch (dismissed):
Looks good to me! Thanks for being patient with my comments!

I have some nits inline that I think we should fix, and I still want to see how @cjquines thinks about our decisions to drop bigint support and on the usage of sort, before we ship this.

src/median.ts
grigorischristainas: Sorry, I missed that, thanks!

src/median.ts
grigorischristainas: That seems better, thanks! I used these types initially to cover all the cases explicitly, but I can see how your approach simplifies the logic.

cjquines (approved):
lgtm

PR #889, opened 2024-10-15
test(splice): split each block to individual tests

eranhirsch:
Replace the test.each block with concrete describe and tests with proper names.


cjquines (approved):
thanks

PR #890, opened 2024-10-15
fix(isEmpty): support interfaces

### LINKED ISSUE #879: `isEmpty` type definition don't work correctly with interfaces
I'm really struggling with `isEmpty`.

I think the current implementation is seriously lacking from a Typescript perspective. Either that, or i'm not understanding it clearly (then might the docs be the problem).

The docs state this very basic example:

```
R.isEmpty({}); //=> true
```

Which is fine. and works as expected, but let's just assign any kind of interface to it:
```
interface MyObject {
  stuff1?: string;
  stuff2?: number;
}

const x: MyObject = {};

R.isEmpty(x);
```

Suddenly i'm met with a Typescript error:

```
No overload matches this call.
  Overload 1 of 3, '(data: string | undefined): data is "" | undefined', gave the following error.
    Argument of type 'MyObject' is not assignable to parameter of type 'string'.
  Overload 2 of 3, '(data: IterableContainer): data is []', gave the following error.
    Argument of type 'MyObject' is not assignable to parameter of type 'IterableContainer'.
  Overload 3 of 3, '(data: Readonly<Record<PropertyKey, unknown>>): data is Record<PropertyKey, never>', gave the following error.
    Argument of type 'MyObject' is not assignable to parameter of type 'Readonly<Record<PropertyKey, unknown>>'.
      Index signature for type 'string' is missing in type 'MyObject'.ts(2769)
```

I have no idea what any of that TS gibberish means, but clearly this isn't expected behavior, right? I'm still just passing a simple empty object, why is this failing at all?

eranhirsch:
Support interface types.

Closes #879

### DIFF
--- a/src/isEmpty.test-d.ts
+++ b/src/isEmpty.test-d.ts
@@ -56,6 +56,19 @@ describe("objects", () => {
       expectTypeOf(data).toEqualTypeOf<Partial<Record<"a" | "b", number>>>();
     }
   });
+
+  test("interfaces", () => {
+    // eslint-disable-next-line @typescript-eslint/consistent-type-definitions -- Intentional
+    interface MyInterface {
+      a: number;
+    }
+    const data = { a: 123 } as MyInterface;
+    if (isEmpty(data)) {
+      expectTypeOf(data).toEqualTypeOf<Record<"a", never>>();
+    } else {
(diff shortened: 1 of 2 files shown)


cjquines (approved):
🎉

PR #898, opened 2024-10-27
refactor(mean): drop `bigint` support

grigorischristainas:
Related to #874, I have also integrated changes based on recent comments about `median` (e.g., removed type-guard related tests and modified the `Mean` type)

Please let me know if any additional adjustments are needed.

### DIFF
--- a/src/internal/types.ts
+++ b/src/internal/types.ts
@@ -45,8 +45,6 @@ export type BrandedReturn<F extends (...args: any) => any> = (
 
 export type NonEmptyArray<T> = [T, ...Array<T>];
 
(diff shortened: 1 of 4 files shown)


cjquines (approved):

cjquines:
technically a breaking change, but i'm fine shipping this as `fix`

eranhirsch:
This didn't cause a release because we only release on "fix" and "feat"... we should drop the "refactor" label as it isn't helpful.

cjquines:
i think `refactor` and `chore` shouldn't trigger a release, but yeah

PR #909, opened 2024-11-03
feat: add `toKebabCase` and `toSnakeCase`

eranhirsch:
By popular demand, adding simplified kebab case and snake case functions.

The return types are trivial (`string`) because type-fest doesn't export `SplitWords` and because its own KebabCase and SnakeCase types create unions of possibilities instead of a concrete single outcome. #818 tries to deal with that issue more concretely and can replace this implementation once the changes in type-fest are approved.

I'm still deliberating with myself whether these functions need to be part of the library or that we should export `splitWords` instead, as they are simple compositions on top of it (with `join` and `toLowerCase`). My issue with split words is that it isn't very smart; it's a very mechanical way to define what a word is and not very linguistic. By exporting it, we take responsibility for fixing all sorts of issues people have with it when they use it for linguistic purposes. On the other hand, the utilities that use it (snake_case, camelCase, kebab-case) are mechanical themselves, and I don't think people will be as surprised if they don't make sense linguistically.


eranhirsch:
I didn't notice that SplitWords isn't exported by type-fest :(

eranhirsch:
Better types waiting for https://github.com/sindresorhus/type-fest/pull/975, but we shouldn't wait; we will release a fix later if they approve it.

cjquines (approved):

filipw01:
Nice, happy to see it merged. I finally found time to finish the type-fest PR so I'm hoping it will be merged soon and we'll have better types

PR #910, opened 2024-11-03
fix(mergeDeep): Support interfaces

### LINKED ISSUE #522: `mergeDeep` does not allow interface types
`mergeDeep` does not allow interface types. The following

```
interface Example {
  b: string
}

const a: Example = { b: 'c' }

R.mergeDeep(a, { d: 'e' })
```
fails with
```
Argument of type 'Example' is not assignable to parameter of type 'Record<string, unknown>'.
  Index signature for type 'string' is missing in type 'Example'.
```

eranhirsch:
Fixes #522.

We don't have a lot of type tests so I don't feel very confident about this change, but I can't think of a specific thing I expect to break with this change. The only reason record was used was because that was easy to understand, as long as the runtime doesn't complain about the lack of data coming from `object` (vs `Record`) than it's unlikely the types themselves break.

And we rely on type-fest for the heavy lifting; they probably handle interfaces well...

### DIFF
--- a/src/mergeDeep.test-d.ts
+++ b/src/mergeDeep.test-d.ts
@@ -1,49 +1,86 @@
 import { mergeDeep } from "./mergeDeep";
 
 it("trivially merges disjoint objects", () => {
-  const a = { foo: "bar" };
-  const b = { bar: "baz" };
-  const result = mergeDeep(a, b);
+  const result = mergeDeep(
+    { foo: "bar" } as { foo: string },
+    { bar: "baz" } as { bar: string },
+  );
   expectTypeOf(result).toEqualTypeOf<{ foo: string; bar: string }>();
 });
 
 it("merges fully overlapping types", () => {
-  const a = { foo: "bar" };
-  const b = { foo: "baz" };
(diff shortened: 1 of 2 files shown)


src/mergeDeep.test-d.ts
```diff
@@ -1,49 +1,86 @@
...
+      foo: Array<{ bar: string; y: number }>;
+    },
+  );
+  expectTypeOf(result).toEqualTypeOf<{
+    foo: Array<{ bar: string; y: number }>;
+  }>();
+});
+
+it("works with interfaces", () => {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-definitions -- Intentional
```
cjquines, line 69: **nit:** we can probably lift this to the file level :shrug:

cjquines (approved):

src/mergeDeep.test-d.ts
eranhirsch: I usually do that once the lint suppression is relevant to more than one test. As long as its isolated I prefer keeping it local.

PR #915, opened 2024-11-06
chore: split internal/types.ts

eranhirsch:
The idea of splitting the file came up recently in a PR. I think it makes sense, as the file has grown bigger and is harder to maintain.

This also makes testing easier and makes gaps in our type tests more apparent.


cjquines (commented):
**question:** do you have thoughts about whether the type files should be `.d.ts`?

this could plausibly be enforced via lint rule, and makes sure that the files in that folder are all type files

eranhirsch:
From my understanding, this is not the intention of .d.ts files. .d.ts files affect the global namespace and are used to tell TS something about the world that it can't otherwise know about. For regular code that could be imported they shouldn't be used. Also, not using them doesn't impact negatively anything about the built package because the compiler should ignore type imports entirely.

https://www.totaltypescript.com/books/total-typescript-essentials/modules-scripts-and-declaration-files#declaration-files

Comparing to type-fest, I think the main difference is that we also have runtime components that do get compiled, whereas type-fest is strictly types-only.

cjquines:
that's fine. another question:
* what do you feel about doing either a barrel file for the types? since it's types-only, the usual concerns about barrel files wouldn't be too big of an issue
* alternatively, what do you think about separating it to a different package? we already have a somewhat monorepo setup with docs and src, we could continue down that path... but idk
(anyway, none of these questions are blocking, so i've approved)

cjquines (approved):

cjquines:
oh, i agree we don't want to support the types externally; by "package" i meant "internal package".

i do think moving to a monorepo might be a good move in the long run if we're going to host our own codemods...

PR #917, opened 2024-11-08
(fix) Added keys and values to export type

brianhuang822:
Added key and value to export. When upgrading to TS 5.6, the typechecker complains this type cannot be inferenced

```
 error TS2742: The inferred type of 'typedKeys' cannot be named without a reference to '../../../../node_modules/remeda/dist/types-B6Nc-8hG'. This is likely not portable. A type annotation is necessary.
 ```

yarn patching it to export fixed the problem, pushing the fix upstream

### DIFF
--- a/src/keys.ts
+++ b/src/keys.ts
@@ -4,7 +4,7 @@ import type {
 } from "./internal/types";
 import { purry } from "./purry";
 
-type Keys<T> = T extends IterableContainer ? ArrayKeys<T> : ObjectKeys<T>;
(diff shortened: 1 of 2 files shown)


eranhirsch (changes requested):
Can you provide a more complete repro we can check out? You might be doing something wrong or are relying on an internal of the library.

As a rule of thumb, we don't export types at all and don't consider them part of the API we offer.

brianhuang822:
We were able refactor our code to not use the internal types

PR #920, opened 2024-11-14
fix(prop): Support headless `prop` without explicit type params

Masyaka:

---

Make sure that you:

---

### Available Types:

existing).

</details>


eranhirsch (changes requested):
We usually don't cater in the library to the use-cases where the function returned from our curried data-last variant is stored and then used as a stand-alone function, e.g.:

```ts
const foo = prop("a");
foo(DATA);
```

We don't consider passing in explicit type parameters as part of our API. That is, if a function doesn't work without any type parameters passed in, then it's either a type bug or not supported by the library.

BUT... in this case, it looks like this works fine and doesn't break anything, so I'm OK with adding this extra signature to the function. This also comes up a lot in issues, so many people would be happy with it. I ALSO managed to create a type test that shows that breaks without this change, which brings me to my main point:

Please add tests to your change. The tests should be built so that they would break before adding your change and would be green after adding it. Please add enough tests to show that this doesn't break any of the existing usages of the utility function. I'm aware that this function isn't tested well to begin with, so it would be really helpful if you could add enough tests to make our testing more robust for it.

eranhirsch:
@Masyaka , are you planning on continuing to work on this?

Masyaka:
> @Masyaka , are you planning on continuing to work on this?

Yes, will finish it at this week

Masyaka:
Why actually I decided to add those changes, there is the jsdoc with signature:
image
This signature was not working, before this PR

eranhirsch (approved):
This is great, thanks for helping us out. I'm sure a lot of people would find this useful. Please check out my nits about the tests, and I will merge this once you send a fix.

src/prop.test-d.ts
```diff
@@ -11,3 +11,21 @@ test("prop typing", () => {
...
+  const valuesA = input.map(standAlonePropA);
+  expectTypeOf(valuesA).toEqualTypeOf<Array<number>>();
+});
+
+test("prop expect error", () => {
+  const standAlonePropB = prop("b");
+  const item = { a: 1 };
+  // @ts-expect-error - b is not a key of typeof item
+  const valueB = standAlonePropB(item);
+  expectTypeOf(valueB).toEqualTypeOf<unknown>();
```
eranhirsch, line 30:
```suggestion
```

eranhirsch (approved):

src/prop.test-d.ts
```diff
@@ -11,3 +11,20 @@ test("prop typing", () => {
...
+  expectTypeOf(valueA).toEqualTypeOf<number>();
+  const valuesA = input.map(standAlonePropA);
+  expectTypeOf(valuesA).toEqualTypeOf<Array<number>>();
+});
+
+test("prop expect error", () => {
+  const standAlonePropB = prop("b");
+  const item = { a: 1 };
+  // @ts-expect-error - b is not a key of typeof item
+  const valueB = standAlonePropB(item);
```
eranhirsch, line 29:
```suggestion
  // @ts-expect-error [ts2322] -- b is not a key of typeof item
  standAlonePropB({ a: 1 });
```

Issue #911, opened 2024-11-05
`pick` typing doesn't treat the provided keys as an array


eranhirsch:
`pick` assumes that the provided array of keys includes all values in it's type, this causes it to return the wrong type. We need to remap the result and build the output similar to how it's done in pickBy.

example: https://www.typescriptlang.org/play/?noUncheckedIndexedAccess=true&noUnusedParameters=true&strictBuiltinIteratorReturn=true&useUnknownInCatchVariables=true&exactOptionalPropertyTypes=true&noFallthroughCasesInSwitch=true&noImplicitOverride=true&noPropertyAccessFromIndexSignature=true&noUnusedLocals=true#code/JYWwDg9gTgLgBAbzgawKYE8DOAaOZgDGycAvnAGZQQhwBEUqIqAJgIa0DcAUKJLInACSmAKIBHAK6sANqQpUatGOjCoAtOVSYYnLl2Wq4ADTgBeRFzhxWAfgBccbVGAA7AObcrAI3twXEkC9UKG4Sbi5mVAJpVgY4AggXbTg2GFYHI24EpPgIGAALYLNEMlZMY3Ds5IZMCWl4c3wiAApU1lw0LGa8wqgASj7uAHohqysAPRs9A1Q4ACUtOoahUUkZAB4ZiHI4GqXcIwA+YdGJqaA

```ts
import { keys, pick } from "remeda";
import { IsEqual } from "type-fest";

type X = {
  a?: string;
  b?: number;
};

declare const data: X;
const other = {} as X;

const result = pick(data, keys(other));
//    ^? const result: Pick<X, keyof X>

type Result = IsEqual<typeof result, X>;
//   ^? type Result = true
```

PR #923, opened 2024-11-15
fix(pick): support optional keys

eranhirsch:
Expanding our TupleParts utility to allow easy access to the optional part of the prefix, and then using this to improve the type of pick so that optional keys are marked as optional in the output.

Closes: #911

### DIFF
--- a/src/internal/types/TupleParts.ts
+++ b/src/internal/types/TupleParts.ts
@@ -8,9 +8,9 @@ import type { IsEqual } from "type-fest";
  *
  * NOTE: The prefix is split into 2 tuples where all items are non-optional so
  * that types that rely on the presence of a specific element can be built more
- * accurately. For backwards compatibility, the `prefix` accessor which hides
- * some of this complexity is also provided (but we might deprecate it in the
- * future).
+ * accurately.
+ *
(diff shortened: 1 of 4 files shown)


src/internal/types/TupleParts.ts
```diff
@@ -2,37 +2,53 @@ import type { IsEqual } from "type-fest";
...
- * the rest parameter (if any), and the item is the type of the rest parameter.
+ * anything before the rest parameter (if any), split between it's required
+ * part, and it's optional part. The suffix is anything after the rest
+ * parameter (if any), and the item is the type of the rest parameter.
+ *
+ * NOTE: The prefix is split into 2 tuples where all items are non-optional so
+ * that types that rely on the presence of a specific element can be built more
+ * accurately. For backwards compatibility, the `prefix` accessor which hides
+ * some of this complexity is also provided (but we might deprecate it in the
+ * future).
```
cjquines, line 13: **thought:** i think we should just deprecate it, if it's not hard to

cjquines (dismissed):

src/internal/types/TupleParts.ts
eranhirsch: It's more work than made sense for this PR; it requires looking into each case and seeing how the optionality changes it. When I built the type for chunk I made an assumption that the tuples don't have optional members.

I'll replace this with a TODO so we know this needs to happen.

PR #927, opened 2024-11-22
chore(ts): Upgrade typescript to 5.7 and add it to the testing matrix

eranhirsch:


eranhirsch:
@cjquines - I couldn't figure out what the issue is with the failing type tests. I checked if updating vitest helps but it was only a patch version bump and it didn't help. Do you have any idea what might be the issue here?

eranhirsch:
nvm - https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#typedarrays-are-now-generic-over-arraybufferlike

cjquines:
hm, but now it's failing the lower version typetests…

eranhirsch:
Fixed as part of #939, checkout https://github.com/remeda/remeda/pull/939/files#r1875847152

PR #929, opened 2024-11-25
fix(toCamelCase, toSnakeCase, toKebabCase): Support iOS 16.3-

### LINKED ISSUE #928: SyntaxError: Invalid regular expression: invalid group specifier name
https://caniuse.com/js-regexp-lookbehind is not supported in 5.62% of devices as of November 2024, and it seems there is not a polyfill for this. Please consider using an alternative implemention instead of look-behinds, as by using this library, we lose support for iOS 16.3 and below.

https://github.com/remeda/remeda/blob/b14511ae8d501df7c41968e5eacb69185faf5e9e/src/internal/words.ts#L38-L54

ariesclark:
Closes #928.

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/internal/words.ts
+++ b/src/internal/words.ts
@@ -31,30 +31,68 @@ const WHITESPACE = [
 ] as const;
 
 // @see https://github.com/sindresorhus/type-fest/blob/main/source/internal/characters.d.ts#L33
-const WORD_SEPARATORS = ["-", "_", ...WHITESPACE] as const;
+const WORD_SEPARATORS = new Set(["-", "_", ...WHITESPACE]);
 
-// We are basing our logic on the type definition of SplitWords from type-fest.
-// @see https://github.com/sindresorhus/type-fest/blob/main/source/split-words.d.ts
-const WORD_SPLITTING_RE = new RegExp(
-  [
-    // **on** any word separator, these would be removed from the output.
-    `[${WORD_SEPARATORS.join("")}]+`,
-    // When the text transitions from a non-digit to a digit.
-    String.raw`(?<=\D)(?=\d)`,
-    // When the text transitions from a digit to a non-digit.
-    String.raw`(?<=\d)(?=\D)`,
-    // When the text transitions from a lower case to upper case.
-    String.raw`(?<=[a-z])(?=[A-Z])`,
-    // When the text transitions from 2 upper case letters to a lower case
-    // letter. (one upper case letter is considered part of the word, e.g.
-    // "Dog").
-    String.raw`(?<=[A-Z])(?=[A-Z][a-z])`,
-  ].join("|"),
-  "u",
-);
+function splitWords(data: string): Array<string> {
+  const words: Array<string> = [];
+  let word = "";
+
+  const flush = (): void => {
+    if (word.length > 0) {
+      words.push(word);
+      word = "";
+    }
+  };
+
+  for (const character of data) {
+    if (WORD_SEPARATORS.has(character)) {
+      // Separator encountered; flush the current word.
+      flush();
+      continue;
+    }
+
+    // Detect transitions:
+    // 1. Lowercase to uppercase (e.g., "helloWorld")
+    if (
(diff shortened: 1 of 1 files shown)


src/internal/words.ts
```diff
@@ -31,30 +31,68 @@ const WHITESPACE = [
...
+    ) {
+      const lastCharacter = word.slice(-1);
+      word = word.slice(0, -1);
+      flush();
+      word = lastCharacter;
+    }
+    // 3. Digit to non-digit or non-digit to digit (e.g., "123abc" or "abc123")
+    else if (
+      (/\d/u.test(word.at(-1)!) && /\D/u.test(character)) ||
+      (/\D/u.test(word.at(-1)!) && /\d/u.test(character))
```
eranhirsch, line 79: Can't these also just be:

```suggestion
      (/\d$/u.test(word) && /\D/u.test(character)) ||
      (/\D$/u.test(word) && /\d/u.test(character))
```

eranhirsch (changes requested):
I'm not that concerned about performance, but there are a few places where it's easy to improve. In general it would probably be the most efficient if we implement this as a state machine and keep a few flags for what the previous character was, instead of looking back. If we are ditching the RegExp approach there's no point not optimizing the implementation.

src/internal/words.ts
```diff
@@ -31,30 +31,55 @@ const WHITESPACE = [
...
+  for (const character of data) {
+    if (WORD_SEPARATORS.has(character)) {
+      // Separator encountered; flush the current word.
+      flush();
+      continue;
+    }
+
+    // Detect transitions:
+    // 1. Lowercase to uppercase (e.g., "helloWorld")
+    if (word.length > 0 && /[a-z]$/u.test(word) && /[A-Z]/u.test(character)) {
```
eranhirsch, line 58: Why do we need the word length check here?

``` suggestion
    if (/[a-z]$/u.test(word) && /[A-Z]/u.test(character)) {
```

src/internal/words.ts
```diff
@@ -31,30 +31,55 @@ const WHITESPACE = [
...
+      flush();
+    }
+    // 2. Uppercase to lowercase following multiple uppercase letters (e.g., "HELLOWorld")
```
eranhirsch, line 61: The explanation above in the RegExp explains why the two consecutive letters are an important differentiator from the lowercase case. Can you also copy that part?

src/internal/words.ts
```diff
@@ -31,30 +31,55 @@ const WHITESPACE = [
...
+    else if (/[A-Z][A-Z]$/u.test(word) && /[a-z]/u.test(character)) {
+      const lastCharacter = word.slice(-1);
+      word = word.slice(0, -1);
+      flush();
+      word = lastCharacter;
+    }
+    // 3. Digit to non-digit or non-digit to digit (e.g., "123abc" or "abc123")
+    else if (
+      (/\d$/u.test(word) && /\D/u.test(character)) ||
+      (/\D$/u.test(word) && /\d/u.test(character))
```
eranhirsch, line 71: \D and \d are opposing conditions so we know that if one is true the other is false. The way the code is structured right now, we will run the check twice in some cases instead of relying on that fact.

```suggestion
      /\d$/u.test(word) !== /\d/u.test(character)
```

src/internal/words.ts
```diff
@@ -31,30 +31,55 @@ const WHITESPACE = [
...
+  const flush = (): void => {
+    if (word.length > 0) {
+      results.push(word);
+      word = "";
+    }
+  };
+
+  for (const character of data) {
+    if (WORD_SEPARATORS.has(character)) {
+      // Separator encountered; flush the current word.
```
eranhirsch, line 51: The comment above also mentions more clearly that the separator would be removed, can you mention that herew?

eranhirsch (approved):
Thanks for handling this.

eranhirsch (approved):
Thanks!

Issue #885, opened 2024-10-12
Replacement for Lodash `countBy`


cjquines:
Right now `countBy(data, fn)` is the composition `mapValues(groupBy(data, fn), length())`, which is a bit unwieldy.

* Lodash and Ramda have `countBy`.
* Python has `collections.Counter`, which is `countBy(data, identity())`, and is what I use most often by far.
* Hack has `Dict\count_values`, which is also `countBy(data, identity())`.
* The Rust crate [`counter`](https://crates.io/crates/counter) is quite popular (21k weekly downloads), and it is also `countBy(data, identity())`.

eranhirsch:
The composed solution is also less efficient as it requires building the whole array just to get the count so there's a perf advantage too.

Interestingly, the most abstract/generic/general-purpose function that would cover all cases would be a groupBy-adjacent function that accepts a second reducer callback:

```ts
function reduceBy<T, K extends PropertyKey, V>(
  data: ReadonlyArray<T>,
  grouper: (item: T) => K,
  reducer: (accumulator: V, item: T) => V,
  initialValue: V,
): Record<K, V>;
```

cjquines:
i agree that's the most general solution, but that's such a horrifying signature that i'm not sure i want it

PR #937, opened 2024-12-01
feat(countBy): Add countBy

grigorischristainas:
Related to #885. It would be nice to have this utility so I decided to give it a try :)

I would be happy to make any further adjustments as needed.

---

Make sure that you:

### DIFF
--- a/src/countBy.test-d.ts
+++ b/src/countBy.test-d.ts
@@ -29,3 +29,10 @@ test("callback function type", () => {
     return x;
   });
 });
+
+test("literal unions", () => {
+  const data = [] as Array<"cat" | "dog">;
+  const result = countBy(data, (x) => x);
+
+  expectTypeOf(result).toEqualTypeOf<Partial<Record<"cat" | "dog", number>>>();
+});
--- a/src/countBy.test-d.ts
+++ b/src/countBy.test-d.ts
@@ -1,0 +1,35 @@
+import { countBy } from "./countBy";
+import { pipe } from "./pipe";
+
+describe("dataFirst", () => {
+  test("countBy", () => {
+    const data = ["a", "b", "c", "B", "A", "a"];
+    const result = countBy(data, (x) => x.toLowerCase());
+
+    expectTypeOf(result).toEqualTypeOf<Record<string, number>>();
+  });
+});
+
+describe("dataLast", () => {
+  test("countBy", () => {
+    const data = ["a", "b", "c", "B", "A", "a"];
+    const result = pipe(
+      data,
+      countBy((x) => x.toLowerCase()),
+    );
+
+    expectTypeOf(result).toEqualTypeOf<Record<string, number>>();
+  });
+});
+
+test("callback function type does not match data array type", () => {
+  const data = [1, 2, 3];
+  // @ts-expect-error [ts2345]: Type 'number' is not assignable to type 'string'.
+  countBy(data, (x: string) => x);
+});
+
+test("array with nullish values", () => {
+  const data = [1, null, undefined];
+  // @ts-expect-error [ts2322] - Type 'number | null | undefined' is not assignable to type 'PropertyKey'.
+  countBy(data, (x) => x);
+});
--- a/src/countBy.test-d.ts
+++ b/src/countBy.test-d.ts
@@ -22,14 +22,10 @@ describe("dataLast", () => {
   });
 });
 
-test("callback function type does not match data array type", () => {
+test("callback function type", () => {
   const data = [1, 2, 3];
-  // @ts-expect-error [ts2345]: Type 'number' is not assignable to type 'string'.
-  countBy(data, (x: string) => x);
-});
-
-test("array with nullish values", () => {
-  const data = [1, null, undefined];
-  // @ts-expect-error [ts2322] - Type 'number | null | undefined' is not assignable to type 'PropertyKey'.
-  countBy(data, (x) => x);
+  countBy(data, (x) => {
+    expectTypeOf(x).toEqualTypeOf<number>();
+    return x;
+  });
 });
--- a/src/countBy.test.ts
+++ b/src/countBy.test.ts
@@ -126,8 +126,7 @@ describe("dataLast", () => {
 });
 
 test("empty array", () => {
-  const data = [] as Array<string>;
-  const result = countBy(data, (x) => x.toLowerCase());
+  const result = countBy([], (x) => x);
 
   expect(result).toStrictEqual({});
 });
--- a/src/countBy.test.ts
+++ b/src/countBy.test.ts
@@ -1,0 +1,112 @@
+import { countBy } from "./countBy";
+import { pipe } from "./pipe";
+
+describe("dataFirst", () => {
+  test("countBy", () => {
+    const data = [1, 2, 3, 2, 1, 5];
+    const result = countBy(data, (x) => x);
+
+    expect(result).toStrictEqual({ 1: 2, 2: 2, 3: 1, 5: 1 });
+  });
+
+  test("array of strings", () => {
+    const data = ["a", "b", "c", "B", "A", "a"];
+    const result = countBy(data, (x) => x.toLowerCase());
+
+    expect(result).toStrictEqual({ a: 3, b: 2, c: 1 });
+  });
+
+  test("array of objects", () => {
+    const data = [
+      { id: 1, category: "A" },
+      { id: 2, category: "B" },
+      { id: 3, category: "A" },
+    ];
+    const result = countBy(data, (x) => x.category);
+
+    expect(result).toStrictEqual({ A: 2, B: 1 });
+  });
+
+  test("mixed data types", () => {
+    const mySymbol = Symbol("mySymbol");
+    const data = [1, "a", 1, mySymbol, "A", mySymbol];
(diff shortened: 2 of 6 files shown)


src/countBy.ts
```diff
@@ -0,0 +1,57 @@
+import { purry } from "./purry";
+
+/**
+ * Categorize and count elements in an array using a defined callback function. The transformation function applies to each element in the array to determine its category and then counts how many elements fall into each category.
```
eranhirsch, line 4: Please add newlines so that all lines are up to 80 characters long. This makes viewing the code side-by-side easier (for example, when reviewing it 😉 ).

src/countBy.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param data - The array.
+ * @param fn - The transformation function.
+ * @signature
+ *   R.countBy(array, fn)
```
eranhirsch, line 9: The param is called data, so let's keep the naming consistent

```suggestion
 *   R.countBy(data, fn)
```

src/countBy.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @example
+ *    R.countBy(
+ *      ["a", "b", "c", "B", "A", "a"],
+ *      (x) => x.toLowerCase()
```
eranhirsch, line 13: We provide a toLowerCase utility, let's use it in the example too.

```suggestion
 *      R.toLowerCase()
```

src/countBy.ts
```diff
@@ -0,0 +1,57 @@
...
+ *    ) // { a: 3, b: 2, c: 1 }
+ * @dataFirst
+ * @category Array
+ */
+export function countBy<T, K extends PropertyKey>(
+  data: ReadonlyArray<T>,
+  fn: (value: T, index: number, data: ReadonlyArray<T>) => K,
+): Record<K, number>;
```
eranhirsch, line 21: This should be `ExactRecord` which makes sure that we make the result partial when K is a literal union (a finite set of values).

```suggestion
): ExactRecord<K, number>;
```

src/countBy.ts
```diff
@@ -0,0 +1,57 @@
...
+ *    R.countBy(
+ *      ["a", "b", "c", "B", "A", "a"],
+ *      (x) => x.toLowerCase()
+ *    ) // { a: 3, b: 2, c: 1 }
+ * @dataFirst
+ * @category Array
+ */
+export function countBy<T, K extends PropertyKey>(
+  data: ReadonlyArray<T>,
+  fn: (value: T, index: number, data: ReadonlyArray<T>) => K,
```
eranhirsch, line 20: In `groupBy,` we allow the callback function to return `undefined` to allow filtering out values while they are grouped. Granted, this matters more for `groupBy`, where we can skip building a large array just to throw it away, a problem that doesn't manifest itself here, where we only keep a `number` per key, but it's still useful to keep the size of the object small if a lot of keys aren't needed.

To make the API surface more consistent let's support this here too.

```suggestion
  fn: (value: T, index: number, data: ReadonlyArray<T>) => K | undefined,
```

src/countBy.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const countByImplementation = <T>(
+  data: ReadonlyArray<T>,
+  fn: (value: T, index: number, data: ReadonlyArray<T>) => PropertyKey,
+): Record<PropertyKey, number> => {
+  const out: Record<PropertyKey, number> = {};
+
+  for (const [index, item] of data.entries()) {
+    const key = fn(item, index, data);
+    out[key] = (out[key] ?? 0) + 1;
```
eranhirsch, line 53: This could also be written as: 

```suggestion
    out[key] ??= 0;
    out[key] += 1;
```

I don't know what reads out more clearly so it's up to you to decide.

src/countBy.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  test("mixed data types", () => {
+    const mySymbol = Symbol("mySymbol");
+    const data = [1, "a", 1, mySymbol, "A", mySymbol];
+    const result = countBy(data, (x) =>
+      typeof x === "string" ? x.toLowerCase() : x,
+    );
+
+    expect(result).toStrictEqual({ 1: 2, a: 2, [mySymbol]: 2 });
+  });
```
eranhirsch, line 38: Great test! But let's make it more concise by only testing a single concept each time. I think testing symbols is great, so I would like to have a test that only checks that aspect of the implementation.

src/countBy.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      countBy((_, index) => (index % 2 === 0 ? "even" : "odd")),
+    );
+
+    expect(result).toStrictEqual({ even: 3, odd: 2 });
+  });
+});
+
+test("empty array", () => {
+  const data = [] as Array<string>;
+  const result = countBy(data, (x) => x.toLowerCase());
```
eranhirsch, line 109: You don't need the type-casting here, this isn't a type test.
```suggestion
  const data = [];
  const result = countBy(data, (x) => x);
```

src/countBy.test-d.ts
```diff
@@ -0,0 +1,35 @@
...
+    );
+
+    expectTypeOf(result).toEqualTypeOf<Record<string, number>>();
+  });
+});
+
+test("callback function type does not match data array type", () => {
+  const data = [1, 2, 3];
+  // @ts-expect-error [ts2345]: Type 'number' is not assignable to type 'string'.
+  countBy(data, (x: string) => x);
```
eranhirsch, line 28: Prefer testing these with an "expectTypeOf" clause like this:

```suggestion
  countBy(data, (x) => {
    expectTypeOf(x).toEqualTypeOf<number>();
    return x; 
  });
```

src/countBy.test-d.ts
```diff
@@ -0,0 +1,35 @@
...
+});
+
+test("array with nullish values", () => {
+  const data = [1, null, undefined];
+  // @ts-expect-error [ts2322] - Type 'number | null | undefined' is not assignable to type 'PropertyKey'.
+  countBy(data, (x) => x);
+});
```
eranhirsch, line 35: This test isn't valuable. It incorrectly states that it tests arrays with nullish values, but those aren't actually the thing being tested (e.g. `countBy([null], (x) => 1)` would work just like any other value would), and what it does test is that the callback function must return a PropertyKey, and that's already enforced by the typing of the function.

```suggestion
```

src/countBy.test-d.ts
```diff
@@ -0,0 +1,35 @@
...
+import { pipe } from "./pipe";
+
+describe("dataFirst", () => {
+  test("countBy", () => {
+    const data = ["a", "b", "c", "B", "A", "a"];
+    const result = countBy(data, (x) => x.toLowerCase());
+
+    expectTypeOf(result).toEqualTypeOf<Record<string, number>>();
+  });
+});
```
eranhirsch, line 11: I'm missing tests for when the callback function returns a union literal. Those would return the wrong value right now. See my comment in the function code itself on how to fix it.

```suggestion

  test("literal unions", () => {
    const result = countBy([], (x): x is "cat" | "dog" => "cat");
    expectTypeOf(result).toEqualTypeOf<Partial<Record<"cat" | "dog", number>>>();
  });
});
```

eranhirsch (changes requested):
This is a really cool utility, check out my comments!

src/countBy.ts
cjquines: i feel like there's a prettier rule that can enforce this, if we wanted to

grigorischristainas: Sorry about that, VS Code displays the content in multiple lines so I missed that.

src/countBy.ts
grigorischristainas: I agree, this is more readable!

src/countBy.test-d.ts
grigorischristainas: Thank you for the test! I believe it needs a small adjustment. The callback in your example is a type guard returning a boolean, but `countBy` expects a callback that returns a `PropertyKey`. I think that it could be revised as follows:

```ts
const data = [] as Array<"cat" | "dog">;
const result = countBy(data, (x) => x);

expectTypeOf(result).toEqualTypeOf<Partial<Record<"cat" | "dog", number>>>();
```

src/countBy.ts
```diff
@@ -0,0 +1,71 @@
...
+import type { ExactRecord } from "./internal/types/ExactRecord";
+
+/**
+ * Categorize and count elements in an array using a defined callback function.
+ * The transformation function applies to each element in the array to
+ * determine its category and then counts how many elements fall into
+ * each category.
+ *
+ * @param data - The array.
+ * @param fn - The transformation function.
```
eranhirsch, line 11: here too...

eranhirsch (dismissed):
Thanks!

src/countBy.ts
```diff
@@ -0,0 +1,71 @@
+import { purry } from "./purry";
+import type { ExactRecord } from "./internal/types/ExactRecord";
+
+/**
+ * Categorize and count elements in an array using a defined callback function.
+ * The transformation function applies to each element in the array to
```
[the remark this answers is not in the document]
grigorischristainas: You are right, that fits better for description, thanks!

PR #993, opened 2025-01-03
docs: add twoslash

cjquines:
still a draft, needs styling. but twoslash adds types on hover

image


eranhirsch:
The docs site has been refactored quite a bit, and this PR would need a lot of rebasing work to allow it to merge cleanly. I'm closing it now, and we can re-do it in the future when we have the capacity to move it forward.

PR #995, opened 2025-01-03
fix(hasSubObject, setPath): support interfaces

### LINKED ISSUE #990: `setPath` doesn't work for interfaces, only for types
"typescript": "5.7.2"
"remeda": "2.19.0"

```
interface User {
  name: {
    first: string;
  }
}

const user: User = { name: { first: 'John' } };

setPath(user, ['name', 'first'], 'Jane')
```

image

If I change User from `interface` to `type`, it's OK

```
// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
type User = {
  name: {
    first: string;
  }
}

const user: User = { name: { first: 'John' } };

setPath(user, ['name', 'first'], 'Jane')
```

image

Is this by design? If so, this limits this function since there are many types from 3rd party libraries are defined interfaces not types and we can't change them

cjquines:


src/hasSubObject.test-d.ts
```diff
@@ -120,6 +120,23 @@ describe("data-first", () => {
     );
   });
 
+  it("allows interfaces", () => {
+    // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
```
eranhirsch, line 124: This will come up a lot, so we should disable this for all test files.

src/setPath.ts
```diff
@@ -3,7 +3,7 @@ import { purry } from "./purry";
 type Path<T, Prefix extends ReadonlyArray<unknown> = []> =
   T extends ReadonlyArray<unknown>
     ? Path<T[number], [...Prefix, number]> | Prefix
-    : T extends Record<PropertyKey, unknown>
+    : T extends object
```
eranhirsch, line 6: Typescript 5.1 is having a hard time with this type. Do you have any idea why?

eranhirsch (changes requested):
Let's release the hasSubObject changes so they aren't blocked by the problem with Typescript 5.1 and setPath...

PR #1001, opened 2025-01-13
fix(uniqueWith): return items that are not equal to themselves

### LINKED ISSUE #999: `uniqueWith` doesn't return items that are not equal to themselves
```ts
const errors = [
  { id: 1, reason: 'No name' },
  { id: 1, reason: 'No name' },
  { reason: 'No name' },
  { reason: 'No name' }
]

R.uniqueWith(errors, (errorA, errorB) => {
  if(errorA.id === undefined || errorB.id === undefined){
    return false
  }
  return errorA.id === errorB.id
})
```

expected
```ts
[
  { id: 1, reason: 'No name' },
  { reason: 'No name' },
  { reason: 'No name' }
]
```

actual
```ts
[
  { id: 1, reason: 'No name' },
]
```

It was unexpected for me that items not equal to themselves were removed from the array

tbui17:
Fixes #999 

- Previous implementation checked each item for a match that existed at a different index, skipping those. It checks the item's own equivalency as part of the implementation, but this can unexpectedly remove items due to interactions with the comparator's logic as shown in the issue.
- In the new implementation, the index equality check simultaneously handles this edge case and reduces unnecessary calls of the comparator.

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/uniqueWith.ts
+++ b/src/uniqueWith.ts
@@ -55,9 +55,16 @@ export function uniqueWith(...args: ReadonlyArray<unknown>): unknown {
   return purryFromLazy(lazyImplementation, args);
 }
 
-const lazyImplementation =
-  <T>(isEquals: IsEquals<T>): LazyEvaluator<T> =>
-  (value, index, data) =>
-    data.findIndex((otherValue) => isEquals(value, otherValue)) === index
-      ? { done: false, hasNext: true, next: value }
-      : SKIP_ITEM;
+const lazyImplementation = <T>(isEquals: IsEquals<T>): LazyEvaluator<T> => {
+  // the approach is to keep track of distinct entities and skip duplicates. we differentiate them by comparing each known distinct entity against the current value, adding the value to the entity list if there are no matches. worst case O(n^2) time complexity and O(n) space complexity.
+  const entities: Array<T> = [];
+  return (value) => {
+    for (const entity of entities) {
+      if (isEquals(entity, value)) {
+        return SKIP_ITEM;
+      }
+    }
+    entities.push(value);
+    return { done: false, hasNext: true, next: value };
+  };
+};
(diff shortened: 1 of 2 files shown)


src/uniqueWith.ts
```diff
@@ -57,7 +57,16 @@ export function uniqueWith(...args: ReadonlyArray<unknown>): unknown {
 
 const lazyImplementation =
   <T>(isEquals: IsEquals<T>): LazyEvaluator<T> =>
-  (value, index, data) =>
-    data.findIndex((otherValue) => isEquals(value, otherValue)) === index
+  (value, index, data) => {
+    const otherIndex = data.findIndex((otherValue) =>
+      isEquals(value, otherValue),
+    );
```
filipw01, line 63: I wonder if we could optimize it like this to avoid calling `isEquals` on every item against every other Item

```
const firstEqualIndex = data.findIndex((otherValue, otherIndex) => {
      return index === otherIndex || isEquals(value, otherValue)
    });
```

src/uniqueWith.ts
```diff
@@ -57,7 +57,16 @@ export function uniqueWith(...args: ReadonlyArray<unknown>): unknown {
 
 const lazyImplementation =
   <T>(isEquals: IsEquals<T>): LazyEvaluator<T> =>
-  (value, index, data) =>
-    data.findIndex((otherValue) => isEquals(value, otherValue)) === index
+  (value, index, data) => {
+    const otherIndex = data.findIndex((otherValue) =>
```
filipw01, line 61: What do you think about `firstEqualIndex` for the name of this variable instead?

filipw01 (changes requested):

src/uniqueWith.ts
tbui17: It's a lot better! The name reflects the role of the variable now. I was only mirroring the variable names before.

src/uniqueWith.ts
tbui17: It's a good idea. I usually leave optimizations for a separate PR but this can definitely avoid some potentially expensive equality checks and it's a simple change.

filipw01 (approved):

cjquines (approved):
lgtm, thanks so much!

PR #1002, opened 2025-01-15
feat(mergeAll): Create more specific return type for mergeAll

### LINKED ISSUE #918: Create more specific return type for mergeAll when the argument is an array
Currently the return type for mergeAll when used on an array is [object](https://github.com/remeda/remeda/blob/4ef05046809d990a6d0060b56a58d0d1ad4b28ce/src/mergeAll.ts#L20)

[Playground for following snippets
](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wEpwGZQiOAIigFMQSATZQgKFEljhgE8wS4BvAZQAtlSKAVQB2wCMIBiwEgBsKqACIkSYOAF8ceAoRZsAtNhKoYhANw0au9oNQkoAdWAweABR7j2AXk5xgFAFxwxlDAwgDmpnBg7sIkgcIAriAARnbq5lZwNnaOzm4eAIKoXDAh4XDeHL4BQaWhEVExcbVlYemWrNa2Dk48AHLI5BU+foHB9ZHCg83j4ZEQYDDAyFMyACpOMiQA-GN15WrmNADG4sYoUFDIzMMA2gC6KOgFl9cAPNk9eU1wAD5Z3VyriaRRKrT+AJyvQG5AAfEdMhgjAkZDA1p1hhwaHBqntWuYcdEPLsWvUIYkUnYCXApuQSbMIti4AslitkOtNjs8RMaIcLKdhOdSKgUTB-EiRaj0WxhhgAHTkKBhEgFGQyAAU-CuzAAlEcBedFcrVTI+iQAO6yhV2Y1qp5wN5rWHqpla66BJHICjiGTMF7ax2wmg6iqwuBrOAAMjgvH4lBEYkk0jkimUYDeSNOUAobxceDYsGYAGkSMwADRwBLCCgkbChShh6NOo4nM7wWLmiWi8XIqUY7xGlVqs3mzWvXVAA)
I would expect the following to work:

```typescript
type UserWithPhone = { id: string; phone: number };
type UserWithPhoneAsString = { id: string; phone: string };
type UserWithName = { id: string; name: string; optianalTitle?: string };

const array = [] as Array<UserWithPhone | UserWithPhoneAsString | UserWithName>;

type ResultType = {
  id: string;
  phone?: string | number;
  name?: string;
  optianalTitle?: string;
};

const result:ResultType = R.mergeAll(array);
```

After some trial an error i came to this

```typescript
const result:ResultType = R.mergeAll(array);

const mergeAllNew = R.mergeAll as <T>(
  array: ReadonlyArray<T>
) => T & SharedUnionFieldsDeep<Record<PropertyKey, undefined> & T>;


const newResult:ResultType = mergeAllNew(array)
```

One problem with that is that optional types like name are explicitly typed as undefined which is wrong.

I am willing to work on this but have to admit that i might not have the  TypeScript skills yet.

Help or feedback would be highly appreciated.

### LINKED ISSUE #1006: mergeAll tuple overload signatures are unsafe
The runtime behavior of `mergeAll` shallowly merges array contents into a new object. Although it is difficult to produce a perfectly accurate signature for merging in general, the current implementation can produce incorrect return types that mislead the consumer into triggering runtime exceptions for a fairly common scenario: nested objects. 

When objects are intersected, shared fields will also be intersected recursively. Due to the mismatch in depth, nested objects in the result can be incorrectly inferred as having a field that they do not actually have at runtime. See the example test for the 2-item tuple overload below.

```typescript
describe("invalid tuple overload signature", () => {
  type InnerA = { firstField: string };
  type A = { inner: InnerA };
  type InnerB = { secondField: string };
  type B = { inner: InnerB };
  const obj1: A = { inner: { firstField: "obj1" } };
  const obj2: B = { inner: { secondField: "obj2" } };
  const input: [A, B] = [obj1, obj2] as const;

  const result: A & B = mergeAll(input);
  const inner: InnerA & InnerB = result.inner;
  const firstField: string = inner.firstField;
  const secondField: string = inner.secondField;

  it("should have both firstField and secondField", () => {
    expect.soft(secondField).toBe("obj2");
    expect.soft(firstField).toBe("obj1");
  });

  it("firstField should not be undefined", () => {
    expect(firstField).toBeDefined();
  });
});

```

```bash
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Tests 2 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  src/mergeAll.test.ts > invalid tuple overload signature > should have both firstField and secondField
AssertionError: expected undefined to be 'obj1' // Object.is equality

- Expected:
"obj1"

+ Received:
undefined

 ❯ src/mergeAll.test.ts:38:29
     36|   it("should have both firstField and secondField", () => {
     37|     expect.soft(secondField).toBe("obj2");
     38|     expect.soft(firstField).toBe("obj1");
       |                             ^
     39|   });
     40|

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/2]⎯

 FAIL  src/mergeAll.test.ts > invalid tuple overload signature > firstField should not be undefined
AssertionError: expected undefined to be defined
 ❯ src/mergeAll.test.ts:42:24
     40|
     41|   it("firstField should not be undefined", () => {
     42|     expect(firstField).toBeDefined();
       |                        ^
     43|   });
     44| });

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[2/2]⎯

 Test Files  1 failed (1)
      Tests  2 failed | 2 passed (4)
   Start at  17:50:54
   Duration  1.36s (transform 58ms, setup 0ms, collect 34ms, tests 7ms, environment 0ms, prepare 487ms)
```

At the very least, the return type should be changed to use `Merge` from type-fest instead of intersection which should handle this scenario. It has at least as much safety as intersection for most other edge cases like arrays containing both objects and primitives. No one would realistically attempt to create this as a tuple and merge it though.

tbui17:
Fixes #918 
Fixes #1006

- The return type for arrays is now a narrowed object encapsulating the possible fields and value types that can appear after merging. There is further type narrowing if the array is known to be nonempty.
- The return type for tuples now correctly resolves the final, specific type of overridden fields during merging. 
- The function now accepts tuples of any size.
- Added internal helper types: 
  - `DisjointUnionFields`

---

Make sure that you:

---

### Available Types:

existing).

</details>


src/mergeAll.ts
```diff
@@ -1,11 +1,38 @@
+import type { SharedUnionFields, Simplify } from "type-fest";
+import type { IsUnion } from "./internal/types/IsUnion";
+import type { SharedUnionFieldsComplement } from "./internal/types/SharedUnionFieldsComplement";
+
+/**
+ * Merges all types of the union into a single object type. Fields that are not shared among all types of the union become optional.
+ */
+type MergeUnionWithOptionalSharedUnionFieldsComplement<T extends object> =
```
tbui17, line 8: I just can't come up with a good name for this

src/internal/types/PickUnionValue.ts
cjquines, line 1: **suggestion:** let's inline this type into `SharedUnionFieldsComplement`

i don't see the value in splitting this out, or having separate tests for it

src/internal/types/SharedUnionFieldKeysComplement.ts
cjquines, line 1: **suggestion:** let's move this type inside `SharedUnionFieldsComplement`, and unexport it

if we'll need to export it later we can separate the file, but i'd rather we keep the internal interface smaller

src/internal/types/PickUnionValue.ts
```diff
@@ -0,0 +1,9 @@
+import type { KeysOfUnion } from "type-fest";
+
+/**
+ * Gets the union of the field value types from the types of the union where the key exists.
+ */
+export type PickUnionValue<T extends object, K extends KeysOfUnion<T>> =
+  T extends Partial<Record<K, unknown>> // if T contains K (distributed, can be optional)
+    ? T[K]
```
cjquines, line 8: **nit:** maybe this is a matter of taste, but i prefer `T extends Partial<Record<K, infer V>> ? V : never`

src/internal/types/SharedUnionFieldsComplement.test-d.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  it("should not have the keys of SharedUnionFields", () => {
+    expectTypeOf<Result>().not.toHaveProperty("a");
+    expectTypeOf<Result>().not.toHaveProperty("b");
+  });
+
+  it("should have the complementary keys of SharedUnionFields", () => {
+    expectTypeOf<Result>().toHaveProperty("c");
+    expectTypeOf<Result>().toHaveProperty("d");
+  });
```
cjquines, line 19: **suggestion:** let's drop these tests, i don't think they do anything

instead let's just name the top-level test "should have the complement…"

src/internal/types/SharedUnionFieldsComplement.test-d.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  it("should have the complement of SharedUnionFields", () => {
+    type ExpectedResult = { c: string | number; d: string };
+
+    expectTypeOf<Result>().toEqualTypeOf<ExpectedResult>();
```
cjquines, line 24: **nit:** as a matter of style, i think we inline types, see the other tests in this folder

```suggestion
    expectTypeOf<SharedUnionFieldsComplement<A | B | C>>().toEqualTypeOf<{
      c: string | number;
      d: string;
    }>();
```

src/mergeAll.ts
```diff
@@ -1,11 +1,34 @@
+import type { EmptyObject, SharedUnionFields, Simplify } from "type-fest";
+import type { SharedUnionFieldsComplement } from "./internal/types/SharedUnionFieldsComplement";
+
+/**
+ * Merges all types of the union into a single object type. Fields that are not shared among all types of the union become optional.
+ */
+type MergeUnionAndPartializeSharedUnionFieldsComplement<T extends object> =
+  SharedUnionFields<T> & Partial<SharedUnionFieldsComplement<T>>; // for single-member "union", the complement would be an empty object. intersecting with it would be a no-op.
```
cjquines, line 8: **suggestion:** let's inline this type, we don't need to give it a name (especially a long name like this)

src/mergeAll.ts
```diff
@@ -1,11 +1,34 @@
...
+
+// In the context of a heterogeneous array, the array may not have objects from every type of the union.
+// This means some fields may be missing in the final object, so we make them optional.
+// If there is a non-optional field shared among all members of the union, then we know that if the array is not empty, the field must be present and avoid becoming optional.
+// If the field is already optional, this doesn't really matter.
+// If the array is empty, we know that the loop won't run so we'll just get the empty object.
+// Since we don't know the order of the items in the array, when we merge common fields, we don't know what the final type for the field will be, but we do know that it is one of the many possible types that are available across the members of the union for that field.
+// We represent these possibilities by combining the field's different types across the union members into a union.
+type MergeAllArrayResult<T extends object> =
```
cjquines, line 17: **nit:** i think our style is to just call this `MergeAll`, or maybe `MergeUnion`

src/mergeAll.ts
```diff
@@ -1,11 +1,34 @@
...
+  | EmptyObject;
+
 /**
  * Merges a list of objects into a single object.
  *
  * @param array - The array of objects.
+ * @returns A new object merged with all of the objects in the list. If the list is empty, an empty object is returned.
  * @signature
- *    R.mergeAll(objects)
+ *    R.mergeAll(array)
```
cjquines, line 27: **question:** what do you think about naming this `objects` instead of `array`, and changing everything else?

src/mergeAll.ts
```diff
@@ -1,11 +1,34 @@
...
+// In the context of a heterogeneous array, the array may not have objects from every type of the union.
+// This means some fields may be missing in the final object, so we make them optional.
+// If there is a non-optional field shared among all members of the union, then we know that if the array is not empty, the field must be present and avoid becoming optional.
+// If the field is already optional, this doesn't really matter.
+// If the array is empty, we know that the loop won't run so we'll just get the empty object.
+// Since we don't know the order of the items in the array, when we merge common fields, we don't know what the final type for the field will be, but we do know that it is one of the many possible types that are available across the members of the union for that field.
+// We represent these possibilities by combining the field's different types across the union members into a union.
+type MergeAllArrayResult<T extends object> =
+  | Simplify<MergeUnionAndPartializeSharedUnionFieldsComplement<T>>
+  | EmptyObject;
```
cjquines, line 19: **question:** do you think it's worth it to have a separate type if the array is non-empty? i'm not actually sure

src/mergeAll.test-d.ts
```diff
@@ -0,0 +1,129 @@
...
+          id: string;
+          phone?: string | number;
+          name?: string;
+          optianalTitle?: string;
+        }
+      | EmptyObject;
+
+    const mergedUserUnion = mergeAll(userUnionArray);
+
+    expectTypeOf(mergedUserUnion).toEqualTypeOf<ExpectedResultType>();
```
cjquines, line 24: **nit:** same nit w.r.t. having inlined types in the assertions

src/mergeAll.ts
```diff
@@ -17,12 +40,15 @@ export function mergeAll<A, B, C, D>(
 export function mergeAll<A, B, C, D, E>(
   array: readonly [A, B, C, D, E],
 ): A & B & C & D & E;
-export function mergeAll(array: ReadonlyArray<object>): object;
-
-export function mergeAll(items: ReadonlyArray<object>): object {
+export function mergeAll<T extends object>(
+  array: ReadonlyArray<T>,
```
cjquines, line 44: **thought:** i'm now wondering whether this should've been a `IterableContainer<T>`

src/mergeAll.ts
tbui17: Yes, that clarifies things a lot. I wasn't sure if the function was intended to only be used with objects. Is there a reason for leaving out the object constraint or is this code just outdated?

src/internal/types/PickUnionValue.ts
tbui17: Yes, I like it. It's rather clever.

src/mergeAll.ts
tbui17: It doesn't seem like it'd be too much effort to add an additional overload for it barring any overload resolution issues. I think it's a good idea.

tbui17:
Added suggested changes. Thanks for the review!

src/mergeAll.ts
```diff
@@ -1,28 +1,67 @@
...
+type MergeUnion<T extends object> = Simplify<
+  SharedUnionFields<T> & Partial<DisjointUnionFields<T>>
+>;
+
+type MergeAll<T extends IterableContainer<object>> =
+  // determine if it's a tuple or array
+  TupleParts<T> extends { item: never }
+    ? T extends readonly []
+      ? EmptyObject
+      : MergeTuple<T>
```
tbui17, line 37: @cjquines Should `MergeTuple` still be inlined? It seems it would get fairly difficult to read here with the additional nested ternaries. Looking at the code for `TupleParts` though, it seems like the team is pretty comfortable with them.

src/internal/types/SharedUnionFieldsComplement.test-d.ts
tbui17: Totally missed this! I must have been tired last week. I'll make these adjustments.

src/mergeAll.ts
```diff
@@ -1,28 +1,67 @@
...
-export function mergeAll<A, B, C, D, E>(
-  array: readonly [A, B, C, D, E],
-): A & B & C & D & E;
-export function mergeAll(array: ReadonlyArray<object>): object;
-
-export function mergeAll(items: ReadonlyArray<object>): object {
+export function mergeAll<T extends object>(
+  objects: NonemptyReadonlyArray<T>,
+): MergeUnion<T>;
+export function mergeAll<T extends IterableContainer<object>>(
```
tbui17, line 58: Needed to change constraint to object, compiler will complain if a non-object is used with spreading (it'll still work at runtime i.e. with strings)

src/mergeAll.ts
cjquines: great, let's change it to object then

src/internal/types/DisjointUnionFields.ts
```diff
@@ -0,0 +1,20 @@
+import type { KeysOfUnion, SharedUnionFields } from "type-fest";
+
+/**
+ * Gets the complement of the keys of {@link SharedUnionFields} from a union.
+ */
+type SharedUnionFieldKeysComplement<T extends object> = Exclude<
```
cjquines, line 6: **nit:** DisjointUnionFieldKeys?

src/mergeAll.ts
```diff
@@ -1,28 +1,67 @@
...
+
+type MergeAll<T extends IterableContainer<object>> =
+  // determine if it's a tuple or array
+  TupleParts<T> extends { item: never }
+    ? T extends readonly []
+      ? EmptyObject
+      : MergeTuple<T>
+    : MergeUnion<T[number]> | EmptyObject;
+
+type NonemptyReadonlyArray<T> = readonly [T, ...Array<T>];
```
tbui17: Referenced existing `NonEmptyArray` for style

src/mergeAll.ts
```diff
@@ -1,28 +1,67 @@
...
+
+type MergeAll<T extends IterableContainer<object>> =
+  // determine if it's a tuple or array
+  TupleParts<T> extends { item: never }
+    ? T extends readonly []
+      ? EmptyObject
+      : MergeTuple<T>
+    : MergeUnion<T[number]> | EmptyObject;
+
+type NonEmptyReadonlyArray<T> = readonly [T, ...Array<T>];
```
cjquines, line 40: **note:** we have this type already, it's called NonEmptyArray

src/mergeAll.test-d.ts
```diff
@@ -0,0 +1,175 @@
...
+import { mergeAll } from "./mergeAll";
+
+describe("arrays", () => {
+  it("custom case", () => {
+    // based on https://github.com/remeda/remeda/issues/918
+
+    const userUnionArray: ReadonlyArray<
+      | { id: string; phone: number }
+      | { id: string; phone: string }
+      | { id: string; name: string; optianalTitle?: string }
```
cjquines, line 11: **typo:** `optionalTitle`

cjquines (approved):
will merge once remaining stuff is resolved

cjquines:
thanks so much for all the work by the way, this was a large PR!

src/mergeAll.ts
tbui17: There is indeed NonEmptyArray, but it isn't read only. Using `readonly NonEmptyArray` as a type apparently isn't valid.

tbui17:
Fixed final bits of style issues. This PR was a nice refresher on advanced Typescript.

src/mergeAll.ts
cjquines: `Readonly<NonEmptyArray<T>>` is what we use for e.g. `sortBy` or `chunk`

cjquines: lemme try

tbui17: Oh right, I forgot `Readonly` exists as a generic type. It's been so long I forgot all the details. 😂
I'll make sure to look around in the codebase in future PRs to smooth things out.

PR #1004, opened 2025-01-15
fix(floor, ceil, round): do not divide and multiply by a floating point number

### LINKED ISSUE #1003: The negative precision `floor` returns floating point result
Hey there 👋 

Thank you for developing such a nice library!

During development I faced an unexpected result caused by the `floor` function when applied with the negative precision:

```ts
expects(floor(123456, -5)).toBe(100000)
// actual is 100000.00000000001
```

I believe the rootcause is somewhere in the [`internal/withPrecision.ts`](https://github.com/remeda/remeda/blob/main/src/internal/withPrecision.ts) file during division operation.

owanturist:
The changes aim to increase the accuracy of the `floor`, `cail`, and `round` functions by using only integers when dividing and multiplying, avoiding floating point numbers.

Resolves #1003 

---

Make sure that you:

- [ ] ~Typedoc added for new methods and updated for changed~
- [ ] ~New methods added to `src/index.ts`~
- [ ] ~New methods added to `docs/src/content/mapping` for both Lodash and Ramda.~

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/ceil.test.ts
+++ b/src/ceil.test.ts
@@ -12,6 +12,11 @@ describe("data-first", () => {
     expect(ceil(8123.4317, -4)).toBe(10_000);
   });
 
+  // Makes sure the issue github.com/remeda/remeda/issues/1003 doesn't happen.
+  test("does not output floating point number for negative precision", () => {
+    expect(ceil(123_456, -5)).toBe(200_000);
+  });
+
   test("should work with precision = 0", () => {
     expect(ceil(8123.4317, 0)).toBe(8124);
   });
--- a/src/internal/withPrecision.ts
+++ b/src/internal/withPrecision.ts
@@ -26,7 +26,14 @@ export function withPrecision(roundingFn: (value: number) => number) {
       return roundingFn(value);
     }
 
-    const multiplier = RADIX ** precision;
-    return roundingFn(value * multiplier) / multiplier;
+    const multiplier = RADIX ** Math.abs(precision);
+
+    if (precision > 0) {
+      return roundingFn(value * multiplier) / multiplier;
+    }
+
+    // It is tricky to divide and multiple by a floating point number.
+    // Instead, it can use integers, but first divide and then multiply.
+    return roundingFn(value / multiplier) * multiplier;
   };
 }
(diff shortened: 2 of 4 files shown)


src/internal/withPrecision.ts
```diff
@@ -26,7 +26,14 @@ export function withPrecision(roundingFn: (value: number) => number) {
...
-    return roundingFn(value * multiplier) / multiplier;
+    const multiplier = RADIX ** Math.abs(precision);
+
+    if (precision > 0) {
+      return roundingFn(value * multiplier) / multiplier;
+    }
+
+    // It is tricky to divide and multiple by a floating point number.
+    // Instead, it can use integers, but first divide and then multiply.
+    return roundingFn(value / multiplier) * multiplier;
```
cjquines, line 37: **nit:** i would rather we split the multiplier in either case, so it's clearer what's happening, wdyt?

```suggestion
    if (precision > 0) {
      const multiplier = RADIX ** precision;
      return roundingFn(value * multiplier) / multiplier;
    }

    // Avoid losing precision by dividing first.
    const divisor = RADIX ** -precision;
    return roundingFn(value / divisor) * divisor;
```

src/ceil.test.ts
```diff
@@ -12,6 +12,11 @@ describe("data-first", () => {
     expect(ceil(8123.4317, -4)).toBe(10_000);
   });
 
+  // Makes sure the issue github.com/remeda/remeda/issues/1003 doesn't happen.
+  test("does not output floating point number for negative precision", () => {
+    expect(ceil(123_456, -5)).toBe(200_000);
```
cjquines, line 17: **nit:** i think we can just add this to the last case, and omit the comment, wdyt

cjquines:
thanks for reporting + fixing btw!!

src/internal/withPrecision.ts
owanturist: yep, makes total sense to me too!

src/ceil.test.ts
owanturist: I usually prefer to keep track of tests against bugs in a separate test case with meaningful info so it is clear why the input is verified. If you are confident that the context won't be lost and forgotten, I can merge it into the previous test case.

cjquines: hmm fair enough, i think that's fine

PR #1005, opened 2025-01-16
fix(mergeAll): Improve type safety of tuple overloads

### LINKED ISSUE #1006: mergeAll tuple overload signatures are unsafe

tbui17:
Fixes #1006 

- Changed tuple overloads to use `Merge` from type-fest instead of intersections to represent merged objects.
- Refactored return types to use recursive type to handle tuples of any arity. Can look into changing the overloads into a single signature as well after #1002 as separate PR as there may be potential complexities with overload resolution.

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/src/internal/types/MergeTuple.ts
+++ b/src/internal/types/MergeTuple.ts
@@ -1,0 +1,11 @@
+import type { Merge } from "type-fest";
+
+/**
+ * Merges a tuple of objects types into a single object type from left to right.
+ */
+export type MergeTuple<
+  T extends ReadonlyArray<unknown>, // ideally unknown is narrowed to object for type safety but this is being used by mergeAll and would require additional object constraint on the function, could be a breaking change
+  _Current = object, // no-op for the first iteration in the successive merges, also infers object as type by default if an empty tuple is used
+> = T extends readonly [infer Head, ...infer PoppedStack] // if stack not empty
+  ? MergeTuple<PoppedStack, Merge<_Current, Head>>
+  : _Current;
--- a/src/mergeAll.ts
+++ b/src/mergeAll.ts
@@ -1,3 +1,5 @@
+import type { MergeTuple } from "./internal/types/MergeTuple";
+
 /**
  * Merges a list of objects into a single object.
  *
@@ -9,11 +11,15 @@
  * @category Array
  */
 export function mergeAll<A>(array: readonly [A]): A;
-export function mergeAll<A, B>(array: readonly [A, B]): A & B;
-export function mergeAll<A, B, C>(array: readonly [A, B, C]): A & B & C;
+export function mergeAll<A, B>(
+  array: readonly [A, B],
+): MergeTuple<typeof array>;
+export function mergeAll<A, B, C>(
+  array: readonly [A, B, C],
+): MergeTuple<typeof array>;
 export function mergeAll<A, B, C, D>(
   array: readonly [A, B, C, D],
-): A & B & C & D;
+): MergeTuple<typeof array>;
 export function mergeAll<A, B, C, D, E>(
   array: readonly [A, B, C, D, E],
 ): A & B & C & D & E;
--- a/src/mergeAll.ts
+++ b/src/mergeAll.ts
@@ -22,7 +22,7 @@ export function mergeAll<A, B, C, D>(
 ): MergeTuple<typeof array>;
 export function mergeAll<A, B, C, D, E>(
   array: readonly [A, B, C, D, E],
-): A & B & C & D & E;
+): MergeTuple<typeof array>;
 export function mergeAll(array: ReadonlyArray<object>): object;
 
 export function mergeAll(items: ReadonlyArray<object>): object {
--- a/src/internal/types/MergeTuple.test-d.ts
+++ b/src/internal/types/MergeTuple.test-d.ts
@@ -1,0 +1,25 @@
+import type { MergeTuple } from "./MergeTuple";
+
+describe("the fields of the rightmost item should have the greatest priority in overrides", () => {
+  it("2 items", () => {
+    type A1 = { a: 1; b: 1 };
+    type A2 = { a: 2 };
+    type Input = [A1, A2];
+    type ExpectedType = { a: 2; b: 1 };
+    type Result = MergeTuple<Input>;
+
+    expectTypeOf<Result>().toEqualTypeOf<ExpectedType>();
+  });
+
+  it("3 items", () => {
+    type A1 = { a: 1; b: 1 };
+    type A2 = { a: 2 };
+    type A3 = { a: 3 };
+    type Input = [A1, A2, A3];
+    type ExpectedType = { a: 3; b: 1 };
+
+    type Result = MergeTuple<Input>;
+
+    expectTypeOf<Result>().toEqualTypeOf<ExpectedType>();
+  });
+});
--- a/src/mergeAll.test-d.ts
+++ b/src/mergeAll.test-d.ts
@@ -1,0 +1,28 @@
+import { mergeAll } from "./mergeAll";
+
+describe("tuple overload", () => {
+  describe("the fields of the rightmost item should have the greatest priority in overrides", () => {
+    it("2 types", () => {
+      type A1 = { a: 1; b: 1 };
+      type A2 = { a: 2 };
+      const input: [A1, A2] = [{ a: 1, b: 1 }, { a: 2 }];
+      type ExpectedType = { a: 2; b: 1 };
+
+      const result = mergeAll(input);
+
+      expectTypeOf(result).toEqualTypeOf<ExpectedType>();
+    });
+
+    it("3 types", () => {
+      type A1 = { a: 1; b: 1 };
+      type A2 = { a: 2 };
+      type A3 = { a: 3 };
+      const input: [A1, A2, A3] = [{ a: 1, b: 1 }, { a: 2 }, { a: 3 }];
+      type ExpectedType = { a: 3; b: 1 };
+
+      const result = mergeAll(input);
+
+      expectTypeOf(result).toEqualTypeOf<ExpectedType>();
+    });
+  });
+});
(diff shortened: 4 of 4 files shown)


src/internal/types/MergeTuple.ts
```diff
@@ -0,0 +1,11 @@
+import type { Merge } from "type-fest";
+
+/**
+ * Merges a tuple of objects types into a single object type from left to right.
+ */
+export type MergeTuple<
+  T extends ReadonlyArray<unknown>, // ideally unknown is narrowed to object for type safety but this is being used by mergeAll and would require additional object constraint on the function, could be a breaking change
```
cjquines, line 7: **suggestion:** you can write `IterableContainer` here instead of `ReadonlyArray<unknown>`, and omit the comment

src/internal/types/MergeTuple.ts
```diff
@@ -0,0 +1,11 @@
...
+  _Current = object, // no-op for the first iteration in the successive merges, also infers object as type by default if an empty tuple is used
```
cjquines, line 8: **nit:** maybe `Result`?

i also don't think we have to prefix with underscore but that's a small nit

src/internal/types/MergeTuple.ts
```diff
@@ -0,0 +1,11 @@
...
+> = T extends readonly [infer Head, ...infer PoppedStack] // if stack not empty
```
cjquines, line 9: **nit:** i think our convention is to use `Rest` or `Tail`

src/internal/types/MergeTuple.ts
```diff
@@ -0,0 +1,11 @@
+import type { Merge } from "type-fest";
+
+/**
+ * Merges a tuple of objects types into a single object type from left to right.
```
cjquines, line 4: **nit:** i think this description could be ambiguous

maybe "Merge a tuple of object types, where props from later objects override earlier props"?

src/internal/types/MergeTuple.ts
cjquines, line 1: **suggestion:** let's inline this into mergeAll

i don't see any point into having it as a separate file with a separate test file, it's literally the same thing as the mergeAll result for tuples

src/mergeAll.ts
```diff
@@ -9,14 +11,18 @@
...
+  array: readonly [A, B, C],
+): MergeTuple<typeof array>;
 export function mergeAll<A, B, C, D>(
   array: readonly [A, B, C, D],
-): A & B & C & D;
+): MergeTuple<typeof array>;
 export function mergeAll<A, B, C, D, E>(
   array: readonly [A, B, C, D, E],
-): A & B & C & D & E;
+): MergeTuple<typeof array>;
```
cjquines, line 25: **question:** maybe you can try this, but can you replace all of these with `mergeAll<T extends IterableContainer>(objects: T): MergeTuple<T>`?

cjquines: if so, then we should probably merge this with #1002, and use `TupleParts` to combine the two cases into a single type

src/internal/types/MergeTuple.ts
tbui17: I was thinking that the type could be reused in future functions at a higher level of abstraction, like possibly a mergeAllDeep. I guess it would be fine to just pull it out of `mergeAll` if the time comes. I had the tests in there as a regression check since the internal scope is rather large.

src/internal/types/MergeTuple.ts
tbui17: Yes! I felt like there was some ambiguity at the time but couldn't really come up with a clearer way to say it haha.

src/internal/types/MergeTuple.ts
tbui17: I was using the parameter position purely as an internal implementation detail for the recursion. Wasn't too sure how to else to denote that.

src/mergeAll.ts
tbui17: Tests indeed still pass with the changes.

tbui17:
PR contents merged into #1002, closing.

PR #1012, opened 2025-01-28
docs: Use typedoc as a Collections API loader

eranhirsch:
This PR is a continuation of the work done in: #997

Instead of running typedoc as a pre-build step manually and then loading and parsing the JSON file we can run TypeDoc via it's programatic API so that it runs as part of the astro build phase.


cjquines (approved):
(i'm pro `toString`, unclear if i'm pro `throws`)

PR #1013, opened 2025-01-30
docs: hot typedoc reload

eranhirsch:
Watch the source files and reload the docs site dynamically when changes are detected.


eranhirsch:
@cjquines  - I'm merging this into the bigger PR because the changes overlap, and I don't think the split makes reviewing this easier.

PR #1024, opened 2025-02-02
fix: remove exported types from `pipe` + version bumps

eranhirsch:
Doing several minor changes surfaced by recent changes in #997 and the montly dependabot bumps:

* dont exporting setPathImplementation
* dont export any types from `pipe`. They are all internal to our impl
* Nits and comment following version bumps (no code changes).


src/swapIndices.ts
```diff
@@ -148,7 +148,8 @@ const swapIndicesImplementation = (
   index2: number,
 ): unknown =>
   typeof data === "string"
-    ? swapArray([...data], index1, index2).join("")
+    ? // eslint-disable-next-line @typescript-eslint/no-misused-spread -- TODO: I'm not sure what the right way to split the string here, there are multiple "correct" answers and each one is meaningfully different: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2521
```
eranhirsch, line 151: @cjquines  - check this out; it's an interesting read. I didn't change the runtime behavior, but it's good to know that this is just one of a few possible ways to consider the string.

cjquines: that's, ugh, yeah, i forgot about unicode

src/mapValues.test-d.ts
```diff
@@ -82,6 +82,7 @@ describe("branded types", () => {
 
     mapValues(userValues, (value, key) => {
       expectTypeOf(value).toEqualTypeOf<number>();
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-template-expression -- TODO: I'm not sure what's going on here, are we doing something wrong or is this code needed so we can test it properly?
       expectTypeOf(key).toEqualTypeOf<`${UserID}`>();
```
cjquines, line 86: i remember branded types need to have a template expression for this to work properly

cjquines (approved):

PR #1029, opened 2025-02-17
chore: monorepo

eranhirsch:
Our current project structure makes it hard to configure prettier and eslint properly for each project, it also makes it harder to add new spin-offs and ideas into our main offering because there's no clear place to put them. with a mono-repo we can segment our source code more effectivly.


cjquines:
ah, yeah, i was going to suggest this at some point

it might also be an appropriate time to switch package managers, if we wanted to. i have not seen npm used with monorepos that much, it's always pnpm (sometimes yarn)

eranhirsch:
redid this in #1031

PR #1031, opened 2025-02-19
chore: monorepo

eranhirsch:


packages/pr-sandbox/README.md
cjquines, line 1: **question:** have you actually used the pr sandbox? i've never used it

remeda.code-workspace
cjquines, line 1: **question:** is this a vs code thing?

cjquines (approved):

packages/pr-sandbox/README.md
eranhirsch: I might have once or twice. I was almost about to remove this because it was too much hassle to set it up correctly, but I managed to get it working, so I kept it for now.

remeda.code-workspace
eranhirsch: yes! It allows vscode/cursor to load different configs for prettier/eslint/vscode side-by-side (so if you have a file open from docs/ and a file open from remeda/ the built-in prettier and eslint plugins would load with different configs for each).

Without this you can't sync the tooling with their CLI counterparts.

PR #1043, opened 2025-02-26
docs: tailwind v4 + react v19

eranhirsch:
Using the shadcn guide for tailwind 4: https://ui.shadcn.com/docs/tailwind-v4. I recreated the whole project config from scratch using the shadcn `init` and `add` CLI commands to get a fresh start (instead of migrating the components individually).

This allows us to migrate to react 19 too, which is the last major version upgrade from recent releases that we needed to sync up to.


cjquines:
i'll have to pull and check how the docs look locally

cjquines (commented):
sidebar looks off in the preview

image

eranhirsch:
Good catch. It's so weird because it works locally. Looking into this...

eranhirsch:
It works now. I didn't change anything; perhaps it was a temporary glitch with how the project was built or how Netlify cached the CSS file(s).

cjquines (approved):
yeah, i checked locally as well. sgtm

PR #1044, opened 2025-03-01
chore: Typescript 5.8

eranhirsch:
Bumping typescript versions across all projects (and any dependents like tsup, type-fest, eslint, etc...).

Issues:
1. isNullish can now narrow the "else" case, so our test needs to reflect that
2. toMatchTypeOf has been deprecated in favour of toExtends and toMatchObjectTypeOf
3. Our workspace wasn't configured correctly to use the node_modules version of typescript
4. the pr-sandbox was not ESM, and wasn't comfortable to use.


cjquines (approved):

cjquines:
awesome, thanks

Issue #1046, opened 2025-03-04
`groupBy` throws on certain valid inputs


eeshugerman:
More specifically, it throws a `TypeError` when `callbackfn` returns a string that is the name of an [`Object` instance method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#instance_methods).

```typescript
const data1 = [{ foo: 'bar' }];
const data2 = [{ foo: 'toString' }];

R.groupBy(data1, R.prop('foo')) // all good
// { bar: [ { foo: 'bar' } ] }

R.groupBy(data2, R.prop('foo')) // error
// Uncaught TypeError: r.push is not a function
//    at p (/Users/elliott.shugerman/devel/bodata/node_modules/remeda/dist/chunk-KF4UUN4H.cjs:1:314)
//    at u (/Users/elliott.shugerman/devel/bodata/node_modules/remeda/dist/chunk-4CZEVLYK.cjs:1:188)
//    at Object.i [as groupBy] (/Users/elliott.shugerman/devel/bodata/node_modules/remeda/dist/chunk-KF4UUN4H.cjs:1:174)

```

The issue stems from [here](https://github.com/remeda/remeda/blob/8354a3adf56b75fcea489b86a27810c7ac5dd2cd/packages/remeda/src/groupBy.ts#L88), where `items` is bound to `output['toString']`, which is a function.

In case anyone's curious, I found this issue thanks to [fast-check](https://fast-check.dev/)

tbui17:
I think this could be relevant in the context of security and prototype pollution. You might be grouping items by a user-provided value when aggregating data. Sanitizing user input would be more of the application developer's responsibility though.

eeshugerman:
👍 I can certainly understand not prioritizing this.

> probably no one will run into in any real-life use-cases

But I wouldn't go that far. What if you're building some kind of static analyzer/tooling for JS/TS? Then it seems quite likely you'd run into this.

It also makes property testing and fuzzing (with fast-check or otherwise) more difficult for projects that use Remeda.

Anyway as a workaround I've copied `groupBy` into my code and replaced the accumulator object with a `Map`. I'll PR that soon as I have a chance.

cjquines:
yeah, i agree this is actually kinda high-pri rather than low-pri; i can't get to this before the weekend though.

PR #1049, opened 2025-03-08
fix: Handling of strings that are `Object` instance method names (in`groupBy` and `countBy`)

eeshugerman:
Closes https://github.com/remeda/remeda/issues/1046

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/packages/remeda/src/groupBy.test.ts
+++ b/packages/remeda/src/groupBy.test.ts
@@ -74,7 +74,8 @@ describe("filtering on undefined grouper result", () => {
   });
 });
 
-describe("tricky string", () => {
+// https://github.com/remeda/remeda/pull/1049
+describe("key is an object instance method name", () => {
   test("groupBy", () => {
     expect(
       groupBy(
--- a/packages/remeda/src/groupBy.test.ts
+++ b/packages/remeda/src/groupBy.test.ts
@@ -73,3 +73,26 @@ describe("filtering on undefined grouper result", () => {
     expect(result.even).toStrictEqual(["a", "c", "e", "g", "i"]);
   });
 });
+
+describe("tricky string", () => {
+  test("groupBy", () => {
+    expect(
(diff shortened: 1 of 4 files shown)


packages/remeda/src/groupBy.test.ts
```diff
@@ -73,3 +73,26 @@ describe("filtering on undefined grouper result", () => {
     expect(result.even).toStrictEqual(["a", "c", "e", "g", "i"]);
   });
 });
+
+describe("tricky string", () => {
```
cjquines, line 77: **suggestion:** let's link the original github issue, i think that's what we usually do for regressions

cjquines: maybe let's just also call the strings "Object instance methods"; so people aren't confused when they see the test

cjquines (commented):
**question:** do you happen to know if there are other cases where this might be an issue? off the top of my head, i can think of `mapValues`, `omitBy`, `indexBy`, `mapToObj`, `countBy`, `pickBy`, `mapKeys`, `evolve`, possibly. i'd rather we solve these all together

packages/remeda/src/groupBy.test.ts
eeshugerman: :+1: linked this ticket and renamed the test

eeshugerman:
> **question:** do you happen to know if there are other cases where this might be an issue? off the top of my head, i can think of `mapValues`, `omitBy`, `indexBy`, `mapToObj`, `countBy`, `pickBy`, `mapKeys`, `evolve`, `addProp` possibly. i'd rather we solve these all together

I don't know. I took a quick look at all those you mentioned here, and the only one that looks like it has this problem is `countBy`. I went ahead and fixed it in this PR.

cjquines (approved):

PR #1061, opened 2025-03-17
feat(groupByProp): add type-safe grouping function with tests

### LINKED ISSUE #1057: Improve `groupBy` to more tightly type tagged unions
## Improving Type Safety in `groupBy` for Tagged Union Types

`remeda`'s `groupBy` function is useful for grouping arrays without using `reduce`. However, its current implementation only supports dynamic grouping and does not preserve type specificity when used with tagged union types. This results in unnecessary type assertions or runtime checks, even when the correct type should be statically inferable.

### Problem

Consider the following example where we group an array of actions by their `type` property:

```ts
type Action = 
  | { type: "create"; name: string } 
  | { type: "delete"; id: string } 
  | { type: "update"; id: string; changes: Record<string, unknown> };

const actions: Action[] = [
  { type: "create", name: "Item A" },
  { type: "delete", id: "2" },
  { type: "update", id: "3", changes: { name: "Updated Name" } }
];

const grouped = groupBy(actions, x => x.type);

// ❌ The returned groups lose the connection between `type` and its corresponding shape
grouped.delete?.map(action => action.id);  // ❌ No type narrowing is performed, leading to a type error
```

### Expected Behavior

When grouping an array of objects by a discriminant property, the resulting groups should maintain their type integrity. Specifically:

- The `delete` group should contain only `{ type: "delete"; id: string }` objects.
- The `create` group should contain only `{ type: "create"; name: string }` objects.
- The `update` group should contain only `{ type: "update"; id: string; changes: Record<string, unknown> }` objects.

### Actual Behavior

Currently, `groupBy` infers each group as a general array containing all possible members of the union. This forces developers to rely on runtime checks or manual type assertions.

### Proposed Solution

To improve type inference, `groupBy` should provide an overload that preserves the static relationship between the key and its corresponding object shape:

```ts
// Data First version
export function groupBy<T extends Record<PropertyKey, unknown>, Key extends keyof T>(
  data: T[], 
  key: Key
): { 
  [K in T[Key] & PropertyKey]: NonEmptyArray<Extract<T, { [_ in Key]: K }>> | undefined 
};
// Data Last version
export function groupBy<Key extends PropertyKey>(
  key: Key
): <T extends Record<Key, unknown>>(data: T[]) => {
  [K in T[Key] & PropertyKey]: NonEmptyArray<Extract<T, { [_ in Key]: K }>> | undefined;
};
```

This modification enables `groupBy` to correctly infer types when used with a tagged union, as demonstrated below:

```ts
const grouped = groupBy(actions, "type");

/* 
  Correctly inferred type: 
  { 
    create: NonEmptyArray<{ type: "create"; name: string }> | undefined; 
    delete: NonEmptyArray<{ type: "delete"; id: string }> | undefined; 
    update: NonEmptyArray<{ type: "update"; id: string; changes: Record<string, unknown> }> | undefined; 
  } 
*/

// ✅ Type-safe access
grouped.create?.map(action => action.name);  // OK
grouped.create?.map(action => action.id);  // ❌ Type error as expected
grouped.update?.map(action => action.id);  // OK
```

Conclusion

By enhancing groupBy with better static type inference, we can improve the developer experience, reducing the need for manual type assertions and ensuring safer, more predictable code. Since this proposal adds an overload rather than modifying existing behavior, it does not introduce breaking changes to existing type definitions.

However, there are some points that need further consideration:

Implementation Complexity: This change requires additional branching in the implementation to handle the new overload.

Type Completeness: While the proposed type signature generally works as expected, it currently includes never properties for some cases. Further refinement is necessary to make it fully robust.

I would love to hear feedback from the community on this proposal. Are there alternative approaches that could achieve the same goal with fewer trade-offs? Any insights or concerns would be greatly appreciated.

bmthd:
Fixes #1057 

The concern is that the implementation type definitions are loose.
Any advice on direction would be appreciated.

---

Make sure that you:

---

### Available Types:

existing).

</details>


packages/remeda/src/groupBy.test-d.ts
```diff
@@ -124,3 +124,34 @@ describe("filtering on undefined grouper result", () => {
     expectTypeOf(rest).toEqualTypeOf({} as const);
   });
 });
+
+test("infer type of groupBy", () => {
```
cjquines, line 128: **suggestion:** let's put this under its own `describe` block, for static groupby specifically

and let's call this test `infers type of groups`

cjquines: i'd like to add more type tests to the static groupby as well, but i'm not sure what. what do you think?

packages/remeda/src/groupBy.test.ts
```diff
@@ -3,7 +3,28 @@ import { pipe } from "./pipe";
 import { prop } from "./prop";
 
 describe("data first", () => {
-  test("groupBy", () => {
+  test("groupBy static", () => {
```
cjquines, line 6: **suggestion:** let's make a separate describe block for the static version, as it seems different enough

cjquines (commented):
**thought:** should this belong to the same function? it has a different signature...

because of how we generate our docs, as well, this wouldn't appear in the docs site with the `key` signature

maybe we should call this `keyBy` and make this a new function, i'm not sure. i'd get a second opinion here. @eranhirsch maybe

eranhirsch:
I also think this might make more sense using a different name. Unlike Lodash/Ramda, we prefer functions to be differentiated mainly by name and not to provide overloaded variants with considerably different signatures.

As for the name, I think `groupByProp(data, propName)` makes sense because it replaces the idiomatic `groupBy(data, prop(propName))`. _groupByKey_ seems wrong because the key refers to the prop's (`[name, value]`) name part, but here the value is meaningful to the logic. If we want to be lavish and diverge from conventions, we can call this `pivot` or `pivotOn` e.g,. https://en.wikipedia.org/wiki/Pivot_table. Names are hard.

eranhirsch (changes requested):
Let's split this out of groupBy.

bmthd:
Thank you both for your feedback!
I also think it would be more natural to extract it into a separate function.
If there is no need to force commonality, I think it would be possible to infer the type as a stable tuple type rather than an indeterminate array, so please let me reimplement it.
I will contact you again once I have implemented it.

bmthd:
Thank you for your patience.
We have implemented groupByProp, which has been separated as a separate function.
The changes from the previous implementation are as follows
- The type to be received is now limited to Record.
- Split by whether or not it is a readonly Record, and separate static types from non-static types.
- For static types, all arrays are fully inferred by covering arrays.

I think there are still some tests that I have not taken into account, so if you have any suggestions, please let me know.

eranhirsch:
Hey, I just wanted to ping this PR to say that I'm really sorry I'm not getting to review it. It's been a really busy month. I hope to get to this next week.

eranhirsch:
Hey @bmthd! I'm sorry this got delayed so long. I had a baby in February and haven't had time to review this PR since then. Rebasing this on the latest main is showing this PR is failing type tests for TypeScript 5.1 and 5.2. I want to get this PR shipped because it's an important and valuable addition to the library. So, are you willing to continue working on this together through review(s), or do you prefer I finish the work by myself? I know people are busy, and getting back to notifications on GitHub isn't always practical, so I'll assume that you aren't interested if you don't reply in a week.

Again, I'm sorry that this PR wasn't handled as quickly as I aimed for contributions to the library, and I totally understand if your focus has shifted elsewhere.

bmthd:
@eranhirsch 
Hi, thanks for getting back to me and congratulations on the new addition to your family!

I’d love to continue contributing to this, but to be honest, I’m not sure what’s causing the type failures on TypeScript 5.1 and 5.2. I’m happy to try digging into it, but if you already have an idea of the problem or if it’s more efficient for you to take over, I’d be totally fine with that too.

Let me know what you think would work best!

eranhirsch:
I have already started chipping away at the problem; I'm sending a few PRs that are related to this type but could be split out. Once those are shipped, I'll push an update to this PR, which should simplify it and also fix the TypeScript issues. You can help me out with this PR by reviewing it, and we'll ship it together.

packages/remeda/src/groupByProp.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }
+
+  // Set the prototype as if we initialized our object as a normal object (e.g.
+  // `{}`). Without this none of the built-in object methods like `toString`
+  // would work on this object and it would act differently than expected.
+  Object.setPrototypeOf(output, Object.prototype);
+
+  // @ts-expect-error [ts2322] -- This is fine! We use a broader type for output while we build it because it more accurately represents the shape of the object *while it is being built*. TypeScript can't tell that we finished building the object so can't ensure that output matches the expected output at this point.
+  return output;
```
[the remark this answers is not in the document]
eranhirsch: We prefer suppressing ts errors instead of casting so that if future versions of TypeScript infer these the way we want, we can remove the suppressions because suppression for non-errors are errors themselves. Casts often hide type errors and might not surface changes in the typing as effectively.

eranhirsch:
Hey @bmthd, Thanks again for all your work getting this utility added. The final code is changed a lot from your version, but it was your version that got me thinking about how we handle other types as well; mainly the idea that we could filter an array by looking at `Record<Prop, Value>` sparked the changes in #1087, and the usage of NonEmptyArray (and @cjquines's #863) drove me to do #1085. In addition, #1080 had some perf optimizations to the runtime implementation that needed to be ported back to this.

I'm really excited about this function because it does a lot of really cool things in a really tight package that would be useful to a lot of projects.

Please feel free to comment, ask questions, and review the PR.

packages/remeda/src/groupByProp.test.ts
cjquines, line 1: **question:** can we add a test against #1046 ?

cjquines (dismissed):
i would revise the groupByProp documentation if i had more time, but it's not blocking

packages/remeda/src/groupByProp.test.ts
eranhirsch: done!

PR #1062, opened 2025-03-19
feat: Overhaul of lazy pipelines

johnw42:
This PR replaces the logic used for lazy pipelines with a much simpler system based on generator functions. All lazy functions have been rewritten as generators. 

Some additional related changes are included:
- Two new functions, `prepend` and `unfold`, have been added to demonstrate how the new system allows lazy functions that were previously impossible to implement.
- The `isIterable` function from `pipe.ts` has been made public.
- The return type of `toArray` has been refined; this change is used in the new lazy implementations.
- The three-argument signature of `purry` has been removed, as it is no longer used in lazy function implementations, and it doesn't appear it was ever meant to be public.
- The tests for `map` and `mapFeedback` have been altered to account for how `pipe` now passes its first argument directly to the first function in the pipeline.
- Various function are now marked as `@lazy`.


johnw42:
Assuming this PR is accepted, I have another one that builds on this one by changing all functions that accept array arguments to accept arbitrary iterables instead.

cjquines:
seems good so far (and i've always wanted to do this rewrite, so thank you so much!!)

i'll take a closer look at this when i get more time, but i'm gonna start with:

* splitting up the `prepend` and `unfold` PRs to a separate PR; adding new functions should get a separate review process
* addressing the coverage issues highlighted by codecov
* investigating the failing CI tests for typescript 5.1—we say we support earlier versions of ts, so we should get those passing
  * it's possible the answer to this is "delete some tests"
* is `isArray` necessary, and can it not be replaced with `Array.isArray`?

johnw42:
I addressed all the issues you pointed out; looks like there are some Netlify tests failing but I assume that's an infrastructure issue and not a problem with the PR.

The only thing I want to push back on is using R.isArray. Yes, it's possible to use Array.isArray, but I'd have to add more type casts, and that seems silly to me when R.isArray does exactly the type checking I need. I've played around with benchmarking before and found node to be very good at inlining trivial functions, but if you're worried about performance hit, it should be possible to eliminate it by redefining R.isArray like this:

```ts
export const isArray = Array.isArray as <T>(
  data: ArrayLike<unknown> | T,
) => data is Extract<T, ReadonlyArray<unknown>> extends never
  ? T extends Iterable<infer E>
    ? ReadonlyArray<E>
    : ReadonlyArray<unknown>
  : IsAny<T> extends true
    ? ReadonlyArray<unknown>
    : Extract<T, ReadonlyArray<unknown>>;
```

cjquines:
i'm not worried about the peformance hit or the bundle size, it's more of a "spiritually we should avoid imports unless needed" lol

the netlify things may or may not be related to the docs ci currently failing?

packages/remeda/src/mapWithFeedback.test.ts
```diff
@@ -90,10 +90,10 @@ describe("data last", () => {
...
     expect(lazyItems).toStrictEqual([
-      [1],
-      [1, 2],
-      [1, 2, 3],
-      [1, 2, 3, 4],
+      [1, 2, 3, 4, 5],
+      [1, 2, 3, 4, 5],
+      [1, 2, 3, 4, 5],
+      [1, 2, 3, 4, 5],
       [1, 2, 3, 4, 5],
```
cjquines, line 97: **issue:** hm, this is a breaking change, no?

cjquines (commented):
more high-level comments, though i'm less sure about these ones

- `doTranduce` etc. are default exports—i think for consistency we should not be doing that? although i've less strong of an opinion (and maybe we should have a lint rule if we shouldn't be doing that shrug)
- is the plan to rewrite tests with `describeWithPipe` &c? i'm not sure how i feel about that

packages/remeda/src/mapWithFeedback.test.ts
johnw42: I'm fine with avoiding default exports. I'm actually not that big of a fan of them, but it seemed like if there was ever a case where they're appropriate, it's when you're exporting a single symbol with the same name as the file.

For rewriting the tests, what I've done (outside this PR) is rewrite them in terms of [`describeIterableArg`](https://github.com/remeda/remeda/blob/5ebbb3bf9af5ea384cdff69b0c093c7b3a4f0dea/packages/remeda/src/internal/describeIterableArg.ts), where the tests suites go from something like this:

```ts
test("data first", () => {
   expect(foo([1,2,3])).toBe(42);
});
```

...to this:

```ts
describeIterableArg("foo", ({wrap}) => {
    test("data first", () => {
       expect(foo(wrap([1,2,3]))).toBe(42);
   });
});
```

Pros:
- Every existing test can be upgraded without duplication by wrapping array inputs with `wrap`.
- We check that functions work with both arrays and non-array Iterables.
- The Iterable returned by the `wrap` function ensures each iterable is only traversed once.
- With an optional argument to wrap, we can also limit the number of items it can be asked for before throwing an error, making it trivial to test for laziness in many cases.
- Putting all tests in a big single `describe` block makes it easy to run all tests in a file with a single click.

Cons:
- Calling `wrap` everywhere is ugly, and failing to call it can make the test pass when it shouldn't.
- An extra level of indentation gets added, making changes in those tests harder to review (but only once).
- The unit tests no longer serve as implicit type tests for fancy tuple-based return types, because the tests only demonstrate calling a signature that accepts an Iterable, and those signatures usually can only return a plain Array<...> type.
- The tests look kind of funny and need an tweak to the linter rules to avoid warnings.

Regarding the test change you pointed out, it is indeed a potentially breaking change, although I'd be *really* surprised if any actual code breaks from the array being longer than expected. I don't see it as a big deal because the existing behavior isn't exactly documented, and there's already an inconsistency between a function called directly (where the original input is used for the third argument) and functions called in a lazy context (where the third argument is a fresh array containing only items seen so far). I could change the behavior to return the shorter array, but that has its own problems, but I'd rather not because it forces a new array to be created when the original input array is usable as-it.

Overall I think the third argument to callbacks is pretty unfortunate, because it requires a temporary array to by created for every call with an Iterable argument, even though most callers don't use that argument at all. IMHO any design involving laziness will break some users' expectations, because that callback pattern has traditionally only been used for array methods where that argument holds a reference to the input array. I considered checking the `length` field of the callback and omitting the last argument when `length` < 3, but that breaks when the callback uses varargs because they're not represented in `length` or any other function field.

While we're talking about breaking changes, I'd like to call out the change I made to `purry`, eliminating its third argument. That argument is technically a documented feature right now, but there's nothing in the documentation about how to pass it correctly, so anyone actually using it is already relying on undocumented behavior. If you feel strongly about it, I think I could add that parameter back and make it translate an old-style `LazyEvaulator` function to use the new style. Just depends on how much you want to support users who rely on undocumented features.

eranhirsch:
Hey, sorry for being quiet; I plan on going over this PR, too, hopefully, this upcoming week. This is a big change and might benefit from considering releasing under a v3 release where we can also drop some older typescript versions and include ES2023 in our target. I was planning on tackling this (and additional changes) for that version and was planning to coincide it with the release of node 24 and the EOL for node 18 around early May.

eranhirsch:
Hey, just pinging this PR to say that I'm sorry I haven't gotten to it already. It's a big PR and it touches on a few things which require extra consideration.

I'll say up front that I don't *think* i'll be able to approve this PR because it may cause issues for users and I don't have time to support those cases. It also doesn't align nicely with what I think is the right way to solve the problem you are trying to solve; which is to move lazy logic **out** of the pipe, so that we are in a better place when native JS pipes are introduced into the language.

Anyway, that won't happen within a minor version update of the package, and I don't have the capacity right now to facilitate a major version update of the language.

On a practical level, I can say this:

* It would be easier to review parts of this PR that could be "cut out" of the bigger PR. You moved the "isIterable" function; that's great! Send a PR just with that change.

* Don't add new testing utils in the same PR that change the functions they test because that makes it impossible to trust the tests, and we are left blind to possible breakages.

* To make the PR even more manageable, I would create an alternative pipe implementation, show that it works well, and then plan to migrate all utilities to it. That would mean that the whole PR could be 3-4 files that are much easier to review.

* There are some implementation issues I've already noticed, e.g., you use a Class for the logic, although you create an instance and then run its only method immediately and then throw the instance away; that's a wasteful way to implement this, and makes the code harder to read or understand. Keep your changes as simple as possible.

```ts
function foo(<params>, <moreParams>) {
  const instance = new MyClass(<params>);
  const result = instance.doSomething(<moreParams>);
  return result;
}
```

This is equivalent to just putting the whole class logic in foo.

eranhirsch (changes requested):
Hey, just pinging this PR to say that I'm sorry I haven't gotten to it already. It's a big PR and it touches on a few things which require extra consideration.

I'll say up front that I don't *think* i'll be able to approve this PR because it may cause issues for users and I don't have time to support those cases. It also doesn't align nicely with what I think is the right way to solve the problem you are trying to solve; which is to move lazy logic **out** of the pipe, so that we are in a better place when native JS pipes are introduced into the language.

Anyway, that won't happen within a minor version update of the package, and I don't have the capacity right now to facilitate a major version update of the language.

On a practical level, I can say this:

* It would be easier to review parts of this PR that could be "cut out" of the bigger PR. You moved the "isIterable" function; that's great! Send a PR just with that change.

* Don't add new testing utils in the same PR that change the functions they test because that makes it impossible to trust the tests, and we are left blind to possible breakages.

* To make the PR even more manageable, I would create an alternative pipe implementation, show that it works well, and then plan to migrate all utilities to it. That would mean that the whole PR could be 3-4 files that are much easier to review.

* There are some implementation issues I've already noticed, e.g., you use a Class for the logic, although you create an instance and then run its only method immediately and then throw the instance away; that's a wasteful way to implement this, and makes the code harder to read or understand. Keep your changes as simple as possible. e.g.

```ts
function foo(<params>, <moreParams>) {
  const instance = new MyClass(<params>);
  const result = instance.doSomething(<moreParams>);
  return result;
}
```

This is equivalent to just putting the whole class logic in foo.

PR #1066, opened 2025-03-24
fix(groupBy): don't use Map internally

eranhirsch:
Reimplementing #1049 by reverting the usage of Map and opting for a safer method of accessing a prop on an object instead.

I'm a little concerned about using a map because we then need to create an additional object (via the fromEntries) whereas this solution only manages a single object. I still want to benchmark the 2 solutions to see if this makes any sense.

### DIFF
--- a/packages/remeda/src/groupBy.bench.ts
+++ b/packages/remeda/src/groupBy.bench.ts
@@ -1,0 +1,292 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type, @typescript-eslint/no-magic-numbers */
+
+import { bench } from "vitest";
+import type { ExactRecord } from "./internal/types/ExactRecord";
+import type { NonEmptyArray } from "./internal/types/NonEmptyArray";
+import { identity } from "./identity";
+import { times } from "./times";
+import { randomInteger } from "./randomInteger";
+import { sample } from "./sample";
+import { shuffle } from "./shuffle";
+import { constant } from "./constant";
+
+const SYMBOLS = times(100, (i) => Symbol(`sym${i.toString()}`));
+const CHARACTERS = [
+  // eslint-disable-next-line @typescript-eslint/no-misused-spread
+  ..."ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789",
+];
+const REGIONS = ["NA", "EU", "APAC", "LATAM", "MEA", "OCE", "AFRICA", "ARCTIC"];
+
+describe.each([
+  {
+    name: "groupBy with small arrays",
+    array: times(10, identity()),
+    grouper: (x: number) => x % 3,
+  },
+  {
+    name: "groupBy with medium arrays",
+    array: times(1000, identity()),
+    grouper: (x: number) => x % 10,
+  },
+  {
+    name: "groupBy with large arrays",
+    array: times(10_000, identity()),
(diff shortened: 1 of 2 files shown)


eranhirsch:
Result of benchmark:

```
  Map - src/groupBy.bench.ts > 'groupBy with small arrays'
    1.60x faster than Object

  Map - src/groupBy.bench.ts > 'groupBy with medium arrays'
    2.55x faster than Object

  Map - src/groupBy.bench.ts > 'groupBy with large arrays'
    2.56x faster than Object

  Map - src/groupBy.bench.ts > 'groupBy with few groups'
    3.21x faster than Object

  Object - src/groupBy.bench.ts > 'groupBy with many groups'
    2.26x faster than Map

  Map - src/groupBy.bench.ts > 'groupBy with string keys'
    1.48x faster than Object

  Map - src/groupBy.bench.ts > 'groupBy with symbol keys'
    1.01x faster than Object

  Map - src/groupBy.bench.ts > 'groupBy with problematic property keys'
    3.57x faster than Object

  Map - src/groupBy.bench.ts > 'users by region (real-world scenario)'
    1.58x faster than Object

  Map - src/groupBy.bench.ts > 'Random integers, small'
    2.32x faster than Object

  Map - src/groupBy.bench.ts > 'Random integers, medium'
    2.15x faster than Object

  Map - src/groupBy.bench.ts > 'Random integers, large'
    1.92x faster than Object

  Map - src/groupBy.bench.ts > 'Random distribution, few groups'
    2.64x faster than Object

  Map - src/groupBy.bench.ts > 'Random distribution, many groups'
    1.72x faster than Object

  Object - src/groupBy.bench.ts > 'keys with variable lengths, short'
    2.26x faster than Map

  Object - src/groupBy.bench.ts > 'keys with variable lengths, long'
    2.22x faster than Map

  Object - src/groupBy.bench.ts > 'Random symbol keys'
    2.23x faster than Map

  Map - src/groupBy.bench.ts > 'Random problematic property keys'
    3.20x faster than Object

  Map - src/groupBy.bench.ts > 'Random realistic data, by region'
    1.46x faster than Object

  Map - src/groupBy.bench.ts > 'Random realistic data, by age range'
    1.73x faster than Object

  Map - src/groupBy.bench.ts > 'mixed data types as keys'
    1.05x faster than Object

  Map - src/groupBy.bench.ts > 'non-contiguous memory access patterns'
    1.53x faster than Object

  Map - src/groupBy.bench.ts > 'Random undefined returns in the callb…'
    1.69x faster than Object

  Map - src/groupBy.bench.ts > Large data with GC pressure
    1.37x faster than Object

  Run 3 with fresh data - Map - src/groupBy.bench.ts > Cold start vs warm cache
    1.16x faster than Run 2 with fresh data - Map
    1.60x faster than Run 1 with fresh data - Map
    3.67x faster than Run 1 with fresh data - Object
    3.74x faster than Run 3 with fresh data - Object
    3.78x faster than Run 2 with fresh data - Object
```

eranhirsch:
The object-based implementation is better when the **number of keys is high** and maybe when the key is long, but otherwise, the map-based implementation is better for most real-life use cases, especially when grouping on a **low cardinality prop**.

I'm abandoning this PR.

cc: @cjquines, this was an interesting case to explore. I didn't have a good hunch ahead of doing this about what would be the better implementation. We should remember these results for other utilities and future changes; the built-in objects (Map, Set) have some performance optimizations that could be taken advantage of.

PR #1074, opened 2025-04-09
fix(mapValues): preserve number keys in the resulting object

### LINKED ISSUE #1071: Hard to use numeric keys after `mapValues`
When passing an object with numeric keys, like `Record<number, unknown>`, to the `mapValues` function, key type becomes `` `${number}` ``. Because of this, it becomes difficult to use numeric keys to index a resulting object:

- plain `number`s lead to a typing error
- `String(number)` returns plain `string`, which doesn't satisfy `` `${number}` `` type

Simple reproduction: https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wEpwGZQiOAcigFMQSATZQgbgCg6BjCAO1XggCMArALkxLMoFADwsAriE4koAGjjsowFgHMAfHAC8cAN4Bfel24BtAIwBdLUWwQItOkeMAmS9sKdkUe3QD0PuADu0ADWqEysqBAANiQAdFEQKgAUjhbyji4AlPThbPAgyGBglFYYsQVgAGrIUeIkqCk88kkAbjWZWhptUbEwEACqRTIAwmgkSZnZDH5wMnhycJzi8KgAFhDiURSBIXDKcFDiLDCgJLmRMfGJSRXFFGbm8reUzubZQA

kosmotema:
Removed converting `number` keys to  the`` `${number}` `` type

Resolves #1071

---

Make sure that you:

---

### Available Types:

existing).

</details>

### DIFF
--- a/packages/remeda/src/mapValues.test-d.ts
+++ b/packages/remeda/src/mapValues.test-d.ts
@@ -118,6 +118,18 @@ test("number keys are converted to string in the mapper", () => {
   });
 });
 
+test("number keys are preserved in the resulting object", () => {
+  const data = {} as Record<number, unknown>;
+
+  const dataFirst = mapValues(data, constant(true));
+
+  expectTypeOf(dataFirst).toEqualTypeOf<Record<number, boolean>>();
+
+  const dataLast = pipe(data, mapValues(constant(true)));
+
+  expectTypeOf(dataLast).toEqualTypeOf<Record<number, boolean>>();
+});
+
 test("maintains partiality", () => {
   const result = mapValues(
     {} as { a?: number; b?: string; c: number; d: string },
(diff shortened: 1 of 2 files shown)


packages/remeda/src/mapValues.test-d.ts
```diff
@@ -118,6 +118,18 @@ test("number keys are converted to string in the mapper", () => {
   });
 });
 
+test("number keys are preserved in the resulting object", () => {
```
eranhirsch, line 121: Can you add a comment with the issue number, or add the issue number in parenthesis to the test name, so it's easy for us to find context in the future if something breaks.

eranhirsch (approved):
This is great! I thought it would be much more complex but I'm happy to see that the return type is contained within this single function!

PR #1076, opened 2025-04-22
fix(takeWhile): improve type inference

alexvuka1:


cjquines:
hi @alexvuka1 thanks for the PR! can you add a test in`takeWhile.test-d.ts` that used to be failing?

while we're at it, can you do a similar thing for e.g. `dropWhile`, `dropLastWhile`, `takeLastWhile`?

alexvuka1:
> hi @alexvuka1 thanks for the PR! can you add a test intakeWhile.test-d.ts that used to be failing?

which one is that?

> while we're at it, can you do a similar thing for e.g. `dropWhile`, `dropLastWhile`, `takeLastWhile`?

done

cjquines:
oh sorry, i meant to say, "can you add a test to `takeWhile.test.d-ts` for this change, such that the test fails before your change, but now passes after your change?"

cjquines:
also thanks for working on the other ones too!

alexvuka1:
> oh sorry, i meant to say, "can you add a test to `takeWhile.test.d-ts` for this change, such that the test fails before your change, but now passes after your change?"

done

cjquines (approved):

eranhirsch:
The new typing of `dropWhile` is wrong; it flips the semantics of dropping, where if the predicate is true the item is *removed* from the array:


```ts
const result = dropWhile([1, "a"], ($) => typeof $ === "string"); //=> [1, "a"]
//    ^? ["a"]
```


https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgEyhMB1AFsANgUzgF84AzNEOAIinxH2QEMqBuAKDYGMIA7AZ3gAPOAF4UaTDgIAKANoBGADTVmAXTiM+cbvxjLpAEgCUogHxwYATzD4IpOAdEixVAVGA8A5lSPsA9H5wQXAAegD8QA

https://codesandbox.io/p/sandbox/d88t6c

PR #1082, opened 2025-05-04
chore: publish to jsr

han-tyumi:
It would be nice to be able to pull this package down from [JSR](https://jsr.io/).
For me, JSR is much more ergonomic when it comes to ECMAScript Modules and TypeScript.
This is especially nice for those developing with [Deno](https://deno.com/) which has similar ergonomics.

Some additional configuration on their website would be needed to link this repository there.
See https://jsr.io/docs/publishing-packages#publishing-from-github-actions.

It seems there is already something registered at https://jsr.io/@remeda/remeda.
So maybe this is already planned, but I didn't see anything related to JSR in the repo.

See https://jsr.io/docs/why & https://jsr.io/docs/publishing-packages for more info.

### DIFF
--- a/.github/workflows/publish-jsr.yml
+++ b/.github/workflows/publish-jsr.yml
@@ -36,8 +36,7 @@ jobs:
 
       - name: 📝 Rewrite package version
         working-directory: ./packages/remeda
-        run: |
-          sed -i 's/"version": "[^"]*"/"version": "${{ github.event.release.tag_name || github.event.inputs.tag }}"/' jsr.json
+        run: jq --arg version "${{ github.event.release.tag_name || github.event.inputs.tag }}" '.version = $version' jsr.json | sponge jsr.json
 
       - name: 🚀 Publish
         run: npm run --workspace=remeda jsr:publish


.github/workflows/ci-library.yml
```diff
@@ -170,7 +173,9 @@ jobs:
         run: npm run --workspace=remeda lint:build
 
       - name: 🚀 Release
-        run: npm run --workspace=remeda release
+        run: |
+          npm run --workspace=remeda release
+          npm exec --workspace=remeda jsr publish --allow-slow-types
```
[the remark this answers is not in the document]
cjquines: that's strange, these aren't part of the public API? like, they're literally not exported

packages/remeda/jsr.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "https://jsr.io/schema/config-file.v1.json",
+  "name": "@remeda/remeda",
```
cjquines, line 3: i don't own this scope. i don't know who does

eranhirsch: I created it a while back but couldn't get publishing to JSR to work.

I've invited you to the project, so you should be able to publish it.

FWIW, we also publish to deno.land/x, if I remember correctly, it works via a webhook: https://deno.land/x/remeda@v2.21.3

eranhirsch (dismissed):
One of the issues I faced in the past when I tried publishing to JSR was how to add the dependency on type-fest properly. I just couldn't get it to work. I see that this PR doesn't have any special config/code that handles type-fest, so I'm assuming that was either fixed by deno/jsr, or that I was mistaken about the requirement. Is there a way for us to test that the published package is working as intended without causing issues for users?

PR #1085, opened 2025-05-12
fix(hasAtLeast): handle more tuple/array shapes

eranhirsch:
Reimplement the type using the new TupleParts. This includes fixes to tuple parts and additional type tests for both.

The type itself was pulled out of the hasAtLeast utility because it would be reused in the implementation of `groupByProp`.

I also refactored the RemedaErrorType to make it easier to use and to use `never` as the underlaying type instead of a symbol so that types that rely on it propagate the "neverness" of the error cases.

### DIFF
--- a/packages/remeda/src/internal/types/ArrayRequiredPrefix.ts
+++ b/packages/remeda/src/internal/types/ArrayRequiredPrefix.ts
@@ -37,7 +37,7 @@ export type ArrayRequiredPrefix<
                 IsNever<TupleParts<T>["item"]>
               > extends true
             ? RemedaTypeError<
-                "ArraySetRequired",
+                "ArrayRequiredPrefix",
                 "The input tuple cannot satisfy the minimum",
                 [T, MinRequiredItems]
               >
@@ -79,13 +79,17 @@ export type ArrayRequiredPrefix<
                 ]
               >
         : RemedaTypeError<
-            "ArraySetRequired",
+            "ArrayRequiredPrefix",
             "Remainder didn't compute to a number?!",
             [T, MinRequiredItems]
           >
-      : RemedaTypeError<"ArraySetRequired", "Failed to distribute union?!", T>
+      : RemedaTypeError<
+          "ArrayRequiredPrefix",
+          "Failed to distribute union?!",
+          T
+        >
     : RemedaTypeError<
-        "ArraySetRequired",
+        "ArrayRequiredPrefix",
         "Only literal minimums are supported!",
         MinRequiredItems
       >;
--- a/packages/remeda/src/internal/types/ArrayRequiredPrefix.ts
+++ b/packages/remeda/src/internal/types/ArrayRequiredPrefix.ts
@@ -10,14 +10,15 @@ import type { CoercedArray } from "./CoercedArray";
 import type { RemedaTypeError } from "./RemedaTypeError";
 import type { TupleParts } from "./TupleParts";
 import type { IterableContainer } from "./IterableContainer";
+import type { ClampedIntegerSubtract } from "./ClampedIntegerSubtract";
 
 export type ArrayRequiredPrefix<T extends IterableContainer, N extends number> =
   IsLiteral<N> extends true
     ? // Distribute T to support union types
       T extends unknown
       ? ClampedIntegerSubtract<
-          ClampedIntegerSubtract<N, TupleParts<T>["required"]["length"]>,
-          TupleParts<T>["suffix"]["length"]
+          N,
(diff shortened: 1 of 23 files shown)


packages/remeda/src/internal/types/TupleParts.ts
```diff
@@ -1,61 +1,127 @@
...
-            prefix: [...PrefixRequired, ...Partial<PrefixOptionals>];
-            required: PrefixRequired;
-            optional: PrefixOptionals;
+            /**
+             * A *trivial* tuple that defines the part of the tuple where all
+             * its elements are required. This will always be the first part of
+             * the  tuple and will never contain any optional or rest elements.
+             * When the array doesn't have a required part this will be an
+             * empty tuple (`[]`).
+             */
```
cjquines, line 56: **thought:** i'm surprised that these docstrings show on hover!

packages/remeda/src/internal/types/RemedaTypeError.ts
```diff
@@ -1,22 +1,12 @@
...
-  : RemedaTypeError<
-      FunctionName,
-      `Internal error ${Message}. Please open a Remeda GitHub issue.`
-    >;
+  // Type errors don't have stack traces so we need good strings to be able to
+  // grep for them and debug them.
+  Name extends string,
+  Message extends string,
+  Metadata = never,
+> = Tagged<never, `RemedaTypeError(${Name}): ${Message}.`, Metadata>;
```
cjquines, line 12: **note:** i remember that the reason i didn't use `never` was because it swallows the type error. in the case of `partialBind` we use it for a user-facing error, so we don't want this

we also don't always want `never` behavior i think? consider the following

```ts
import type { RemedaTypeError } from "./RemedaTypeError";

declare function maybeNumber<T>(): T extends string
  ? RemedaTypeError<"test", "this is a library error", T>
  : number;

it("acts as never in conditionals", () => {
  maybeNumber<string>() satisfies number;
});
```

with old behavior this produces a type error (as expected)

image

with new behavior there's no error

cjquines: do we want two types here? RemedaTypeError and RemedaInternalTypeError?

packages/remeda/src/internal/types/ArrayRequiredPrefix.ts
```diff
@@ -0,0 +1,149 @@
...
+        Minuend,
+        Subtrahend,
+        [...SubtrahendBag, unknown],
+        ResultBag
+      >;
+
+// Instead of using type-fest utilities to compute the index for pivoting, and
+// then using ArraySlice twice and spreading the result; we implement a much
+// simpler iterative solution to make the beginning of the optional part
+// required and the rest of it optional.
```
cjquines, line 133: **praise:** oh, this is indeed simpler

packages/remeda/src/internal/types/ArrayRequiredPrefix.ts
```diff
@@ -0,0 +1,149 @@
...
+// Notice that we make the decision based on T, but output the type based on U.
+type WithSameReadonly<Source, Destination> =
+  IsEqual<Source, Readonly<Source>> extends true
+    ? Readonly<Destination>
+    : Destination;
+
+// We built our own version of Subtract instead of using type-fest's one because
+// we needed a simpler implementation that isn't as prone to excessive recursion
+// issues and that is clamped at 0 so that we don't need to handle negative
+// values using even more utilities.
```
cjquines, line 101: **thought:** this is a useful type; i'm going to pull it out for #863

packages/remeda/src/internal/types/TupleParts.ts
eranhirsch: 🥳

eranhirsch: Just noticed that still uses the term trivial tuple but I changed that to fixed tuple. I'll change it here too

packages/remeda/src/internal/types/RemedaTypeError.ts
eranhirsch: It's the difference between "input" types and "output" types. When the type is for output, we want `never` because we want to effectively remove cases from the tree of possible output values, but when the type is used for input, we want a breaking type that can't be satisfied so that the user can't continue and gets the error instead.

I needed the `never` option because ArrayRequiredPrefix is used in the output type of a type-guard, and it needs to extend the input type to work (e.g., `(x: string | number): x is boolean` is an error).

I have an idea of how to fix this.

cjquines (approved):
good fix!

PR #1114, opened 2025-06-04
feat(conditional): support default case fallbacks directly within the implementation

eranhirsch:
I found a way to stay backward-compatible and still add the default/fallback functionality directly into the function implementation. This allows users to avoid the need to use the boilerplate `[constant(true), ...]` for the last case.

---

Old description:

`defaultCase` was added when the utility was added in v1 to help migrate from lodash and ramda where there is an implicit fallback default case that returns undefined; 

The remeda version throws when no matching cases are found, which in turn allows us to remove the `| undefined` from the return, even when the user doesn't want or need it. This aligns with our preference in Remeda to surface errors at compile time rather than relying on implicit fallbacks.

`defaultCase` was seen as a way to offer a simpler API for handling this last case. This was before we offered `constant` utility, which makes writing the default case considerably easier.

`conditional` is the only function in Remeda that isn't exported as a proper function. This breaks our docs site (which is currently missing an entry for it) and also requires us to play with the types to make JSR.io happy.

The cost outweighs the benefits!

I've added a deprecation warning and a TODO to remove the utility in 3 months. If we release v3 before that, it would be part of that migration too. However, if we do not ship v3 soon, we will still follow up with a removal within the v2 cycle, as it only changes a single utility slightly and doesn't require a full major version bump.

### DIFF
--- a/.github/workflows/ci-library.yml
+++ b/.github/workflows/ci-library.yml
@@ -260,54 +260,63 @@ jobs:
       id-token: write
 
     steps:
-      - name: 🫥 Extract Missing Version
-        id: extract-missing
+      - name: 🏷️ Repo Latest Released Version?
+        id: latest-released
         run: |
-          # Get the latest version tag
+          set -x
           VERSION=$(
-            git ls-remote --tags https://github.com/${{ github.repository }}.git \
-            | grep --only-matching --extended-regexp 'v[0-9]+\.[0-9]+\.[0-9]+$' \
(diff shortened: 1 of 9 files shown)


cjquines (approved):

eranhirsch:
@cjquines - Re-requesting review following a major change in the impl.

cjquines (approved):
very nice, and probably better performing too :)

Issue #1075, opened 2025-04-18
pickBy: adds undefined for unbounded records


cassus:
When I try to filter out `null` values from a record type, the `null`s turn into `undefined`s in the type

```ts
const input: Record<string, number | null> = {
  a: 1,
  b: null,
}

// problem: {   [key: string]: number | undefined;} ❗
const problem = pickBy(input, (val) => val !== null)

// const good: { [key: string]: number | null;} ✅
const good = pickBy(input, (val) => val !== 1)
```

Tested with version 2.21.3

rakyi:
`pick` also adds `undefined` to the values type when the keys are not known at compile time.

```ts
import * as R from 'remeda';

const obj: Record<number, number> = {};
for (let i = 0; i < 10; i++) {
    obj[i] = i * 10;
}

const x = R.pick(obj, [1, 5, 5]);
// Type of x stays Record<number, number>.
console.log(x)

const keys = []
for (let i = 0; i < 10; i++) {
    if (i % 2 === 0) {
        keys.push(i)
    }
}

const y = R.pick(obj, keys)
// Type of y becomes Record<number, number | undefined>
console.log(y)
```

rakyi:
At least for `pick` the behavior seems to be introduced in https://github.com/remeda/remeda/pull/923.

eranhirsch:
Hey, I'm sorry it took so long to send out a fix for this, but #1127 should solve this issue, and you can check it out either by patching your project with: `npm i https://pkg.pr.new/remeda@1127` to see if this fixes it for you, or wait for that PR to ship.

@rakyi - Your reported issue is unrelated to this one, so I'll split it out into a separate issue.

PR #1127, opened 2025-06-25
fix(pickBy): dont add `undefined` to value type of unbounded records

eranhirsch:
We were handling unbounded records for the non-narrowing case, but didn't address the typing for unbounded records for the narrowing case. The solution is similar, where we need to reconstruct the record, but this time we also need to extract the narrowed type from the original type.

Fixes: #1075


cjquines (approved):
i'm wondering if there are other cases where we're running into this. `omitBy` would be the other case, but i'm not actually sure if it'd be wrong in that case?

Issue #1128, opened 2025-06-27
pick: adds `undefined` to the value types when the keys are not known at compile time


eranhirsch:
> `pick` also adds `undefined` to the values type when the keys are not known at compile time.
> 
> ```ts
> import * as R from 'remeda';
> 
> const obj: Record<number, number> = {};
> for (let i = 0; i < 10; i++) {
>     obj[i] = i * 10;
> }
> 
> const x = R.pick(obj, [1, 5, 5]);
> // Type of x stays Record<number, number>.
> console.log(x)
> 
> const keys = []
> for (let i = 0; i < 10; i++) {
>     if (i % 2 === 0) {
>         keys.push(i)
>     }
> }
> 
> const y = R.pick(obj, keys)
> // Type of y becomes Record<number, number | undefined>
> console.log(y)
> ``` 
>
> At least for pick the behavior seems to be introduced in https://github.com/remeda/remeda/pull/923.

 _Originally posted by @rakyi in [#1075](https://github.com/remeda/remeda/issues/1075#issuecomment-2826877588)_

rakyi:
Thank you for the fix and all the great work on Remeda!

thelinuxlich:
I don't understand this change, now `pick` is returning `T | undefined` where it shouldn't

eranhirsch:
reopening while we debug

eranhirsch:
@Bessonov,

This returns `undefined` due to you not having `exactOptionalPropertyTypes` enabled in your `tsconfig.json` file. Because we can't adjust types or branch them based on TypeScript settings we need to pick which settings we assume are enabled and build our types to match those settings.  We recommend turning on `exactOptionalPropertyTypes` in projects using Remeda to make sure types work as we wrote them. TypeScript is adding `| undefined` here, not us, due to how it treats optional properties.

Starting from TypeScript 5.9 this config is also recommended for all new projects: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-9.html#minimal-and-updated-tsc---init

Playground with `exactOptionalPropertyTypes` enabled:

[...]

PR #1129, opened 2025-06-30
fix(pick): Add support for unions and unbounded records

eranhirsch:
Fixes: #1128


packages/remeda/src/internal/types/If.ts
```diff
@@ -0,0 +1,9 @@
+import type { IsNever } from "type-fest";
+
+// TODO [type-fest@>=5] -- This type is shipping in type-fest v5 and standardizes all IfXXX types. We backported it so that we can already use it because v5 is shipped.
```
cjquines, line 3: v5 hasn't shipped yet, no? but yes, we should get ready for type-fest v5

cjquines (dismissed):

PR #1148, opened 2025-07-09
docs: fix alignment issue caused by search modal

### LINKED ISSUE #1147: Docs: Entire page suddenly gets aligned right
I haven't figured out why this happens, but it looks like this:

Image

The reason is that the body element has the following classes `container mx-auto`, but for some reason something sets `style="margin-right: 0px;"` on the same element. I haven't found any obvious culprits in the code base, so it might be some kind of Astro plugin.

I'll update this issue if I come across any steps to repro it.

karl-run:
When the search modal disables scrolling, it messes with the alignment of the primary content because it uses the body-tag to for its `container mx-auto` classes. Moving these to a child allows the algoria search modal to disable the scroll without messing with the content.

Resolves #1147.


eranhirsch (approved):
Thanks for helping us out!

Issue #752, opened 2024-06-26
Branded type keys with "entries" giving undesired return type


JeongJuhyeon:
Same as https://github.com/remeda/remeda/issues/684 but then for `entries`. After using entries, the keys can no longer be used in their Branded way.

```typescript
type Color = string & { __brand: "rgb(r,g,b)" };
function f(c: Color) {
   const o: Record<Color, number> = {[c]: 3};
   const e = entries(o);
   const a = e[0][0];
   // a: `${string & { __brand: "rgb(r,g,b)";}}`
}
```

cjquines:
this is correct, because we're matching [`Object.entries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) behavior, so the keys are coerced to strings. in particular, the type of a is `${Color}`

eranhirsch:
@cjquines I might be missing something in my understanding, but doesn't this issue have merit because the result of running `keys` and running `entries` on the same object should result in the same typing for the key component, and where it currently doesn't?

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgUwHYysZBnANHAa2QE9s4BfOAMyghDgCIpkRkATAQwYG4AoXmMTDI4AYQgAbaHAC8cbBmCoA5nABkiOAH0tAIygdUbAFyMoy3QAoouZbl0BKBhT68AxhFQK4AEQCCACp+soiUHGQASsgeUGwAPOJSNvKKKgB8rh5e8ADaHAC6IUSklv5BDnwA9JVwtXAAegD87p7eOTm6+YVyaIo4pYF+FbzVdXVN-GzREhzM1ACuqG4wwJ7UEBCWHkmmidAOpgBuEMBsrlQblhzDF5uOfEA

In addition, it looks like the typing for `keys` also isn't optimal when it comes to branded keys as `${Color}` is equivalent in runtime to Color, but breaks on typing.

cjquines:
hm, that's fair. i'll reopen

PR #1151, opened 2025-07-15
fix(keys,entries): support branded keys

eranhirsch:
Fix: #752

Return the branded type as-is (not wrapped with a string template) when used as a record key.

### DIFF
--- a/packages/remeda/src/internal/types/ToString.ts
+++ b/packages/remeda/src/internal/types/ToString.ts
@@ -1,0 +1,15 @@
+/**
+ * A utility to preserve strings as-is, convert numbers to strings, and fail on
+ * anything else. This happens a lot in JS when accessing objects or when
+ * enumerating over keys.
+ *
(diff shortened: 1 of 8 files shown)


packages/remeda/src/internal/types/ToString.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * A utility to preserve strings as-is, convert numbers to strings, and fail on
+ * anything else. This happens a lot in JS when accessing objects or when
+ * enumerating over keys.
+ *
+ * Notice that symbols are not supported, which is consistent with how built-in
+ * functions like `Object.keys` and `Object.entries` behave.
+ */
+export type ToString<T> = T extends unknown
```
cjquines, line 9: **suggestion:** we should maybe use this for `invert`, `keys`, `pickBy`,

cjquines (approved):

Issue #823, opened 2024-08-11
Make `constant` return specific type


gustavopch:
# Current behavior
```ts
const value = constant('something')
// <Args extends ReadonlyArray<unknown>>(...args: Args) => string
```

# Expected behavior
```ts
const value = constant('something')
// <Args extends ReadonlyArray<unknown>>(...args: Args) => 'something'
```

# Relevant source code
https://github.com/remeda/remeda/blob/4d2100c2c2e25053648bb0b6d8e645577bdd0899/src/constant.ts#L28-L32

# Proposed solution
```diff
-export function constant<T>(
+export function constant<const T>(
   value: T,
 ): <Args extends ReadonlyArray<unknown>>(...args: Args) => T {
   return () => value;
 }
```

cjquines:
idk if there's a reason we don't have any const generics in the codebase, @eranhirsch ?

it means dropping support for typescript <5.0, but the minimum ts version is now 5.1 anyway

eranhirsch:
`constant` was added in v1 when the minimum ts version was much lower. Now that it can be fully supported, it makes sense to add the `const` modifier to the `constant` utility function.

More widely, the `const` modifier is not always the right choice because, for every case where it improves ergonomics, there are cases where it will make it harder. Adding `as const` when you need to treat an input more specifically is easier than adding `as <my less specific type>` when you need to negate it; this means that for the data field (usually arrays and objects), it's better without it. Numbers are already treated as `as const`, so it doesn't add anything for those cases. The remaining cases where it does bring value are for strings and booleans, but there aren't many like these.

gustavopch:
BTW, genuine question: why does `constant` exist? I mean, isn’t it just extra characters and an extra import compared to what it does internally?
```
() => 'something'
constant('something')
```
Is it for completeness/symmetry or is it useful in other aspects that I’m not considering? Maybe nicer to use in `pipe`?

eranhirsch:
I wanted to deprecate `noop` in v2 because it could be used to do two different things which I felt is a bad API. Semantically `noop` is actually `doNothing`, but it only works when you need a function that returns `void` (which makes sense because some lint rules don't like empty functions, and `() => {}` can confuse a reader. Instead of adding another function that would return `undefined` (which is what `noop` **actually** did) I generalized it so that it would return any constant. When migrating from `noop` you'd use `constant(undefined)`, but now you can also do other stuff.

[...]

PR #1160, opened 2025-07-29
fix(const): prefer narrow typing for literals

eranhirsch:
Fixes: #823

This change only affects the output when a literal is used, it would return a narrow type based on the exact shape of the literal instead of widening `string`, `object`, and arrays.


cjquines (approved):
nice

PR #1163, opened 2025-08-01
fix: prevent redundant type computation paths

eranhirsch:
`If` is a mistake. When I saw it in type-fest it felt like its a safer, cleaner, way to do branching logic in complex types; but it has a huge caveat that is easy to err against: it results in both branches being computed. On the contrary, TypeScript only computes the relevant branch in an `extends` statement. This means that if one of the sides of the If statement are heavy to compute, it will always be computed.

Further more, Ifs don't actually narrow the type in their branches, so each branch needs to handle wide types - this means that types need to be over protective or run the risk of infinite loops and exessive recursions.


cjquines (approved):
nice. i was never a fan of `If` so it's nice to be vindicated 😅

PR #1178, opened 2025-08-11
docs: fix click surface for migration dropdown links

karl-run:
The click area of the links in the migrations dropdown doesn't match it's visual space as well as its hover-state. Making it easy to miss the link, but the dropdown still closes, so it feels weird.

Example:
image

This makes the dropdown element be the anchor itself, so it takes all the space, and matches the visual size.


eranhirsch (approved):
Thanks!

Issue #775, opened 2024-07-09
isEmpty errors on null


kwengontra:
```
> R.isEmpty(null)
Uncaught TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at Module.Wo (chunk-4DPIFY6Y.js:1:127)
    at <anonymous>:1:3
```

eranhirsch:
isEmpty does not accept `null` as an input.

kwengontra:
Could isEmpty be updated to accept null as an input? If not, what is the motivation?

eranhirsch:
`null` is not a container and has no concept of being empty. Returning `false` for `isEmpty(null)` would be expected, but it's not the only semantically correct response. It's easy to do `isEmpty(x ?? [])` or even `x === null || isEmpty(x)` if that's the logic you need, but someone else might want to handle `null`s differently.

In Remeda, we prefer the input types be as strict as possible so that you get compile-time errors if a change upstream causes a mismatching type.

J3m5:
> isEmpty does not accept `null` as an input.

This should be added to the docs IMO

cjquines:
i think it's fair to add it to the docs, considering that `undefined` gets a mention. i'll reopen and rename the ticket

scamden:
It's worth mentioning that on the other hand `R.isEmpty(undefined)` returns true.

programming-with-ia:
Hey! 👋

I needed `isEmpty` to be fully type-safe and handle `null`, so I wrapped `remeda's` version like this (with a little help from GPT):

```

import { isEmpty as REmpty } from "remeda";

type IterableContainer<T = unknown> = ReadonlyArray<T> | readonly [];

// Overload: string | undefined | null
export function isEmpty<T extends string | undefined | null>(
    value: T,
): value is
    | ("" extends T ? "" : never)
    | (undefined extends T ? undefined : never)
    | (null extends T ? null : never);

// Overload: iterable containers (like array)
export function isEmpty(value: IterableContainer): value is [];

// Overload: plain object
export function isEmpty<T extends object>(
    value: T,
): value is Record<keyof T, never>;

// Implementation
export function isEmpty(value: unknown): boolean {
    if (value === null) return true;
    return REmpty(value as any);
}

```

risalfajar:
> In Remeda, we prefer the input types be as strict as possible so that you get compile-time errors if a change upstream causes a mismatching type.

[...]

PR #1180, opened 2025-08-11
feat(isEmptyish): isEmpty++

eranhirsch:
Closes: #775

### DIFF
--- a/packages/remeda/src/isEmptyish.test.ts
+++ b/packages/remeda/src/isEmptyish.test.ts
@@ -1,0 +1,134 @@
+import { expect, test, describe } from "vitest";
+import { isEmptyish } from "./isEmptyish";
+
+test("nullish", () => {
+  expect(isEmptyish(null)).toBe(true);
+  expect(isEmptyish(undefined)).toBe(true);
+});
+
+test("strings", () => {
+  expect(isEmptyish("")).toBe(true);
+  expect(isEmptyish("test")).toBe(false);
+});
+
+test("arrays", () => {
+  expect(isEmptyish([])).toBe(true);
+  // eslint-disable-next-line unicorn/new-for-builtins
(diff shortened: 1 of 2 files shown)


packages/remeda/src/isEmpty.ts
```diff
@@ -3,47 +3,59 @@ import type { IterableContainer } from "./internal/types/IterableContainer";
...
- *    R.isEmpty([]) //=> true
- *    R.isEmpty({}) //=> true
- *    R.isEmpty('test') //=> false
- *    R.isEmpty([1, 2, 3]) //=> false
- *    R.isEmpty({ length: 0 }) //=> false
+ *    R.isEmpty(''); //=> true
+ *    R.isEmpty([]); //=> true
+ *    R.isEmpty({}); //=> true
+ *
+ *    R.isEmpty(undefined); //=> false
```
cjquines, line 25: **question:** should we include this given that it's not the actual runtime behavior right now?

cjquines (commented):
will review closer later, but my main question is: is there a point to keeping both `isEmpty` and `isEmptyish`? can we not just deprecate `isEmpty` in v3?

eranhirsch:
I didn't respond to your concerns about having both functions. I still strongly believe that there's value in a utility that can also surface that the input itself might be broken (e.g., something that was once `string` became `string | undefined`) or that you are checking something that is semantically wrong (e.g., checking some random class that doesn't have a notion of empty). Because it's unlikely that you'd use the type after an isEmpty, even narrowing to `never` might go unnoticed for a while.

isEmptyish accepts anything, and can effectively "turn off" narrowing when it doesn't know how to narrow; that means that isEmptyish would cause compile-time errors very rarely, especially on bad inputs.

I do think we should backport the more advanced typing I added in isEmptyish to isEmpty.

cjquines (approved):

Issue #729, opened 2024-06-12
Explain how the 3rd (data) param works for lazy invocations


mass8326:
## Description

The parameters passed into `map()` do not match between data-first/data-last vs pipe/piped.

It seems to happen in the latest version v2.0.9 with `map()` and seems to go as far back as v1.0.0 with `map.indexed()`

Is this expected behavior? I didn't seem to catch this in the documentation for either `map()` or `pipe()`.

REPL: https://stackblitz.com/edit/remeda-piped-map-issue?file=index.js

## Input
```js
import * as remeda from 'remeda';

console.log('Data first');
remeda.map([1, 2, 3, 4], (val, i, arr) => console.log(arr));

console.log('Data last');
remeda.map((val, i, arr) => console.log(arr))([1, 2, 3, 4]);

console.log('Pipe');
remeda.pipe(
  [1, 2, 3, 4],
  remeda.map((val, i, arr) => console.log(arr))
);

console.log('Piped');
remeda.piped(remeda.map((val, i, arr) => console.log(arr)))([1, 2, 3, 4]);
```

## Output
```
Data first
[ 1, 2, 3, 4 ]
[ 1, 2, 3, 4 ]
[ 1, 2, 3, 4 ]
[ 1, 2, 3, 4 ]
Data last
[ 1, 2, 3, 4 ]
[ 1, 2, 3, 4 ]
[ 1, 2, 3, 4 ]
[ 1, 2, 3, 4 ]
Pipe
[ 1 ]
[ 1, 2 ]
[ 1, 2, 3 ]
[ 1, 2, 3, 4 ]
Piped
[ 1 ]
[ 1, 2 ]
[ 1, 2, 3 ]
[ 1, 2, 3, 4 ]
```

cjquines:
i'm not sure where we can document this better—we have this pattern for nearly all of our lazy indexed functions.

eranhirsch:
You can disable lazy invocation by wrapping a data-first call: 
```ts
remeda.pipe(
  [1, 2, 3, 4],
  ($) => remeda.map($, (val, i, arr) => console.log(arr))
);
```

@cjquines, we could add an article to the top of the docs explaining this behavior. I'll remove the wontfix tag so we can track this, but I admit it's gonna be pretty low-pri as I wouldn't add an article about this before we also add other articles first.

PR #1182, opened 2025-08-15
docs(pipe): improve docs

eranhirsch:
Pipe is really important but our docs were very lacking.

Closes: #729

### DIFF
--- a/packages/remeda/src/pipe.ts
+++ b/packages/remeda/src/pipe.ts
@@ -19,15 +19,83 @@ type LazyOp = LazyDefinition & ((input: unknown) => unknown);
 /**
  * Perform left-to-right function composition.
  *
- * @param value - The initial value.
- * @param operations - The list of operations to apply.
+ * By taking advantage of Remeda's built-in currying, `pipe` enables you to
+ * easily convert deeply nested, onion-like, transformations into a readable
+ * top-to-bottom data flow that matches how you think about the transformation.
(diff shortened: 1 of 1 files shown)


cjquines (commented):
hm... i think we might want to rethink where this documentation should live. having such a long docstring is bad for editors because it just makes the jsdoc too long

eranhirsch:
I understand your concerns, and my initial idea was also to use our articles feature in the docs site to create a separate entry for it in the docs, but I decided against it when I realized that most users would never read it there.

Looking at the original docs for `pipe`, I can't imagine how users understand the full extent of what the function is capable of; its documentation **had** to be expanded. Once you expand the docs for `pipe`, it's already *the* place to talk about lazy evaluation and data-last conversions; putting it in a different place *too* would just be redundant. And if you assume that most users interact with the docs site via search and take the fact that `pipe` is the most searched for term, it's clear that more users would see this information here than anywhere else.

Regarding making the JSDoc big (or the tooltip unwieldy), what is the main concern here and the negative impact you want to prevent? I don't think users even interact much with those tooltips, so when they do, they are already in "reading mode" so the length doesn't change their experience. `pipe` doesn't take complex params so people aren't using the JSDoc as a referesher for what each param is, they are either reading it as a whole to learn it, or ignoring it completely.

cjquines (approved):

Issue #1175, opened 2025-08-06
exporting a function that returns a generic type based on the return type of `remeda.values` gives `error TS2742` when there is  `"composite":true,  "module": "node16"` in `tsconfig.json`,


hab25:
Running `tsc` on the following files:
```json
// tsconfig.json
{
    "compilerOptions": {
        "composite": true,
        "module": "node16",
        "lib": ["es2020"]
    },
    "files": ["index.mts"]
}
```

```typescript
// index.mts
import * as R from "remeda";

export const getValuesWithRemeda = <T extends Record<string, unknown>, >(record: T) =>
    R.values(record);

// For comparison, try uncommenting below; it compiles successfully!
/*
export const getValues = <T extends Record<string, unknown>, >(record: T) =>
    Object.values(record);
*/
```

gives output like
```
index.mts:3:14 - error TS2742: The inferred type of 'getValuesWithRemeda' cannot be named without a reference to './node_modules/remeda/dist/EnumerableStringKeyedValueOf-BU9R_cEk.js'. This is likely not portable. A type annotation is necessary.

3 export const getValuesWithRemeda = <T extends Record<string, unknown>, >(record: T) =>
               ~~~~~~~~~~~~~~~~~~~

index.mts:3:14 - error TS2742: The inferred type of 'getValuesWithRemeda' cannot be named without a reference to './node_modules/remeda/dist/IterableContainer-CtfinwiH.js'. This is likely not portable. A type annotation is necessary.

3 export const getValuesWithRemeda = <T extends Record<string, unknown>, >(record: T) =>
               ~~~~~~~~~~~~~~~~~~~


Found 2 errors in the same file, starting at: index.mts:3
```

Minimal reproduction example: https://github.com/hab25/remeda-typescript-error-minimal-reproduction-example

hab25:
Update: the error happens if one uses just `"declaration": true` instead of `"composite": true` (which implies `"declaration": true`), so the former is a more minimal reproduction.

hab25:
> Update: the error happens if one uses just `"declaration": true` instead of `"composite": true` (which implies `"declaration": true`), so the former is a more minimal reproduction.

I've updated the MRE with this.

I've updated the MRE with a comment explaining that a workaround is to add 

```
// index.mts
import type { E } from './node_modules/remeda/dist/EnumerableStringKeyedValueOf-BU9R_cEk.js';
[...]

PR #1241, opened 2025-12-31
feat(build): generate a monolithic type definition file

eranhirsch:
Fixes: #1175

Our current build genreates tiny files that are good for tree-shaking but not good for reexporting our types when used within a multi-output project.

To fix this I needed to restructure our build process to allow these 2 different aspects to play along nicely.

NOTE: I'm using a `feat` PR type instead of a `fix` although the change is largely under-the-hood and shouldn't cause any breakages. This is just in case, so that user get a proper "heads up" that our build artifacts changed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22475,9 +22475,6 @@
     "packages/remeda": {
       "version": "2.0.0",
       "license": "MIT",
-      "dependencies": {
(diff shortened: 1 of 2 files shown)


cjquines:
oh, i think this is a good idea

PR #1313, opened 2026-04-05
fix(isDeepEqual): allow null-prototype objects

### LINKED ISSUE #1228: `isDeepEqual` is not the same as v1's `equals`
I just upgraded v1 to v2, changed `equals` -> `isDeepEqual` (the docs suggest it's just a rename) and our unittests started failing. 
I did some digging and found a conspicuous line that doesn't exist in `equals`: https://github.com/remeda/remeda/blob/b865a350d4f9278bfc327001a04ab677917cfa68/src/isDeepEqual.ts#L86
The `getPrototypeOf` comparison breaks for us (I removed it to test). 
I understand there are trade-offs and you guys decided on performance above backward-compatibility, but the docs should at least mention this. Especially because Node's built-in `querystring.parse` doesn't play nicely with this decision; See https://nodejs.org/api/querystring.html#querystringparsestr-sep-eq-options: `The object returned by the querystring.parse() method does not prototypically inherit from the JavaScript Object.`

cyphercodes:
## Summary

Allow `isDeepEqual` to treat plain objects and null-prototype objects as comparable when their enumerable properties match.

## Changes

- Removed the optional `strict` parameter to avoid `purry` optional-parameter ambiguity.
- Kept the existing prototype check for non-null/custom prototype mismatches.
- Added runtime coverage for null-prototype objects, recursive comparison, data-last usage, and custom prototype mismatches.

Fixes #1228


eranhirsch (changes requested):
The way `purry` detects data-first and data-last calls in runtime prevents us from being able to use optional function parameters easily, and functions that use them require us to write our own purrying logic (see `toCamelCase` for example).

But I think that the right fix here would be to skip adding an additional options parameter at all, and instead simply fixing the prototype check to explicitly objects that have a `null` prototype (like those created via `Object.create(null)`) to pass the prototype check against objects that have the raw `Object.prototype` as their prototype. e.g.,

```ts
const dataPrototype = Object.getPrototypeOf(data);
const otherPrototype = Object.getPrototypeOf(other);

if (
  dataPrototype !== otherPrototype &&
  (dataPrototype !== null || otherPrototype !== Object.prototype) &&
  (dataPrototype !== Object.prototype || otherPrototype !== null)
) {
  // ...
}
```

This would effectively fix the function for the `null` prototype case, but it wouldn't work in the generic case like `querystring` where I already mentioned in my response in the attached issue I don't think is a bug and shouldn't be accepted.

cyphercodes:
Updated based on the review feedback.

Changes made:
- Removed the optional `strict` parameter so `isDeepEqual` stays compatible with `purry`'s runtime arity detection.
- Changed the prototype check to allow `null` prototype objects to compare equal with plain `Object.prototype` objects when their enumerable properties match.
- Kept custom/non-null prototype mismatches unequal.
- Added runtime coverage for null-prototype objects, recursive comparison, data-last usage, and custom prototype mismatches.
- Updated the PR title/body to match the new scope.

Verified locally:
- `npx prettier --write packages/remeda/src/isDeepEqual.ts packages/remeda/src/isDeepEqual.test.ts`
- `git diff --check`
- `npx eslint --max-warnings 0 src/isDeepEqual.ts src/isDeepEqual.test.ts`
- `npm run test:runtime --workspace remeda -- isDeepEqual.test.ts --run`
- `npm run check --workspace remeda`

packages/remeda/src/isDeepEqual.ts
```diff
@@ -159,6 +160,22 @@ function isDeepEqualImplementation<T>(data: unknown, other: T): data is T {
   return true;
 }
 
+function hasDifferentComparablePrototypes(
```
eranhirsch: We're pulling a function out so let's not use negated semantics; instead of `hasDifferentComparablePrototypes` `isComparablePrototype`. It will also make the implementation more readable by allowing a set of early conditional returns instead of a single one.

(I know this doesn't read out as clearly, but `isXXX` is a good convention for boolean predicates over `areXXX`)

packages/remeda/src/isDeepEqual.ts
```diff
@@ -159,6 +160,22 @@ function isDeepEqualImplementation<T>(data: unknown, other: T): data is T {
...
+  data: object,
+  other: object,
+): boolean {
+  const dataPrototype: unknown = Object.getPrototypeOf(data);
+  const otherPrototype: unknown = Object.getPrototypeOf(other);
```
eranhirsch: Avoid casting or hard-coding types, we prefer explicit lint/typescript suppression comments instead.

packages/remeda/src/isDeepEqual.ts
```diff
@@ -159,6 +160,22 @@ function isDeepEqualImplementation<T>(data: unknown, other: T): data is T {
...
+
+  return (
+    dataPrototype !== otherPrototype &&
+    !(
+      (dataPrototype === null && otherPrototype === Object.prototype) ||
+      (dataPrototype === Object.prototype && otherPrototype === null)
+    )
+  );
```
eranhirsch: once we flip the semantics of the function this becomes:


```suggestion
  if (dataPrototype === otherPrototype) {
    return true;
  }
  
  if (dataPrototype === null) {
    return otherPrototype === Object.prototype;
  }
  
  return dataPrototype === Object.prototype && otherPrototype === null;
```

packages/remeda/src/isDeepEqual.test.ts
```diff
@@ -410,3 +411,38 @@ function func1(): void {
...
+    class TestObject {
+      public readonly a = 1;
+    }
+
+    expect(isDeepEqual(new TestObject() as unknown, { a: 1 })).toBe(false);
+  });
+});
+
+function createNullPrototypeObject<T extends object>(data: T): T {
+  return Object.assign(Object.create(null) as object, data);
```
eranhirsch: why do we need the `as object` here? Prefer suppressing typescript errors explicitly with @ts-expect-error instead

packages/remeda/src/isDeepEqual.test.ts
```diff
@@ -410,3 +411,38 @@ function func1(): void {
 function func2(): void {
   throw new Error("This function should never be called!");
 }
+
+describe("null-prototype objects", () => {
+  test("are equal to equivalent plain objects", () => {
+    const nullProtoObject = createNullPrototypeObject({ a: 1, b: "2" });
+
+    expect(isDeepEqual(nullProtoObject, { a: 1, b: "2" })).toBe(true);
+    expect(isDeepEqual({ a: 1, b: "2" }, nullProtoObject)).toBe(true);
```
eranhirsch: pull out the common "data" value


```suggestion
    const data = { a: 1, b: "2" };
    const nullProtoObject = createNullPrototypeObject(data);

    expect(isDeepEqual(nullProtoObject, data)).toBe(true);
    expect(isDeepEqual(data, nullProtoObject)).toBe(true);
```

packages/remeda/src/isDeepEqual.test.ts
```diff
@@ -410,3 +411,38 @@ function func1(): void {
...
+      nested: createNullPrototypeObject({ a: 1 }),
+    });
+
+    expect(isDeepEqual(nullProtoObject, { nested: { a: 1 } })).toBe(true);
+  });
+
+  test("work in data-last form", () => {
+    const nullProtoObject = createNullPrototypeObject({ a: 1 });
+
+    expect(pipe({ a: 1 }, isDeepEqual(nullProtoObject))).toBe(true);
```
eranhirsch: Similarly


```suggestion
    const data = { a: 1 };
    const nullProtoObject = createNullPrototypeObject(data);

    expect(pipe(data, isDeepEqual(nullProtoObject))).toBe(true);
```

packages/remeda/src/isDeepEqual.test.ts
```diff
@@ -410,3 +411,38 @@ function func1(): void {
...
+  });
+
+  test("objects with different non-null prototypes are not equal", () => {
```
eranhirsch, line 461: Great test!

packages/remeda/src/isDeepEqual.test.ts
```diff
@@ -1,5 +1,6 @@
 import { describe, expect, test } from "vitest";
```
eranhirsch, line 1: I'm missing a few tests:
1. A test that compares 2 equal objects with null prototypes.
2. Negative tests to ensure we don't short-circuit the deep check, e.g. comparing `{ a: 1 }` with `createNullPrototypeObject({ a: 2 })`
3. A negative test for 2 null-prototyped objects (a combination of both of my previous concerns).

packages/remeda/src/isDeepEqual.test.ts
```diff
@@ -410,3 +411,38 @@ function func1(): void {
...
+  test("objects with different non-null prototypes are not equal", () => {
+    class TestObject {
+      public readonly a = 1;
+    }
+
+    expect(isDeepEqual(new TestObject() as unknown, { a: 1 })).toBe(false);
+  });
+});
+
+function createNullPrototypeObject<T extends object>(data: T): T {
```
eranhirsch: Just for readability, perhaps a shorter name would make more sense, e.g. `nullifyPrototype` or `withNullPrototype`

eranhirsch (changes requested):
Thanks for addressing my comments!

cyphercodes:
Updated this PR to address the latest review feedback.

Changes made:
- Kept the purry signature unchanged; the previous optional-parameter approach is no longer present/needed for this scoped fix.
- Renamed the prototype helper to positive `isComparablePrototype` semantics and rewrote it with early returns.
- Replaced the hard-coded `unknown` prototype annotations with explicit lint suppressions matching existing low-level prototype checks.
- Reworked the null-prototype tests to share common data, renamed the helper to `withNullPrototype`, removed unnecessary casts, and added the missing equal/negative null-prototype coverage.

Verification:
- `npx prettier --write packages/remeda/src/isDeepEqual.ts packages/remeda/src/isDeepEqual.test.ts` ✅
- `npx eslint --max-warnings 0 src/isDeepEqual.ts src/isDeepEqual.test.ts` ✅
- `npm run test:runtime --workspace remeda -- isDeepEqual.test.ts --run` ✅
- `npm run test:types --workspace remeda -- isDeepEqual.test-d.ts --run` ✅
- `npm run check --workspace remeda` ✅
- `git diff --check` ✅

packages/remeda/src/isDeepEqual.ts
```diff
@@ -130,7 +130,8 @@ function isDeepEqualImplementation<T>(data: unknown, other: T): data is T {
     return data.toString() === (other as unknown as RegExp).toString();
   }
 
-  // At this point we only know that the 2 objects share a prototype and are not
+  // At this point we only know that the 2 objects share a prototype (or one of
+  // them has a null prototype while the other uses Object.prototype) and are not
```
eranhirsch: This is too verbose, it repeats what the code already shows, it's enough to mention that the 2 objects have a comparable prototype.


```suggestion
  // At this point we only know that the 2 objects have a comparable prototype and are not
```

packages/remeda/src/isDeepEqual.test.ts
```diff
@@ -410,3 +411,63 @@ function func1(): void {
...
+    class TestObject {
+      public readonly a = 1;
+    }
+
+    expect(isDeepEqual(new TestObject(), { a: 1 })).toBe(false);
+  });
+});
+
+function withNullPrototype<T extends object>(data: T): T {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return -- Object.create(null) intentionally creates a null-prototype object and is typed as `any`.
```
eranhirsch: This is too verbose, we just need to explain why this is OK, not repeat what the code already shows us.


```suggestion
  // eslint-disable-next-line @typescript-eslint/no-unsafe-return -- we rely on the type of Object.create which isn't typed.
```

packages/remeda/src/isDeepEqual.test.ts
```diff
@@ -410,3 +411,63 @@ function func1(): void {
...
+
+    expect(isDeepEqual(data, nullProtoObject)).toBe(false);
+    expect(isDeepEqual(nullProtoObject, data)).toBe(false);
+  });
+
+  test("still compare properties against null-prototype objects", () => {
+    expect(
+      isDeepEqual(withNullPrototype({ a: 1 }), withNullPrototype({ a: 2 })),
+    ).toBe(false);
+  });
```
eranhirsch: I feel that the "still..." phrasing is off here, it assumes tests are being read in order, but if someone just sees a single test they won't have the same context.

I don't have a strong suggestion for a rephrasing here, but i'm sure an LLM would be able to come up with something.

eranhirsch (dismissed):
This looks great, thanks for working on it and seeing through my comments too. I've added some nits about phrasing, if you have time to go over them it'd really help.

packages/remeda/src/isDeepEqual.ts
eranhirsch: We also need to add a line in the docs about the null prototype comparison, as it's not trivial and users would need to know about it.

cyphercodes:
Updated this PR for the latest review nits:
- Added the null-prototype comparison behavior to the `isDeepEqual` docs.
- Shortened the explanatory comments/test names around comparable prototypes.

Verification:
- `git diff --check` ✅
- `cd packages/remeda && npx eslint --max-warnings 0 src/isDeepEqual.ts src/isDeepEqual.test.ts` ✅
- `cd packages/remeda && npm run test:runtime -- isDeepEqual.test.ts --run` ✅
- `cd packages/remeda && npm run test:types -- isDeepEqual.test-d.ts --run` ✅
- `cd packages/remeda && npm run check` ✅

eranhirsch (approved):
amazing! thanks!

PR #1322, opened 2026-04-26
chore(cc): Nits from code review

eranhirsch:
When shipping #1320 I had a few review comments I didn't have time to go over.

This PR just tightens the implementation a bit based on those comments


eranhirsch:
### Code review

Found 1 issue:

1. `hasFilesNewerThan` constructs the wrong path for nested files. With `readdirSync(dir, { withFileTypes: true, recursive: true })`, `Dirent.name` is the bare filename, not a relative path; nested entries (e.g. `src/internal/foo.ts`) carry the directory in `Dirent.parentPath`. So `path.join(directory, name)` resolves to `src/foo.ts` and `statSync` throws ENOENT for any file under `src/internal/`. The check will crash on the first edit to an internal helper rather than detecting staleness. Use `path.join(entry.parentPath, entry.name)` (or pass `entry` and join `parentPath`+`name`).

https://github.com/remeda/remeda/blob/9d5f3114de212fa4b1a12546ec644f133dce3822/.claude/hooks/docs-local-ci-checks.ts#L77-L80

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1344, opened 2026-05-07
feat(hasProp): add `hasProp`

lo1tuma:
Adds two object key-existence type-predicates:

  - `hasKey(data, key)` — mirrors `in`. Includes inherited properties.
  - `hasOwnKey(data, key)` — mirrors `Object.hasOwn`. Own properties only.

  Both narrow `data` to `T & Record<Key, unknown>`. The narrowed value type is
  `unknown` because key presence makes no claim about the value — use `prop`
  to read it.

  Refs #461. Re-attempts #515, which was closed when its base branch was
  deleted (and had unresolved typing concerns).

  ## Addressing @eranhirsch's #515 review

  - **Mirror runtime exactly.** `hasKey({ a: undefined }, "a")` is `true`. The
    value type stays `unknown` rather than narrowing `undefined` away —
    *"`undefined` is not special when it comes to prop values."*
  - **No special handling for unions or arrays.** Input is plain `T extends
    object`; TypeScript narrows naturally. Arrays satisfy `object` at runtime
    (covered in tests) but the type doesn't shape around them — *"We already
    offer `hasAtLeast` to narrow arrays."*
  - **No custom type helpers; no `Simplify`.** Narrowing is just `T &
    Record<Key, unknown>` — addresses both the type-fest preference and the
    TS-perf concern.
  - **Predicate assignability.** `T & Record<Key, unknown>` is always
    assignable to `T`, so the `Omit`-based path that broke #515 doesn't come
    up here.
  - **TS#31025 acknowledged, not fought.** The `.test-d.ts` files assert the
    unevaluated intersection rather than a hand-collapsed shape.
  - **TS 5.5+ predicate inference.** The internal helpers don't re-declare
    `data is …` — TS infers it from `key in data` / `Object.hasOwn(data, key)`.

  ## Naming

  `hasKey` / `hasOwnKey` over `hasProp` because these check **key
  existence**, not property values, and directly mirror `in` / `Object.hasOwn`.
  "Key" also matches the rest of Remeda's surface (`pick`, `omit`, `fromKeys`,
  `keys`, `mapKeys`). Happy to rename if you prefer `hasProp`/`hasOwnProp` —
  flagging it explicitly.


packages/remeda/src/hasOwnKey.ts
```diff
@@ -0,0 +1,63 @@
+import { purry } from "./purry";
+
+/**
+ * Checks if `data` has `key` as an own property, mirroring `Object.hasOwn`.
```
eranhirsch: let's link this to MDN so people who don't know what this means can learn more
```suggestion
 * Checks if `data` has `key` as an own property, mirroring [`Object.hasOwn`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn).
```

packages/remeda/src/hasOwnKey.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Inherited properties on the prototype chain are ignored, but own properties
+ * whose value is `undefined` still return `true`.
```
eranhirsch: We don't need to state this, it's given when `exactOptionalPropertyTypes` is enabled.


```suggestion
 * Inherited properties on the prototype chain are ignored.
```

packages/remeda/src/hasOwnKey.ts
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Acts as a type-predicate, narrowing `data` to require `key` to be present.
+ * The narrowed value type is `unknown` because runtime presence makes no claim
+ * about the value — use `prop` to read it. To follow the prototype chain
```
eranhirsch: This is redundant, we don't need to explain our return type logic, that fits as a comment on the type itself, the user would expect our return type to make sense.


```suggestion
 * — use `prop` to read it. To follow the prototype chain
```

packages/remeda/src/hasOwnKey.ts
```diff
@@ -0,0 +1,63 @@
...
+ * instead (i.e. mirror the `in` operator), use `hasKey`.
+ *
```
eranhirsch: Refer to other utilities as a list so it's easier to scan, less prose


```suggestion
 * — use `prop` to read a prop. 
 * - use `hasKey` for a prototype-following check instead (which uses the `in` operator). 
```

packages/remeda/src/hasOwnKey.ts
```diff
@@ -0,0 +1,63 @@
...
+ *   R.hasOwnKey(data, key)
+ * @example
+ *   R.hasOwnKey({ a: 1 }, "a"); //=> true
+ *   R.hasOwnKey({ a: 1 }, "b"); //=> false
+ *   R.hasOwnKey({ a: undefined }, "a"); //=> true
+ *   R.hasOwnKey({}, "toString"); //=> false (inherited, not own)
+ * @dataFirst
+ * @category Object
+ */
+export function hasOwnKey<T extends object, Key extends PropertyKey>(
```
eranhirsch: Key is not bound to T, meaning that any key is acceptable, even keys which we know at compile time which would never return `true`. This prevents us from surfacing possible typos and changes to the underlying types.

e.g.,

```ts
const data = { a: string } as const;

// This is meaningless check which could be surfaced at compile time...
hasOwnKey(data, "b");
```

packages/remeda/src/hasOwnKey.test.ts
```diff
@@ -0,0 +1,94 @@
+import { describe, expect, test } from "vitest";
+import { hasOwnKey } from "./hasOwnKey";
+import { pipe } from "./pipe";
+
+describe("data-first", () => {
```
eranhirsch: A lot of utilities have this pattern of data-first as a full describe block, but i'm migrating our tests to new format where the data-first tests are not grouped automatically, they don't need to, it's obvious from context.


```suggestion
```

packages/remeda/src/hasOwnKey.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  test("returns false for inherited prototype properties", () => {
+    expect(hasOwnKey({}, "toString")).toBe(false);
+  });
+
+  test("returns false for properties only on a custom prototype", () => {
+    const proto = { inherited: 1 };
+    const data = Object.create(proto) as object;
+
+    expect(hasOwnKey(data, "inherited")).toBe(false);
```
eranhirsch: inline everything, and no need to use assertions, non-type tests shouldn't have any type constraint, if you encounter something prefer `// @ts-expect-error` comments instead.


```suggestion
    expect(hasOwnKey(Object.create({ inherited: 1 }), "inherited")).toBe(false);
```

packages/remeda/src/hasOwnKey.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  });
+
+  test("returns true for own properties on a null-prototype object", () => {
+    const data = Object.assign(Object.create(null) as object, { a: 1 });
+
+    expect(hasOwnKey(data, "a")).toBe(true);
```
eranhirsch: similarly


```suggestion
    expect(hasOwnKey(Object.assign(Object.create(null), { a: 1 }), "a")).toBe(true);
```

packages/remeda/src/hasOwnKey.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  });
+
+  test("returns false for missing properties on a null-prototype object", () => {
+    const data = Object.create(null) as object;
+
+    expect(hasOwnKey(data, "toString")).toBe(false);
+  });
```
eranhirsch: This test seems redundant and repeats the premise of "returns false for inherited prototype properties"

```suggestion
```

packages/remeda/src/hasOwnKey.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  test("returns true for own symbol keys", () => {
+    const symbol = Symbol("test");
+
+    expect(hasOwnKey({ [symbol]: 1 }, symbol)).toBe(true);
+  });
+
+  test("returns true for numeric keys", () => {
+    expect(hasOwnKey({ 0: "a" }, 0)).toBe(true);
+  });
```
eranhirsch: good test!

packages/remeda/src/hasOwnKey.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    expect(hasOwnKey([1, 2, 3], 5)).toBe(false);
+  });
+
+  test("returns false for inherited array members", () => {
+    expect(hasOwnKey([1, 2, 3], "push")).toBe(false);
+  });
+
+  test("returns true for the array length own property", () => {
+    expect(hasOwnKey([1, 2, 3], "length")).toBe(true);
+  });
```
eranhirsch: let's group all array tests in a single describe


```suggestion
  describe("arrays", () => {
    test("returns true for an array index that exists", () => {
      expect(hasOwnKey([1, 2, 3], 1)).toBe(true);
    });

    test("returns false for an array index past the end", () => {
      expect(hasOwnKey([1, 2, 3], 5)).toBe(false);
    });

    test("returns false for inherited array members", () => {
      expect(hasOwnKey([1, 2, 3], "push")).toBe(false);
    });

    test("returns true for the array length own property", () => {
      expect(hasOwnKey([1, 2, 3], "length")).toBe(true);
    });
  });
```

packages/remeda/src/hasOwnKey.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  test("returns true for own symbol keys", () => {
+    const symbol = Symbol("test");
+
+    expect(pipe({ [symbol]: 1 }, hasOwnKey(symbol))).toBe(true);
+  });
+
+  test("returns true for numeric keys", () => {
+    expect(pipe({ 0: "a" }, hasOwnKey(0))).toBe(true);
+  });
+});
```
eranhirsch: great!

packages/remeda/src/hasOwnKey.test-d.ts
```diff
@@ -0,0 +1,148 @@
+import { describe, expectTypeOf, test } from "vitest";
+import { hasOwnKey } from "./hasOwnKey";
+import { pipe } from "./pipe";
+
+describe("data-first", () => {
```
eranhirsch: No need for a top-level describe for data-first, it's assumed...


```suggestion
```

packages/remeda/src/hasOwnKey.test-d.ts
```diff
@@ -0,0 +1,148 @@
...
+  test("returns boolean", () => {
+    expectTypeOf(hasOwnKey({ a: 1 }, "a")).toEqualTypeOf<boolean>();
+  });
```
eranhirsch: This is given, that's just what type-predicates do, there's no value in testing it

packages/remeda/src/hasOwnKey.test-d.ts
```diff
@@ -0,0 +1,148 @@
...
+    } else {
+      expectTypeOf(data).toEqualTypeOf<{ a?: string }>();
+    }
+  });
+
+  test("narrows missing keys with unknown", () => {
+    const data = {} as { a: number };
+
+    if (hasOwnKey(data, "b")) {
+      expectTypeOf(data).toEqualTypeOf<{ a: number } & Record<"b", unknown>>();
```
eranhirsch: This should be `never`! or better yet, 'b' is supposed to raise a type-error. Although technically TypeScript's duck typing means that an object provided to this function might have other props that are not listed in it's type, that's not how types are used in actual projects.


```suggestion
      expectTypeOf(data).toEqualTypeOf<never>();
```

packages/remeda/src/hasOwnKey.test-d.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+  });
+
+  test("rejects non-object data", () => {
+    // @ts-expect-error [ts2345] - primitives are not allowed
+    hasOwnKey("not an object", "a");
+
+    // @ts-expect-error [ts2345] - null is not assignable to object
+    hasOwnKey(null, "a");
+  });
```
eranhirsch: this test isn't useful, the typing prevents this already, we don't need to test that TypeScript enforces input types


```suggestion
```

packages/remeda/src/hasOwnKey.test-d.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  test("narrows union types naturally without special handling", () => {
+    const data = {} as { a: number } | { b: string };
+
+    if (hasOwnKey(data, "a")) {
+      expectTypeOf(data).toEqualTypeOf<
+        { a: number } | ({ b: string } & Record<"a", unknown>)
+      >();
+    }
+  });
```
eranhirsch: This would probably work the opposite of how people would expect it to work.

If my type is `{ a: number } | { b: string }` and I ask if `a` is in the object, what i'm actually asking is do i have an object of shape `{ a: number }` or a shape of shape `{ b: string }` so I can safely access the props. Widening the type would prevent this flow.


```suggestion
  test("narrows union types naturally without special handling", () => {
    const data = {} as { a: number } | { b: string };

    if (hasOwnKey(data, "a")) {
      expectTypeOf(data).toEqualTypeOf<{ a: number }>();
    } else {
      expectTypeOf(data).toEqualTypeOf<{ b: string }>();
  });
```

packages/remeda/src/hasOwnKey.test-d.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  test("acts as a type predicate for an optional property", () => {
+    const data = {} as { a?: string };
+
+    if (hasOwnKey(data, "a")) {
+      expectTypeOf(data).toEqualTypeOf<{ a?: string } & Record<"a", unknown>>();
+    } else {
+      expectTypeOf(data).toEqualTypeOf<{ a?: string }>();
+    }
+  });
```
eranhirsch: i'd expect the type to be `{ a: string }` here, are we not narrowing the result enough?


```suggestion
  test("acts as a type predicate for an optional property", () => {
    const data = {} as { a?: string };

    if (hasOwnKey(data, "a")) {
      expectTypeOf(data).toEqualTypeOf<{ a: string }>();
    } else {
      expectTypeOf(data).toEqualTypeOf<EmptyObject>();
    }
  });
```

packages/remeda/src/hasOwnKey.test-d.ts
```diff
@@ -0,0 +1,148 @@
...
+    hasOwnKey("not an object", "a");
+
+    // @ts-expect-error [ts2345] - null is not assignable to object
+    hasOwnKey(null, "a");
+  });
+
+  test("rejects non-PropertyKey keys", () => {
+    // @ts-expect-error [ts2345] - boolean is not a PropertyKey
+    hasOwnKey({ a: 1 }, true);
+  });
```
eranhirsch: same


```suggestion
```

eranhirsch (changes requested):
This is a great push and really useful utility!

But it lacks the kind of typing we have in other utilities and that would make it the most useful in real-life projects, this has two parts to it:
1. **Coupling** the input key to T (via type-fest's `KeysOfUnion` to allow unions too) so that only keys that are plausible for the object are accepted to begin with, raising a TypeScript error when providing a key that can't exist on the object (e.g. the "typo" scenario: `hasKey({ name?: string }, "naem")`, this is what we do in `pick`, `omit`, `prop` and more), and is what allows Remeda to offer a level of type-safety that looks upstream and not only downstream.
2. **Narrowing** the output by removing optionality from existing keys instead of _widening_ the output by tacking on missing props. This reverses the way that the `in` operator works in TypeScript (since TS 4.9) but is more likely to be what people need more.

Additionally, there is no need for two utilities. In modern TS/JS it is unlikely that people are working with complex prototype chains, and are more likely using `in` and `hasOwn` interchangeably. TypeScript doesn't have any notion of "prototype" props vs own props so the types for the two functions are identical so they don't add anything one on top of the other. `in` and `hasOwn` end up being implementation details of the general notion of "is this optional prop part of this object or not"; for that `hasOwn` is probably more accurate semantically exactly because of edge-cases like `toString` which can throw people off and will be typed incorrectly.

A nit about the names: I suggest going with `hasProp` vs `hasKey`, as it's more accurate to what the object has (a prop is the combination of a key and it's value). This is also how MDN refers to what `hasOwn` examines ("returns true if the specified object has the indicated **property**..."). 

```ts
declare const DATA: ({ error: ... } | { data: ... })[]

filter(DATA, hasKey("data"));

// vs
filter(DATA, hasProp("data"));
```

I'm assuming you are doing this PR with the assistance of an agent so allowing me to be more verbose with the review; this PR also doesn't handle the lodash/ramda mapping entries at all. Go over our __MISSING files in the docs content folder and add entries for the now implemented functions. PR #515 also addressed the mapping, and you can use that as the guiding example.

---

I'm sorry if this contradicts things that i've said in the review of #515, it has been two years since and some of our ideas around the library might have changed.

lo1tuma:
> This is a great push and really useful utility!
> 
> [...]

Yes I was indeed using AI assistence, great catch. Thanks for the detailed review.

-------------------

I have addressed most of your comments:

  - **Dropped `hasOwnKey`.** Agree on the in/own collapse. `hasKey` now uses
    `Object.hasOwn` at runtime.
  - **Coupled key to T** via `KeysOfUnion<T>`:
    ```ts
    hasKey({ name: "x" }, "naem"); // ts2345 — typo
    hasKey({} as Record<string, unknown>, "anything"); // ok — broad string keys
    ```

  - Narrows by removing optionality. The predicate target is now `T & Required<Pick<T, Key>>` distributed over T's union members. Members that don't have the key drop out, and the matched key loses its ?:
     ```ts    
       // input: { a?: string; b: number }
       if (hasKey(data, "a")) {
         //  data is now { a?: string; b: number } & Required<Pick<…, "a">>
         //  i.e. `a` is required — but you'll see the alias on hover, not the
         //  collapsed `{ a: string; b: number }` form
       }
     ```

  - Mapping entries added for lodash `has`/`hasIn`, ramda `has`/`hasIn`, and
  just `has`. Skipped `hasPath`,  path-based variants stay out of scope, as
  before.

### Pushing back on `hasProp`

Read your reasoning and went the other way, sticking with `hasKey`. Two reasons:

  1. **What we actually check is the key, not the prop.** `Object.hasOwn` answers
     "is this key present",  the value type after narrowing is whatever the input
     type already said it would be (`T[Key]`), the runtime check doesn't tell us
     anything new about it. The MDN phrasing groups the two together because in
     plain JS there's no separate "key" concept, but here we have one and it's
     what the function operates on.
  2. **"Key" is what the rest of the API uses** — `pick`, `omit`, `fromKeys`,
     `keys`, `mapKeys` all talk in keys, not props. `hasKey` slots in
     consistently next to them.

  Happy to rename if you feel strongly; flagging it explicitly so we can settle
  it before merge.

packages/remeda/src/hasOwnKey.ts
lo1tuma: Linked to MDN in the JSDoc.

packages/remeda/src/hasOwnKey.ts
lo1tuma: Dropped that line.

packages/remeda/src/hasOwnKey.ts
lo1tuma: Trimmed.

packages/remeda/src/hasOwnKey.ts
lo1tuma: Switched the related-utilities note to a list.

packages/remeda/src/hasOwnKey.ts
lo1tuma: `Key extends KeysOfUnion<T>` now constrains the key, so `hasKey({ a: 1 } as const, "b")` is a compile-time error. Broad types like `Record<string, unknown>` still accept arbitrary string keys.

packages/remeda/src/hasOwnKey.test.ts
lo1tuma: Dropped the data-first `describe` wrapper.

packages/remeda/src/hasOwnKey.test.ts
lo1tuma: Inlined the `Object.create` call, but kept the `as Record<string, unknown>` cast — without it the new `KeysOfUnion<T>` constraint rejects `"inherited"` since `keyof typeof Object.create(...)` widens to `never`.

packages/remeda/src/hasOwnKey.test.ts
lo1tuma: Same — inlined the construction, kept the `Record<string, unknown>` cast for the same reason.

packages/remeda/src/hasOwnKey.test.ts
lo1tuma: Removed.

packages/remeda/src/hasOwnKey.test.ts
lo1tuma: Grouped under a `describe("arrays", …)` block.

packages/remeda/src/hasOwnKey.test-d.ts
lo1tuma: Dropped.

packages/remeda/src/hasOwnKey.test-d.ts
lo1tuma: Removed.

packages/remeda/src/hasOwnKey.test-d.ts
lo1tuma: Not applicable anymore — the `KeysOfUnion<T>` constraint makes `hasKey({ a: 1 }, "b")` a compile-time error, so there is nothing to narrow against and the test was removed.

packages/remeda/src/hasOwnKey.test-d.ts
lo1tuma: Removed.

packages/remeda/src/hasOwnKey.test-d.ts
lo1tuma: Fixed — both branches now narrow to the matching union member; added the else-branch assertion.

packages/remeda/src/hasOwnKey.test-d.ts
lo1tuma: Switched the predicate target to `T & SetRequired<T, Key>` so accessing `data.a` resolves to `string` (clean). The hover still shows the intersection (`{ a?: string } & { a: string }`) because we don't wrap the whole result in `Simplify` — happy to add it if you'd rather have the cleaner hover, just want to flag the perf cost you raised in #515.

packages/remeda/src/hasOwnKey.test-d.ts
lo1tuma: Same — addressed by the `T & SetRequired<…>` switch in the previous test.

packages/docs/src/content/mapping/just/has.md
```diff
@@ -0,0 +1,17 @@
+---
+category: Objects
+remeda: hasKey
+---
+
+- Just's `has` accepts deep paths (e.g. `["a", "b"]`). Remeda's `hasKey` only
+  checks a single key — for nested paths, narrow step by step.
```
eranhirsch: No need to expand on this, if we don't show a trivial example then we don't mention the solution because it might be hard to implement or not work, we try to only recommend one-liner solutions as drop-in replacements.


```suggestion
- Just's `has` accepts deep paths (e.g. `["a", "b"]`). Remeda's `hasKey` only
  checks a single key.
```

packages/docs/src/content/mapping/just/has.md
```diff
@@ -0,0 +1,17 @@
+---
+category: Objects
+remeda: hasKey
+---
+
+- Just's `has` accepts deep paths (e.g. `["a", "b"]`). Remeda's `hasKey` only
```
eranhirsch: Also mention the dot-notation syntax


```suggestion
- Just's `has` accepts deep paths (e.g. `["a", "b"]`, `"a.b"`). Remeda's `hasKey` only
```

packages/docs/src/content/mapping/just/has.md
```diff
@@ -0,0 +1,17 @@
...
+  checks a single key — for nested paths, narrow step by step.
+- Remeda's `hasKey` is also a type-predicate: it removes optionality from the
+  matched key and drops union members that don't have the key.
```
eranhirsch: This is irrelevant for someone migrating off, the goal of the doc is simply to provide them with instructions on how to port their code. If the source doesn't contain any typing issues, the stronger typing shouldn't result in requiring any changes, so this is not meaningful for a migration.
```suggestion
```

packages/docs/src/content/mapping/lodash/has.md
```diff
@@ -0,0 +1,17 @@
+---
+category: Object
+remeda: hasKey
+---
+
+- Lodash's `has` accepts deep paths (`"a.b.c"` or `["a", "b", "c"]`). Remeda's
+  `hasKey` only checks a single key — for nested paths, narrow step by step.
```
eranhirsch: See my explanation in the just mapping

```suggestion
- Lodash's `has` accepts deep paths (`"a.b.c"` or `["a", "b", "c"]`). Remeda's
  `hasKey` only checks a single key.
```

packages/docs/src/content/mapping/lodash/has.md
```diff
@@ -0,0 +1,17 @@
...
+- `hasKey` is also a type-predicate: it removes optionality from the matched
+  key and drops union members that don't have the key.
```
eranhirsch: See my explanation in the just mapping


```suggestion
```

packages/docs/src/content/mapping/lodash/hasIn.md
```diff
@@ -0,0 +1,20 @@
...
+remeda: hasKey
+---
+
+- Lodash's `hasIn` walks the prototype chain, so it returns `true` for
+  inherited properties (e.g. `hasIn({}, "toString")`). `hasKey` mirrors
+  `Object.hasOwn` and only considers own properties.
+- In modern code, the distinction between own and inherited properties rarely
+  matters — most objects users care about don't sit on a custom prototype, and
+  edge cases like `toString` are usually unintentional matches anyway.
+- `hasIn` accepts deep paths; `hasKey` only checks a single key.
```
eranhirsch: The phrasing here is much more terse than the other mapping files where the same exact point is raised (e.g. they surface an example). Use the fuller version here too

packages/docs/src/content/mapping/lodash/hasIn.md
```diff
@@ -0,0 +1,20 @@
...
+category: Object
+remeda: hasKey
+---
+
+- Lodash's `hasIn` walks the prototype chain, so it returns `true` for
+  inherited properties (e.g. `hasIn({}, "toString")`). `hasKey` mirrors
+  `Object.hasOwn` and only considers own properties.
+- In modern code, the distinction between own and inherited properties rarely
+  matters — most objects users care about don't sit on a custom prototype, and
+  edge cases like `toString` are usually unintentional matches anyway.
```
eranhirsch: Not needed, we don't need to provide rationale for the differences in the mapping file. Link the bullet above to MDN where the reader can learn more if they are unfamiliar with own-property semantics.
```suggestion
```

packages/docs/src/content/mapping/lodash/hasIn.md
```diff
@@ -0,0 +1,20 @@
+---
+category: Object
+remeda: hasKey
+---
+
+- Lodash's `hasIn` walks the prototype chain, so it returns `true` for
+  inherited properties (e.g. `hasIn({}, "toString")`). `hasKey` mirrors
+  `Object.hasOwn` and only considers own properties.
```
eranhirsch: 1. This should be surfaced as an example below, e.g.
2. instead of `walks the prototype chain` refer to the `in` operator specifically and provide a link to MDN for both so the reader can learn more and consider the difference without us restating it.

```ts
// Lodash
hasIn({}, "toString"); //=> true

// Remeda
hasKey({}, "toString"); //=> false
```


```suggestion
- Lodash's `hasIn` uses [the `in` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) which returns `true` for
  **inherited** properties. Remeda uses [`Object.hasOwn`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) which ignores them.
```

packages/docs/src/content/mapping/ramda/has.md
```diff
@@ -0,0 +1,19 @@
+---
+category: Object
+remeda: hasKey
+---
+
+- Both check for an own property using `Object.hasOwn` semantics, so the
+  runtime behavior is identical.
```
eranhirsch: If it's identical it doesn't need to be mentioned

packages/docs/src/content/mapping/ramda/has.md
```diff
@@ -0,0 +1,19 @@
...
+- Remeda's `hasKey` is also a type-predicate: it removes optionality from the
+  matched key and drops union members that don't have the key.
```
eranhirsch: Redundant, see my comments on the other mapping articles

packages/docs/src/content/mapping/ramda/has.md
```diff
@@ -0,0 +1,19 @@
...
+
+```ts
+// Ramda
+has("a", obj);
+R.has("a")(obj);
+
+// Remeda
+hasKey(obj, "a");
+pipe(obj, hasKey("a"));
+```
```
eranhirsch: If the utility is completely identical and could be trusted to work as a drop-in replacement (ignoring the structural differences between remeda and ramda around data-last and data-first, which we assume implicitly) then we don't need to provide any further explanation.


```suggestion
```

packages/docs/src/content/mapping/ramda/hasIn.md
```diff
@@ -0,0 +1,18 @@
+---
+category: Object
+remeda: hasKey
+---
+
+- Ramda's `hasIn` uses the `in` operator and walks the prototype chain;
```
eranhirsch: see my comment for the lodash `hasIn` migration article

packages/remeda/src/hasKey.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @param key - The key to look up. Constrained to keys that could plausibly
+ * exist on `data` so typos surface as type errors.
+ * @signature
+ *   hasKey(key)(data)
+ * @example
+ *   pipe({ a: 1 }, hasKey("a")); //=> true
+ *   filter(
+ *     [{ a: 1 }, { b: 2 }] as Array<{ a?: number; b?: number }>,
+ *     hasKey("a"),
+ *   ); //=> typed as Array<{ a: number; b?: number }>
```
eranhirsch: This example is a bit too elaborate for what it tries to show; because it's main point is to show how the type would be narrowed, but it also tries to show *where* it would be useful and that is over-kill for our examples:


```suggestion
 *   filter([] as { a?: number; }[], hasKey("a")); //=> typed as { a: number }[]
```

packages/remeda/src/hasKey.ts
```diff
@@ -0,0 +1,73 @@
...
+type HasKey<T, Key extends PropertyKey> = T extends unknown
+  ? Key extends keyof T
+    ? T & SetRequired<T, Key>
+    : never
+  : never;
+
+/**
+ * Checks whether `key` is an own property of `data`, mirroring
+ * [`Object.hasOwn`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn).
+ * Inherited properties on the prototype chain are ignored.
```
eranhirsch: 1. The 'own'-ness of the solution is an implementation detail, it should be dropped from the main description line.
2. When our functions proxy built-in functions as-is I try to keep our description as close as possible to the ones in MDN. Following my ask that this function would be renamed as `hasProp`, the description becomes somewhat easier because the function pretty much does what it says on the box
3. We should mention `hasOwn`, but could leave it to the next paragraph so that it isn't part of the one-liner description.


```suggestion
 * Checks if `data` has a prop `key`.
 * 
 * Uses [`Object.hasOwn`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) which only checks own properties, skipping properties inherited through the prototype chain.
```

packages/remeda/src/hasKey.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * - Use `prop` to read the value.
+ * - Use `hasSubObject` to test multiple keys at once.
```
eranhirsch: This is a stretch, hasSubObject checks the shape, not just keys, so it's not a "multi-key" version of this function. I don't think they are related enough to link the two.
```suggestion
```

packages/remeda/src/hasKey.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @param data - The object to test.
+ * @param key - The key to look up. Constrained to keys that could plausibly
+ * exist on `data` so typos surface as type errors.
```
eranhirsch: This is too verbose, this behaviour would be expected.

```suggestion
 * @param key - The key to look up.
```

packages/remeda/src/hasKey.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @signature
+ *   hasKey(data, key)
+ * @example
+ *   hasKey({ a: 1 }, "a"); //=> true
+ *   hasKey({}, "toString"); //=> false (inherited, not own)
+ * @dataFirst
+ * @category Object
```
eranhirsch: category should be Guard


```suggestion
 * @category Guard
```

packages/remeda/src/hasKey.test.ts
```diff
@@ -0,0 +1,82 @@
...
+      Object.assign(Object.create(null) as Record<string, unknown>, { a: 1 }),
+      "a",
+    ),
+  ).toBe(true);
+});
+
+test("returns true for own symbol keys", () => {
+  const symbol = Symbol("test");
+
+  expect(hasKey({ [symbol]: 1 } as Record<symbol, unknown>, symbol)).toBe(true);
```
eranhirsch: why do we need the assertion here?
```suggestion
  expect(hasKey({ [symbol]: 1 }, symbol)).toBe(true);
```

packages/remeda/src/hasKey.test.ts
```diff
@@ -0,0 +1,82 @@
...
+});
+
+test("returns true for numeric keys", () => {
+  expect(hasKey({ 0: "a" } as Record<number, unknown>, 0)).toBe(true);
```
eranhirsch: why do we need the assertion here?


```suggestion
  expect(hasKey({ 0: "a" }, 0)).toBe(true);
```

packages/remeda/src/hasProp.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  const symbol = Symbol("test");
+
+  expect(hasKey({ [symbol]: 1 } as Record<symbol, unknown>, symbol)).toBe(true);
+});
+
+test("returns true for numeric keys", () => {
+  expect(hasKey({ 0: "a" } as Record<number, unknown>, 0)).toBe(true);
+});
+
+describe("arrays", () => {
```
eranhirsch, line 49: In the previous iteration you also tested `length` and `push` which are valid keys of an array, why did you drop those tests, they are valuable

packages/remeda/src/hasKey.test.ts
```diff
@@ -0,0 +1,82 @@
...
+});
+
+describe("arrays", () => {
+  test("returns true for an array index that exists", () => {
+    expect(hasKey([1, 2, 3] as number[], 1)).toBe(true);
```
eranhirsch: why do we need the assertion here?


```suggestion
    expect(hasKey([1, 2, 3], 1)).toBe(true);
```

packages/remeda/src/hasKey.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  test("returns false for an array index past the end", () => {
+    expect(hasKey([1, 2, 3] as number[], 5)).toBe(false);
```
eranhirsch: why do we need the assertion here?

```suggestion
    expect(hasKey([1, 2, 3], 5)).toBe(false);
```

packages/remeda/src/hasKey.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  test("returns false for inherited prototype properties", () => {
+    expect(pipe({} as Record<string, unknown>, hasKey("toString"))).toBe(false);
+  });
+
+  test("filters and narrows union members", () => {
+    const data = [{ a: 1 }, { b: 2 }] as { a?: number; b?: number }[];
+
+    expect(filter(data, hasKey("a"))).toStrictEqual([{ a: 1 }]);
+  });
```
eranhirsch: narrowing is a type concern, this test doesn't add any runtime validations that other tests don't already validate.


```suggestion
```

packages/remeda/src/hasKey.ts
```diff
@@ -0,0 +1,73 @@
...
+import { purry } from "./purry";
+
+// Distribute over `T`'s union members. Each member that has `Key` keeps it
+// (with the optional modifier removed via type-fest's `SetRequired`); members
+// that don't have `Key` drop out because the runtime check rules them out.
+// The leading `T &` makes the result structurally a subtype of `T`, which is
+// required for the type-predicate assignability check.
+type HasKey<T, Key extends PropertyKey> = T extends unknown
+  ? Key extends keyof T
+    ? T & SetRequired<T, Key>
```
eranhirsch: Wrap this with Simplify from type-fest so that the output type is fully resolved.


```suggestion
    ? Simplify<T & SetRequired<T, Key>>
```

packages/remeda/src/hasKey.ts
```diff
@@ -0,0 +1,73 @@
+import type { KeysOfUnion, SetRequired } from "type-fest";
+import { purry } from "./purry";
+
+// Distribute over `T`'s union members. Each member that has `Key` keeps it
+// (with the optional modifier removed via type-fest's `SetRequired`); members
+// that don't have `Key` drop out because the runtime check rules them out.
+// The leading `T &` makes the result structurally a subtype of `T`, which is
+// required for the type-predicate assignability check.
```
eranhirsch: why are all comments front-loaded before the actual code, put relevant documentation above the line it is talking about, and make sure that you don't just describe the line, but provide context about **why** the code looks the way it does. For a type as simple as this, the amount of documentation here seems excessive. The only real non-trivial part is mentioning why we do `T &`, and I don't mind mentioned distributing unions because that's always a footgun.

packages/remeda/src/hasKey.test-d.ts
```diff
@@ -0,0 +1,104 @@
...
+import { filter } from "./filter";
+import { hasKey } from "./hasKey";
+import { pipe } from "./pipe";
+
+test("removes optionality from a known optional property", () => {
+  const data = {} as { a?: string; b: number };
+
+  if (hasKey(data, "a")) {
+    expectTypeOf(data).toEqualTypeOf<
+      { a?: string; b: number } & { b: number; a: string }
```
eranhirsch: This kind of output makes using the utility difficult, see my comment on HasKey to use Simplify, make sure the output types **make sense** to you as a reader.

packages/remeda/src/hasKey.test-d.ts
```diff
@@ -0,0 +1,104 @@
...
+  } else {
+    expectTypeOf(data).toEqualTypeOf<{ a?: string; b: number }>();
+  }
+});
+
+test("does not narrow Record types beyond their index signature", () => {
+  const data = {} as Record<string, number>;
+
+  if (hasKey(data, "foo")) {
+    expectTypeOf(data).toEqualTypeOf<Record<string, number>>();
```
eranhirsch: This type is sub-optimal; once TypeScript knows `foo` is in Record, we want the type to represent that so that `data.foo` is not typed as `number | undefined` (when `exactOptionalPropertyTypes` is enabled, as is in Remeda).

This is probably due to how SetRequired works because type-fest don't assume `exactOptionalPropertyTypes` is enabled, and their types reflect that. Instead, we can build our own simplified version of the type that would probably work better for these cases: `Required<Pick<T, Key>>`

packages/remeda/src/hasKey.test-d.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
```
eranhirsch: add else clauses even whey they are trivial, that way it "locks" our typing against future breakages.

packages/remeda/src/hasKey.test-d.ts
```diff
@@ -0,0 +1,104 @@
...
+test("narrows symbol keys", () => {
+  const symbol = Symbol("test");
+  const data = {} as { [symbol]?: number };
+
+  if (hasKey(data, symbol)) {
+    expectTypeOf(data).toEqualTypeOf<
+      { [symbol]?: number } & { [symbol]: number }
+    >();
+  }
+});
```
eranhirsch: what about number keys?

packages/remeda/src/hasKey.test-d.ts
```diff
@@ -0,0 +1,104 @@
...
+
+test("drops union members that don't have the key", () => {
+  const data = {} as { a: number } | { b: string };
+
+  if (hasKey(data, "a")) {
+    expectTypeOf(data).toEqualTypeOf<{ a: number }>();
+  } else {
+    expectTypeOf(data).toEqualTypeOf<{ b: string }>();
+  }
+});
```
eranhirsch: chefs kiss

packages/remeda/src/hasKey.test-d.ts
```diff
@@ -0,0 +1,104 @@
...
+      | ({ a?: string; tag: "y" } & { tag: "y"; a: string })
+    >();
+  }
+});
+
+test("rejects keys that can't exist on the data", () => {
+  const data = {} as { name: string };
+
+  // @ts-expect-error [ts2345] - "naem" is a typo
+  hasKey(data, "naem");
```
eranhirsch: inline data when it's only used once


  
    
      

```suggestion
  hasKey(
    {} as { name: string }, 
    // @ts-expect-error [ts2345] - "naem" is a typo
    "naem",
  );
```

packages/remeda/src/hasKey.test-d.ts
```diff
@@ -0,0 +1,104 @@
...
+});
+
+test("accepts arbitrary string keys when input is a broad Record", () => {
+  const data = {} as Record<string, unknown>;
+
+  // ok - any string is a possible key on a Record<string, ...>
+  hasKey(data, "literally-anything");
+});
```
eranhirsch: This has already been tested with foo above, we don't need another test for it, even if the premise is different

packages/remeda/src/hasKey.test-d.ts
```diff
@@ -0,0 +1,104 @@
...
+    expectTypeOf(filter(data, hasKey("a"))).toEqualTypeOf<
+      ({ a?: number; b?: number } & { b?: number; a: number })[]
+    >();
+  });
+
+  test("rejects keys that can't exist on the piped data", () => {
+    const data = {} as { name: string };
+
+    // @ts-expect-error [ts2345] - "naem" is a typo
+    pipe(data, hasKey("naem"));
```
eranhirsch: inline data

packages/remeda/src/hasKey.test-d.ts
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  test("accepts arbitrary string keys when piped data is a broad Record", () => {
+    const data = {} as Record<string, unknown>;
+
+    // ok
+    pipe(data, hasKey("literally-anything"));
+  });
```
eranhirsch: prefer a more robust test using filter and checks the output type

packages/remeda/src/hasKey.test-d.ts
```diff
@@ -0,0 +1,104 @@
+import { describe, expectTypeOf, test } from "vitest";
```
eranhirsch: I'm missing some tests:

- check a union of keys for the key type, all tests test a single literal key (e.g., `hasKey({} as { a?: string; b?: number }, "a" | "b")`).
- check primitive keys (e.g., `string`).
- an always exists test (e.g., `hasKey({ a: string }, 'a')`);

packages/remeda/src/hasKey.ts
```diff
@@ -0,0 +1,73 @@
...
+    ? T & SetRequired<T, Key>
+    : never
+  : never;
+
+/**
+ * Checks whether `key` is an own property of `data`, mirroring
+ * [`Object.hasOwn`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn).
+ * Inherited properties on the prototype chain are ignored.
+ *
+ * - Use `prop` to read the value.
```
eranhirsch: Add a line to `prop`s jsdoc similar to this one to refer people to here to close the loop.

eranhirsch (changes requested):
Thanks for following up on my comments, the remaining issues are mainly nits or cleanups.

### Regarding the name
We should use **`hasProp`** because our naming convention says to use official names when those exist; the underlying function we use (`hasOwn`) [uses *Property* since inception](https://tc39.es/proposal-accessible-object-hasownproperty/) but final implementation had to drop the *Property* part because it collided with `Object.prototype.hasOwnProperty`. We don't need the *own* part because it's a low-level detail that isn't relevant to our offering, and *Prop* instead of *Property* because that's what we already used in `prop`, `groupByProp`, `addProp`, etc...

packages/docs/src/content/mapping/just/has.md
lo1tuma: Done.

packages/docs/src/content/mapping/just/has.md
lo1tuma: Added the dot-notation form.

packages/docs/src/content/mapping/just/has.md
lo1tuma: Removed.

packages/docs/src/content/mapping/lodash/has.md
lo1tuma: Done.

packages/docs/src/content/mapping/lodash/has.md
lo1tuma: Removed.

packages/docs/src/content/mapping/lodash/hasIn.md
lo1tuma: Expanded to match the `lodash/has.md` phrasing.

packages/docs/src/content/mapping/lodash/hasIn.md
lo1tuma: Removed the rationale bullet.

packages/docs/src/content/mapping/lodash/hasIn.md
lo1tuma: Restructured the first bullet around the `in` / `hasOwn` MDN links and added a `toString` example below.

packages/docs/src/content/mapping/ramda/has.md
lo1tuma: Dropped.

packages/docs/src/content/mapping/ramda/has.md
lo1tuma: Dropped.

packages/docs/src/content/mapping/ramda/has.md
lo1tuma: Dropped — file is now just frontmatter since the mapping is a clean drop-in.

packages/docs/src/content/mapping/ramda/hasIn.md
lo1tuma: Applied the same restructure — MDN links and a `toString` example replace the prose rationale.

packages/remeda/src/hasKey.ts
lo1tuma: Applied the compact form.

packages/remeda/src/hasKey.ts
lo1tuma: Folded — renamed to `hasProp` across the source, tests, and migration docs, and rewrote the JSDoc description using your wording verbatim.

packages/remeda/src/hasKey.ts
lo1tuma: Dropped — agreed the framing oversold the relationship.

packages/remeda/src/hasKey.ts
lo1tuma: Trimmed both `@param` blocks.

packages/remeda/src/hasKey.ts
eranhirsch: notice i had a typo originally, I just fixed it in the suggestion, but you might have not picked it up. It's just the "the" before `data`, e.g. "Checks if `data`" and not "Checks if the `data`"

packages/remeda/src/hasKey.ts
lo1tuma: Switched both to `@category Guard`.

packages/remeda/src/hasKey.ts
lo1tuma: Wrapped with `Simplify` and switched the inner from `SetRequired` to `Required<Pick<T, Key>>` per comment further down (the two changes are tightly coupled). Narrowed hover is now `{ a: string; b: number }` for the optional case and `{ [x: string]: T; foo: T }` for the Record case.

packages/remeda/src/hasKey.ts
lo1tuma: Inlined the comments next to the non-obvious lines (union distribution + the `T &` assignability trick) and dropped the rest.

packages/remeda/src/hasKey.ts
lo1tuma: Added `To check whether a key exists on the object, use `hasProp`.` to both `prop` JSDoc blocks.

packages/remeda/src/hasKey.test.ts
lo1tuma: Dropped the cast — leftover from an earlier iteration.

packages/remeda/src/hasKey.test.ts
lo1tuma: Dropped the cast — leftover from an earlier iteration.

packages/remeda/src/hasKey.test.ts
lo1tuma: Dropped the cast — leftover from an earlier iteration.

packages/remeda/src/hasKey.test.ts
lo1tuma: Dropped the cast — leftover from an earlier iteration.

packages/remeda/src/hasProp.test.ts
lo1tuma: Brought both back.

packages/remeda/src/hasKey.test.ts
lo1tuma: Removed.

packages/remeda/src/hasKey.test-d.ts
lo1tuma: Resolved by the `Simplify` + `Required<Pick<>>` switch — the assertion now expects the clean shape (`{ a: string; b: number }` etc.). See the reply on the `HasProp` thread.

packages/remeda/src/hasKey.test-d.ts
lo1tuma: Switched the inner narrowing from type-fest's `SetRequired` to `Required<Pick<T, Key>>` as you suggested — the Record assertion now narrows to `{ [x: string]: number; foo: number }`, which keeps `data.foo` typed as `number` (no `| undefined`).

packages/remeda/src/hasKey.test-d.ts
lo1tuma: Added the trivial else clause where it was missing.

packages/remeda/src/hasKey.test-d.ts
lo1tuma: Added a parallel `narrows number keys` test below the symbol one.

packages/remeda/src/hasKey.test-d.ts
lo1tuma: 🙏

packages/remeda/src/hasKey.test-d.ts
lo1tuma: Inlined.

packages/remeda/src/hasKey.test-d.ts
lo1tuma: Removed.

packages/remeda/src/hasKey.test-d.ts
lo1tuma: Inlined.

packages/remeda/src/hasKey.test-d.ts
lo1tuma: Rewrote it as a filter-based test that asserts the narrowed item shape (`Array<{ [x: string]: unknown; foo: unknown }>`).

packages/remeda/src/hasKey.test-d.ts
lo1tuma: Added the three tests. Verified shapes against the compiler:  - **Always-exists key** (`{ a: string }` + `"a"`): if-branch is unchanged, else-branch narrows to `never` since the predicate type equals the input type. - **Union of literal keys** (`"a" | "b"` on `{ a?: string; b?: number }`): narrowing distributes — if-branch is `{ a: string; b?: number } | { a?: string; b: number }`, else-branch is the original union. - **Primitive `string` key** on a `Record<string, number>`: no narrowing in the if-branch (the predicate type equals the input), else-branch is `never` for the same reason as the always-exists case.

packages/remeda/src/hasKey.ts
lo1tuma: Fixed — dropped the stray "the" in both blocks.

lo1tuma:
Thanks again for the detailed review — all comments have been addressed in df6dbe7.

packages/remeda/src/hasProp.ts
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * - Use `prop` to read the value.
+ *
+ * @param data - The object to test.
+ * @param key - The key to look up.
+ * @signature
+ *   hasProp(data, key)
+ * @example
+ *   hasProp({ a: 1 }, "a"); //=> true
+ *   hasProp({}, "toString"); //=> false (inherited, not own)
```
eranhirsch: nit: This example made sense when we had two versions of the function, now it just confuses.

```suggestion
```

packages/remeda/src/hasProp.test-d.ts
```diff
@@ -0,0 +1,149 @@
...
+  const data = {} as Record<string, number>;
+  const key = "foo" as string;
+
+  if (hasProp(data, key)) {
+    expectTypeOf(data).toEqualTypeOf<Record<string, number>>();
+  } else {
+    // The predicate type equals the input type, so TS proves the else
+    // branch is unreachable and narrows to `never`.
+    expectTypeOf(data).toEqualTypeOf<never>();
+  }
```
eranhirsch, line 87: I'm glad we added this test — it surfaces a real issue with the current type. The `never` on the else branch prevents this predicate from being useful for unbounded keys, and they're common enough that we need to address it.

What we need to do is "downgrade" the type-predicate to a `boolean` to stop TypeScript from narrowing incorrectly. We do this by adding an additional overload to the function that returns `boolean`, and allow TypeScript to route call sites by making the incompatible version uncallable — its parameter becomes `never` conditionally. We route all non-bounded keys to this overload because only bounded (finite) key types can be narrowed meaningfully.

```ts
export function hasProp<T extends object, Key extends KeysOfUnion<T>>(
  data: T,
  key: IsBounded<Key> extends true ? Key : never,
): data is HasProp<T, Key>;
export function hasProp<T extends object>(
  data: T,
  key: KeysOfUnion<T>,
): boolean;
```

Add a test for unbounded string literals (e.g. `prefix_${string}`) to lock the contract.

packages/remeda/src/hasProp.test-d.ts
```diff
@@ -0,0 +1,149 @@
...
+  );
+});
+
+describe("data-last", () => {
+  test("narrows union members under filter", () => {
+    const data = [] as { a?: number; b?: number }[];
+
+    expectTypeOf(filter(data, hasProp("a"))).toEqualTypeOf<
+      { a: number; b?: number }[]
+    >();
```
eranhirsch: prefer inlining variables that aren't used more than once


```suggestion
    expectTypeOf(filter(
      [] as { a?: number; b?: number }[],
      hasProp("a"),
    )).toEqualTypeOf<{ a: number; b?: number }[]>();
```

packages/remeda/src/hasProp.test-d.ts
```diff
@@ -0,0 +1,149 @@
+import { describe, expectTypeOf, test } from "vitest";
```
eranhirsch, line 1: I'm missing type tests for:

- the matching runtime array tests, I don't know exactly how our types would behave with an array input.
- interfaces (which are unique in the sense that they don't extend `Record<PropertyKey, unknown>`, unlike types) often cause issues for us if we aren't careful (i don't think this type would have an issue, but the test locks it in).
- `undefined` as a value of an optional prop. This is non-trivial when considering `exactOptionalPropertyTypes` so we need to have the contract written out.

packages/docs/src/content/mapping/ramda/hasIn.md
```diff
@@ -0,0 +1,21 @@
...
+```ts
+// Ramda
+hasIn("a", obj);
+
+// Remeda
+hasProp(obj, "a");
+
+// Inherited properties
+hasIn("toString", {}); //=> true
+hasProp({}, "toString"); //=> false
```
eranhirsch: nit: Same as lodash/hasIn — each example should be its own `####` section with its own code block.

packages/docs/src/content/mapping/lodash/hasIn.md
```diff
@@ -0,0 +1,23 @@
...
+```ts
+// Lodash
+hasIn(obj, "a");
+
+// Remeda
+hasProp(obj, "a");
+
+// Inherited properties
+hasIn({}, "toString"); //=> true
+hasProp({}, "toString"); //=> false
```
eranhirsch: nit: Each example should be its own `####` section with its own code block (see `lodash/defaultTo.md` for the pattern). Split the inherited-properties demo into a separate `#### Inherited properties` section.

eranhirsch (changes requested):
I'm happy that we added the primitive string test, it surfaced a real issue that I didn't think about when reviewing the code. I added the fix in that comment, it should be trivial to patch up.

packages/remeda/src/hasProp.ts
lo1tuma: Applied.

packages/remeda/src/hasProp.test-d.ts
lo1tuma: Applied your suggested overload pair with `IsBounded<Key>` gating. Updated the `accepts a primitive string key` test to reflect that the unbounded overload now returns plain `boolean` (no `never` else branch) and added a sibling test for template-literal keys (`\`prefix_${string}\``) to lock the contract.

packages/remeda/src/hasProp.test-d.ts
lo1tuma: While writing the array tests you asked for, I hit a type/runtime mismatch I want to surface before locking the contract.

For prototype methods on arrays (and Map/Set/Date/etc.), the type system can't distinguish own from inherited properties:

```ts
const arr = [] as number[];

if (hasProp(arr, "push")) {
  // type: number[]  ← narrowing fires (push is in keyof number[])
} else {
  // type: never  ← TS thinks unreachable
  // runtime: actually fires (Object.hasOwn([], "push") is false)
}
```

Did a survey to see if anyone solves this:

- **TypeScript itself** ships `Object.hasOwn(o, v): boolean` — no predicate. Microsoft closed the narrowing proposal ([microsoft/TypeScript#58877](https://github.com/microsoft/TypeScript/issues/58877)) as **"not planned"** for exactly this reason.
- **ts-extras's [`objectHasOwn`](https://github.com/sindresorhus/ts-extras/blob/HEAD/source/object-has-own.ts)** uses `Key extends PropertyKey` (no `keyof T` constraint) and `ObjectType & Record<Key, unknown>` (no `Required<Pick>` optionality removal). The `push`/`never` collapse can't happen with that shape, but it gives up both the typo-catching and the clean optionality-removal we deliberately added in rounds 1–2.
- **Effect's [`Predicate.hasProperty`](https://effect-ts.github.io/effect/effect/Predicate.ts.html)** takes the same shape as ts-extras.
- **`Object.prototype.hasOwnProperty` narrowing ([microsoft/TypeScript#21732](https://github.com/microsoft/TypeScript/issues/21732))** is also widely-discussed-and-unsolved.

So it looks like a true three-way trade and nobody has all three:

| Property                            | ts-extras / Effect | hasProp (current) |
| ----------------------------------- | ------------------ | ----------------- |
| Typo catching on `Key`              | ✗                  | ✓                 |
| Optionality removal                 | ✗                  | ✓                 |
| Correct prototype-method ergonomics | ✓                  | ✗                 |

Three paths from here:

1. **Adopt the ts-extras shape** — drop the `KeysOfUnion<T>` constraint and switch the predicate to `T & Record<Key, unknown>`. Loses the two properties you asked for in earlier rounds; gains correct prototype-method behavior.
2. **Keep current design, document the limitation** — add the `push` test with an inline comment + link to microsoft/TypeScript#58877. Locks the (technically wrong) behavior so we catch regressions and the constraint is visible in code.
3. **Filter function-valued keys via `T[Key] extends (...) => unknown`** in the bounded overload's `key` type. Routes `push`/`pop`/etc. to the boolean overload, fixing arrays. Trade-off: legitimate callback-shaped props (`{ onClick?: () => void }`) also lose narrowing. No library ships this — flagged in blog posts as "this isn't quite right".

I'd lean toward (2) — it matches the ecosystem's stance and keeps the design properties you asked for. But this is your call.

packages/docs/src/content/mapping/ramda/hasIn.md
lo1tuma: Split into `#### Own properties` and `#### Inherited properties` sections, matching the `defaultTo.md` pattern.

packages/docs/src/content/mapping/lodash/hasIn.md
lo1tuma: Split into `#### Own properties` and `#### Inherited properties` sections, matching the `defaultTo.md` pattern.

packages/remeda/src/hasProp.test-d.ts
eranhirsch: There's also option 4:

Move arrays to the non-type-narrowing overload so that the function is still usable for them at the runtime level without handling them at the type level.

This also makes sense conceptually because there are no practical use-cases for calling `hasProp` on arrays, beyond those that would be served better by other functions we offer:
- If the user wants to access an index safely with `noUncheckedIndexedAccess` they should use `hasAtLeast` which narrows correctly.
- If the user wants to discriminate between an array and an object in a union (e.g., `{ a: string } | number[]`) they should use `isArray` and not `hasProp(data, "length")`.

We can achieve this with:
```ts
type HasProp<T, Key extends PropertyKey> = T extends readonly unknown[]
  ? never
  : ...

export function hasProp<T extends object, Key extends KeysOfUnion<T>>(
  data: T extends readonly unknown[] ? never : T,
  key: IsBounded<Key> extends true ? Key : never,
): data is HasProp<NonArray<T>, Key>;
```

One caveat is that we can't change `T extends object` in the type-param declaration to `T extends Record<PropertyKey, unknown>` because that would block using the function with interfaces (that's why I asked for the interface test).

Another thing to notice, is that this solution still passes `Map`, `Set`, `Date`, `RegExp` and other built-in types through the narrowing variant, and that would probably be the case for custom classes too. We can start blocking each case with an additional `never` branch, but I think that the function is unlikely to be used with them so the risk is very small. If you want, you can add tests to a describe block for known limitations which simply communicates the fact that we know this isn't ideal.

lo1tuma: Going with Option 4 — thanks, that's clean. Applied:

- `T extends readonly unknown[] ? never : T` on the data parameter of the narrowing overload (both data-first and data-last), so arrays route to the `boolean`-returning fallback.
- Same conditional mirrored at the top of the `HasProp` type, so the type-predicate assignability check stays satisfied. (Without the mirror, `Simplify<NonArray<T> & …>` doesn't structurally match `NonArray<T>` and TS rejects the predicate.)
- Added the array, interface, and `{ a?: string | undefined }` tests you asked for. All three narrow cleanly now.
- Added a `describe("known limitations", …)` block locking the Map/Set behavior — both still narrow the else branch to `never` for prototype methods. Comment in the block links microsoft/TypeScript#58877.

eranhirsch (approved):
that's a wrap! thanks for pushing on this to the finish line! I'm happy we now have this utility

PR #1351, opened 2026-05-21
fix(prop): add undefined for index-signature key access

### LINKED ISSUE #1274: `prop`: missing `undefined` when accessing simple Records
When the prop path contains a simple Record (e.g. `Record<string, MyObject>`) it doesn't properly add the `| undefined` to the type.

See: 

```ts
import { prop } from "remeda";

declare const DATA: {
  a: Record<string, { b: { c: "hello"; d: boolean; e: string } }>;
};


const x0 = prop(DATA, "a");
//    ^? Record<string, { b: { c: "hello"; d: boolean; e: string } }>

const x1 = prop(DATA, "a", "x");
//    ^? { b: { c: "hello"; d: boolean; e: string } }

const x2 = prop(DATA, "a", "x", "b");
//    ^? { c: "hello"; d: boolean; e: string }

const x3 = prop(DATA, "a", "x", "b", "c");
//    ^? "hello"

const x4 = prop(DATA, "a", "x", "b", "d");
//    ^? boolean

const x5 = prop(DATA, "a", "x", "b", "e");
//    ^? string
```

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzmKExwL5wGapHAIigFMRiATAQwIG4AoO84gYwBtKS5mIA7AZ3gARAIIAVYQC5EdOHEpSASi2jkAPAKjAeAcwA0iOACMpSZlIIALYq1YRacclMMQIrYpR404xKRq3bMTAA+egx6Bm5+eAAPAAY4AF5kVDAAChFxfQJqAEp6AHp82VkAPQB+CN4BOGiARkTktHSxYSzqLOiCPLpC4rhyyqiagCYGlCaM1sJ2wk6swy6CouKBukjq6IBmMZTmzOmCDsPCBazmRZ7l0oq1qpiAFh2JlrbjgjmTt-IL3pWb9ZiAFYnmlJq8jvM3sQfld+mUgA

ATOM00blue:
## Problem
Accessing a key of a `Record<string, V>` (or any index-signature type) via `prop` returned `V` instead of `V | undefined`. That is unsound: an arbitrary `string` key is not guaranteed to exist at runtime, so the value can be `undefined`.

## Fix
Use `OmitIndexSignature<T>` to distinguish explicitly-declared properties from index-signature-only keys. Explicit keys keep their exact type (`T[Key]`); keys matched only by an index signature get `T[Key] | undefined`. This mirrors `noUncheckedIndexedAccess` but is applied within `prop` itself, so the result is sound regardless of the consumer's `tsconfig`.

## Tests
Added type tests in `prop.test-d.ts` asserting `prop` on a `Record` index access yields `V | undefined`, while access of known object properties is unchanged.

Closes #1274

### DIFF
--- a/packages/remeda/src/prop.test-d.ts
+++ b/packages/remeda/src/prop.test-d.ts
@@ -421,6 +421,110 @@ describe("optional props", () => {
   });
 });
 
+describe("simple records (Issue #1274)", () => {
+  describe("data-first", () => {
+    test("string keys add undefined", () => {
+      expectTypeOf(prop({} as Record<string, number>, "a")).toEqualTypeOf<
+        number | undefined
+      >();
+    });
+
+    test("number keys add undefined", () => {
+      expectTypeOf(prop({} as Record<number, string>, 1)).toEqualTypeOf<
+        string | undefined
+      >();
+    });
+
+    test("union keys add undefined", () => {
+      expectTypeOf(
+        prop({} as Record<string | number, boolean>, "a"),
+      ).toEqualTypeOf<boolean | undefined>();
+    });
+
+    test("template-literal keys add undefined", () => {
+      expectTypeOf(
+        prop({} as Record<`id_${number}`, string>, "id_1"),
+      ).toEqualTypeOf<string | undefined>();
+    });
+
+    test("literal values add undefined", () => {
+      expectTypeOf(prop({} as Record<string, 1 | 2>, "a")).toEqualTypeOf<
+        1 | 2 | undefined
+      >();
+    });
+
+    test("a union of keys is still missing", () => {
+      expectTypeOf(
+        prop({} as Record<string, number>, "a" as "a" | "b"),
+      ).toEqualTypeOf<number | undefined>();
+    });
+
+    test("nested records propagate undefined down the path", () => {
+      const data = {} as {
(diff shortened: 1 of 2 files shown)


eranhirsch (changes requested):
Hi, and thanks for the contribution.

I'm Remeda's only maintainer, and reviewing a type-system change like this thoroughly takes hours — that cost only makes sense when there's a motivated human on the other side of the review.

Looking at your PR history, this submission doesn't read that way: it landed single-shot with no engagement on #1274, as part of a broader sweep of low-hanging open issues across popular repos. I'm not opposed to agent assistance — Remeda ships a contributor harness for exactly that — but contributions here must be **driven** by someone who *needs* the fix — they're hitting the bug, or they care about this specific issue. Our open issues aren't a backlog to harvest.

Thank you for your tokens, but generating fixes isn't the bottleneck. I can run an agent against any of our open issues myself. The bottleneck is validating that the fix is correct, and that requires a collaborator who will iterate with me through review — this PR doesn't offer that.

If you're personally invested in #1274, reopen with a response and I'll work through the review with you. Otherwise, closing.

eranhirsch:
agent-driven sweep with 0% human involvement.

PR #1388, opened 2026-07-03
chore(deps): major bumps: astro 7 (@astrojs/mdx 7) and eslint 10 (eslint-plugin-astro 2)

eranhirsch:
Takes the two dependabot major bumps (#1384, #1385) together with the
changes they actually require to keep CI green:

- @astrojs/mdx 7 peer-depends on astro ^7.0.0, so the docs site moves to
  Astro 7, which is built on Vite 8; the root vite override moves to ~8.1.3
  (vitest 4 already supports Vite 8).
- eslint-plugin-astro 2 requires ESLint 10, so the docs package moves to
  eslint ~10.2.0 (matching the library package). eslint-plugin-react and
  eslint-plugin-jsx-a11y don't declare ESLint 10 support yet and still call
  rule-context APIs removed in v10, so they are wrapped with @eslint/compat's
  fixupPluginRules and their eslint peer is overridden at the root.
- The docs package's typescript moves to ^6.0.2 to match the library. With
  two TypeScript majors in the workspace the astro parser built types with
  TS 5.9 while typescript-eslint read TS 6's renumbered TypeFlags, crashing
  no-unsafe-enum-comparison (EnumLiteral in 6.0 reuses 5.9's Undefined bit).
- TS 6 resolves side-effect imports (ts2882), so the bare @docsearch/css
  import (which only has a css "main") becomes an explicit .css file path
  covered by the ambient *.css module declarations.
- With eslint 10 hoisted for the whole workspace, @vitest/eslint-plugin's
  types now check cleanly, so the @ts-expect-error in the library's eslint
  config is no longer used and is removed.

The lockfile is regenerated from scratch: the surgical update kept nesting
single-version packages (astro, typescript) under packages/docs, which broke
type resolution for hoisted packages that import them.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01NbKGDmWjqaFsdL5JNLpib3

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -58,9 +58,9 @@
       }
     },
     "node_modules/@actions/http-client/node_modules/undici": {
-      "version": "6.25.0",
-      "resolved": "https://registry.npmjs.org/undici/-/undici-6.25.0.tgz",
-      "integrity": "sha512-ZgpWDC5gmNiuY9CnLVXEH8rl50xhRCuLNA97fAUnKi8RRuV4E6KG31pDTsLVUKnohJE0I3XDrTeEydAXRw47xg==",
+      "version": "6.27.0",
+      "resolved": "https://registry.npmjs.org/undici/-/undici-6.27.0.tgz",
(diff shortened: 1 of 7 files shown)


eranhirsch:
Redundant with #1389

PR #1411, opened 2026-08-10
feat(funnel): Add `peakPreparedData` Return Value

joealden:
Towards https://github.com/remeda/remeda/discussions/1391

---

Make sure that you:

- [ ] New methods added to `docs/src/content/mapping` for both Lodash and Ramda.


eranhirsch (changes requested):
Hey, Thanks for contributing and sorry for taking longer to get feedback back to you.

I think this is a good addition and am happy to see it through, but we need to change a few things so it fits better in our library:
1. The structured clone is problematic as it only works if the preparedData supports it, and there are a lot of real-world cases where it doesn't. It also adds additional runtime cost that might not be warranted. Node (and other runtimes!) being largely single-threaded, within a single frame the data being observed can't change, so there's no need to clone it ahead of observing it, unless you also want to mutate it, and in that case the caller can wrap it with their own structuredClone, and they also know that the data can be cloned safely.
2. Peak is the summit of a mountain, peek is taking a look. Regardless, for this kind of data we should use the same pattern we use for `isIdle` and provide the functionality as a readonly accessor (e.g. `get preparedData() {}`). We can even consider renaming the concept to `pendingData` to better signal the in-flight nature of it.
3. There's an edge case when this is used without any reducer, where the preparedData would be our internal VOID_REDUCER_SYMBOL which we **don't** want to expose, we'll need to check for that value and return `undefined` instead.
4. We need to document this! it needs to be added to the JSDoc block.
5. Perhaps add a few more tests, with some interesting edge-cases.

Thanks again for the contribution!

joealden:
Sorry - didn't see your response! Will try to get this updated this weekend.

1. Sounds good - I was just copying what TanStack Pacer was doing, but those are all good points.
2. 🤦 As an native English speaker, I should've caught that! `pendingData` sounds good - I borrowed the naming from TanStack Pacer too, but "peek" doesn't make sense if we arent going to be cloning it.
3. 👍 
4. Sounds good - I thought I mirrored what we were doing elsewhere for returned values, but I presume I missed somewhere.
5. What exactly do you mean by "interesting edge-cases"? I can add a test for the no reducer case, but if we're not going to be cloning anymore, what other edge cases are you thinking of?

PR #1439, opened 2026-09-10
fix(endsWith, startsWith): reject disjoint prefix/suffix instead of disabling the return type

### LINKED ISSUE #1432: `startsWith`: Disallow prefix that is not a member of the string literal union
For `startsWith` / `endsWith` operating on a string literal union, disallow passing a prefix/suffix that is not a member of said union.

`startsWith` / `endsWith` operating on a plain string are not affected.

The benefit of this is to warn against stale code.

e.g.:

```ts
const something = "something-lol" | "something-lmao" | "something-rofl"
if (startsWith(something, "someOtherThing-") { // type error
  // ...
}
```

```patch
diff --git a/dist/index.d.ts b/dist/index.d.ts
index f3f37ca065690cdf69b0be38a0722edbf4e7005b..0ab0ba97514631f2a903948baf04478efb89bad0 100644
--- a/dist/index.d.ts
+++ b/dist/index.d.ts
@@ -6448,6 +6448,13 @@ declare function dropWhile<T extends IterableContainer>(data: T, predicate: (ite
 declare function dropWhile<T extends IterableContainer>(predicate: (item: T[number], index: number, data: T) => boolean): (data: T) => T[number][];
 //#endregion
 //#region src/endsWith.d.ts
+/**
+ * Reported instead of `never` when a literal affix cannot match, so
+ * the compiler error names the affix that has no possible match.
+ */
+type NoUnionMemberMatchesAffix<Affix extends string> = {
+  "remeda-affix-mismatch": `No member of this string literal union matches "${Affix}"`;
+};
 /**
  * Determines whether a string ends with the provided suffix, and refines the
  * output type if possible.
@@ -6466,8 +6473,8 @@ declare function dropWhile<T extends IterableContainer>(predicate: (item: T[numb
  * @dataFirst
  * @category String
  */
-declare function endsWith<T extends string, Suffix extends string>(data: T, suffix: string extends Suffix ? never : Suffix): data is T & `${string}${Suffix}`;
-declare function endsWith(data: string, suffix: string): boolean;
+declare function endsWith<T extends string, Suffix extends string>(data: T, suffix: string extends Suffix ? never : string extends T ? Suffix : [Extract<T, `${string}${Suffix}`>] extends [never] ? NoUnionMemberMatchesAffix<Suffix> : Suffix): data is T & `${string}${Suffix}`;
+declare function endsWith<Suffix extends string>(data: string, suffix: string extends Suffix ? Suffix : never): boolean;
 /**
  * Determines whether a string ends with the provided suffix, and refines the
  * output type if possible.
@@ -11470,8 +11477,8 @@ declare function splitWhen<T>(predicate: (item: T, index: number, data: readonly
  * @dataFirst
  * @category String
  */
-declare function startsWith<T extends string, Prefix extends string>(data: T, prefix: string extends Prefix ? never : Prefix): data is T & `${Prefix}${string}`;
-declare function startsWith(data: string, prefix: string): boolean;
+declare function startsWith<T extends string, Prefix extends string>(data: T, prefix: string extends Prefix ? never : string extends T ? Prefix : [Extract<T, `${Prefix}${string}`>] extends [never] ? NoUnionMemberMatchesAffix<Prefix> : Prefix): data is T & `${Prefix}${string}`;
+declare function startsWith<Prefix extends string>(data: string, prefix: string extends Prefix ? Prefix : never): boolean;
 /**
  * Determines whether a string begins with the provided prefix, and refines the
  * output type if possible.
```

eranhirsch:
Following my recent changes in #1434 and the follow up comment: https://github.com/remeda/remeda/issues/1432#issuecomment-5610099531 I took another look at what cases are common and where we might have been hedging our type too broadly.

To give the best experience for detecting typos and dead code we now fail directly at the param level instead of falling to the return value being unusable; similar to how `prop`, `pick`, `drop`, etc... would reject disjoint prop names.

This was done via restructuring our data-first fallbacks so that a disjoint prefix/suffix have no target to fall back to, so they get fully rejected instead.

Closes: #1432


packages/remeda/src/endsWith.test-d.ts
```diff
@@ -1,5 +1,4 @@
 import { describe, expectTypeOf, test } from "vitest";
-import { $typed } from "../test/$typed";
 import { endsWith } from "./endsWith";
```
eranhirsch:
```suggestion
import { $typed } from "../test/$typed";
import { endsWith } from "./endsWith";
```

packages/remeda/src/endsWith.test-d.ts
```diff
@@ -51,6 +50,17 @@ describe("data-first", () => {
...
+    const data = "a.test.ts" as `${string}.test.ts`;
+    if (endsWith(data, "a.test.ts")) {
+      expectTypeOf(data).toEqualTypeOf<
+        `${string}.test.ts` & `${string}a.test.ts`
+      >();
+    } else {
+      expectTypeOf(data).toEqualTypeOf<`${string}.test.ts`>();
+    }
+  });
+
```
eranhirsch:
```suggestion

```

packages/remeda/src/endsWith.test-d.ts
```diff
@@ -211,6 +215,18 @@ describe("data-last", () => {
...
+      [] as `${string}.test.ts`[],
+      endsWith("a.test.ts"),
+    );
+
+    expectTypeOf(yes).toEqualTypeOf<
+      (`${string}.test.ts` & `${string}a.test.ts`)[]
+    >();
+    expectTypeOf(no).toEqualTypeOf<`${string}.test.ts`[]>();
+  });
+
```
eranhirsch:
```suggestion

```

packages/remeda/src/endsWith.test-d.ts
```diff
@@ -333,27 +349,43 @@ describe("data-last", () => {
 // @see https://github.com/remeda/remeda/issues/1432
 describe("reject disjoint suffixes (#1432)", () => {
   test("const data that doesn't match", () => {
-    // @ts-expect-error [ts1345] -- Intentional! we check that the result of `endsWith` cannot be coerced to a boolean value!
-    // eslint-disable-next-line @typescript-eslint/no-unused-expressions, @typescript-eslint/strict-boolean-expressions -- Intentional! this form is the most concise way to test what we need.
-    !endsWith("helloworld" as const, "foo");
+    endsWith(
+      "helloworld" as const,
+      // @ts-expect-error [ts2769] -- Intentional! this is what we're testing...
```
eranhirsch:
```suggestion
      // @ts-expect-error [ts2769] -- Neither data-first overload accepts a
      // dead suffix, so the error lands here instead of on the return type.
```

packages/remeda/src/endsWith.test-d.ts
```diff
@@ -333,27 +349,43 @@ describe("data-last", () => {
...
+    );
   });
 
   test("literal union where no member matches", () => {
-    // @ts-expect-error [ts1345] -- Intentional! we check that the result of `endsWith` cannot be coerced to a boolean value!
-    // eslint-disable-next-line @typescript-eslint/no-unused-expressions, @typescript-eslint/strict-boolean-expressions -- Intentional! this form is the most concise way to test what we need.
-    !endsWith("cat" as "cat" | "dog", "bird");
+    endsWith(
+      "cat" as "cat" | "dog",
+      // @ts-expect-error [ts2769] -- Intentional! this is what we're testing...
```
eranhirsch:
```suggestion
      // @ts-expect-error [ts2769] -- Neither data-first overload accepts a
      // dead suffix, so the error lands here instead of on the return type.
```

packages/remeda/src/endsWith.test-d.ts
```diff
@@ -333,27 +349,43 @@ describe("data-last", () => {
...
+    );
   });
 
   test("union suffix where no member matches", () => {
-    // @ts-expect-error [ts1345] -- Intentional! we check that the result of `endsWith` cannot be coerced to a boolean value!
-    // eslint-disable-next-line @typescript-eslint/no-unused-expressions, @typescript-eslint/strict-boolean-expressions -- Intentional! this form is the most concise way to test what we need.
-    !endsWith("cat" as "cat" | "dog", "bird" as "bird" | "fish");
+    endsWith(
+      "cat" as "cat" | "dog",
+      // @ts-expect-error [ts2769] -- Intentional! this is what we're testing...
```
eranhirsch:
```suggestion
      // @ts-expect-error [ts2769] -- Neither data-first overload accepts a
      // dead suffix, so the error lands here instead of on the return type.
```

packages/remeda/src/endsWith.test-d.ts
```diff
@@ -333,27 +349,43 @@ describe("data-last", () => {
...
+    );
   });
 
   test("template suffix that no literal matches", () => {
-    // @ts-expect-error [ts1345] -- Intentional! we check that the result of `endsWith` cannot be coerced to a boolean value!
-    // eslint-disable-next-line @typescript-eslint/no-unused-expressions, @typescript-eslint/strict-boolean-expressions -- Intentional! this form is the most concise way to test what we need.
-    !endsWith("cat" as "cat" | "dog", "_1" as `_${number}`);
+    endsWith(
+      "cat" as "cat" | "dog",
+      // @ts-expect-error [ts2769] -- Intentional! this is what we're testing...
```
eranhirsch:
```suggestion
      // @ts-expect-error [ts2769] -- Neither data-first overload accepts a
      // dead suffix, so the error lands here instead of on the return type.
```

packages/remeda/src/endsWith.test-d.ts
```diff
@@ -333,27 +349,43 @@ describe("data-last", () => {
...
+    );
+  });
+
+  test("template data disjoint from the suffix", () => {
+    endsWith(
+      "1_bar" as `${number}_bar`,
+      // @ts-expect-error [ts2769] -- Intentional! this is what we're testing...
+      "world",
+    );
   });
```
eranhirsch:
```suggestion

```

packages/remeda/src/endsWith.test-d.ts
```diff
@@ -388,45 +420,37 @@ describe("reject disjoint suffixes (#1432)", () => {
...
       );
     });
+
+    test("template data disjoint from the suffix", () => {
+      filter(
+        [] as `${number}_bar`[],
+        // @ts-expect-error [ts2769] -- Intentional! this is what we're testing...
+        endsWith("world"),
+      );
+    });
```
eranhirsch:
```suggestion

```

packages/remeda/src/endsWith.test-d.ts
```diff
@@ -388,45 +420,37 @@ describe("reject disjoint suffixes (#1432)", () => {
...
+    // `IsDisjoint` proves overlap by containment in either direction. A
+    // numeric placeholder makes both directions fail: `${string}1_bar` isn't
+    // contained in `${number}_bar` (it allows non-numeric heads), and
+    // `${number}_bar` isn't contained in `${string}1_bar` (it allows other
+    // digits), so the shared value `"1_bar"` goes unnoticed.
+    endsWith(
+      "1_bar" as `${number}_bar`,
+      // @ts-expect-error [ts2769] -- `"1_bar"` satisfies both templates, so this should be accepted, narrowing to `${number}_bar` & `${string}1_bar`.
+      "1_bar",
+    );
```
eranhirsch:
```suggestion
    // Rejecting on the argument requires the suffix parameter to carry the
    // dead-suffix conditional. An unresolved `T` leaves it deferred, and
    // TypeScript won't check an argument against a deferred conditional, so a
    // generic wrapper is rejected even though `T` could still match. Only the
    // call site is affected; with the error suppressed the narrowing below is
    // unchanged. The suppression sits on the call inside `endsWithBar`.
    expectTypeOf(endsWithBar($typed<string>())).toEqualTypeOf<
      `${string}bar` | undefined
    >();
  });

  describe("template literals with an impossible suffix aren't rejected", () => {
    test("data-first", () => {
      const data = "1_bar" as `${number}_bar`;

      const isEndsWith = endsWith(data, "world");

      // eslint-disable-next-line @typescript-eslint/no-invalid-void-type -- If template literals worked the same as literals and union literals it would resolve to `void` here.
      expectTypeOf(isEndsWith).not.toEqualTypeOf<void>();

      if (isEndsWith) {
        // Rejecting an impossible suffix relies on TypeScript reducing the
        // intersection with the suffix template to `never`. It only does that
        // for bounded types; an intersection of two unbounded template
        // literals is left as-is even when they are disjoint, so the check is
        // accepted and the `true` branch is typed with an uninhabitable
        // intersection instead.
        // @see https://github.com/microsoft/TypeScript/issues/60446
        expectTypeOf(data).toEqualTypeOf<`${number}_bar` & `${string}world`>();

        // No strings satisfy this type, so it should be equivalent to `never`.
        expectTypeOf(data).not.toEqualTypeOf<never>();
      } else {
        expectTypeOf(data).toEqualTypeOf<`${number}_bar`>();
      }
    });

    test("data-last", () => {
      const [yes, no] = partition([] as `${number}_bar`[], endsWith("world"));

      expectTypeOf(yes).toEqualTypeOf<(`${number}_bar` & `${string}world`)[]>();
      // Once the intersection above correctly reduces to `never`, `yes` would
      // become `never[]`; until then this stays green as a canary.
      expectTypeOf(yes).not.toEqualTypeOf<never[]>();
      expectTypeOf(no).toEqualTypeOf<`${number}_bar`[]>();
    });
```

packages/remeda/src/endsWith.test-d.ts
```diff
@@ -482,10 +506,6 @@ describe("known issues!", () => {
   });
 });
```
eranhirsch:
```suggestion

// eslint-disable-next-line @typescript-eslint/explicit-function-return-type -- Intentional, we need the types to be inferred "through" the type parameter.
const endsWithBar = <T extends string>(data: T) =>
  // @ts-expect-error [ts2769] -- Pinned in "known issues!": an unresolved `T`
  // leaves the dead-suffix conditional deferred, so the call is rejected.
  endsWith(data, "bar") ? data : undefined;

```

packages/remeda/src/startsWith.test-d.ts
```diff
@@ -333,27 +347,43 @@ describe("data-last", () => {
...
+    );
+  });
+
+  test("template data disjoint from the prefix", () => {
+    startsWith(
+      "foo_1" as `foo_${number}`,
+      // @ts-expect-error [ts2769] -- Intentional! this is what we're testing...
+      "hello",
+    );
   });
```
eranhirsch:
```suggestion

```

packages/remeda/src/startsWith.test-d.ts
```diff
@@ -1,5 +1,4 @@
 import { describe, expectTypeOf, test } from "vitest";
-import { $typed } from "../test/$typed";
 import { filter } from "./filter";
```
eranhirsch:
```suggestion
import { $typed } from "../test/$typed";
import { filter } from "./filter";
```

packages/remeda/src/startsWith.test-d.ts
```diff
@@ -388,45 +418,37 @@ describe("reject disjoint prefixes (#1432)", () => {
...
+    // `IsDisjoint` proves overlap by containment in either direction. A
+    // numeric placeholder makes both directions fail: `v1${string}` isn't
+    // contained in `v${number}` (it allows non-numeric tails), and
+    // `v${number}` isn't contained in `v1${string}` (it allows other digits),
+    // so the shared value `"v1"` goes unnoticed.
+    startsWith(
+      "v1" as `v${number}`,
+      // @ts-expect-error [ts2769] -- `"v1"` satisfies both templates, so this should be accepted, narrowing to `v${number}` & `v1${string}`.
+      "v1",
+    );
```
eranhirsch:
```suggestion
  describe("template literals with an impossible prefix aren't rejected", () => {
    test("data-first", () => {
      const data = "foo_1" as `foo_${number}`;

      const isStartsWith = startsWith(data, "hello");

      // eslint-disable-next-line @typescript-eslint/no-invalid-void-type -- If template literals worked the same as literals and union literals it would resolve to `void` here.
      expectTypeOf(isStartsWith).not.toEqualTypeOf<void>();

      if (isStartsWith) {
        // Rejecting an impossible prefix relies on TypeScript reducing the
        // intersection with the prefix template to `never`. It only does that
        // for bounded types; an intersection of two unbounded template
        // literals is left as-is even when they are disjoint, so the check is
        // accepted and the `true` branch is typed with an uninhabitable
        // intersection instead.
        // @see https://github.com/microsoft/TypeScript/issues/60446
        expectTypeOf(data).toEqualTypeOf<`foo_${number}` & `hello${string}`>();

        // No strings satisfy this type, so it should be equivalent to `never`.
        expectTypeOf(data).not.toEqualTypeOf<never>();
      } else {
        expectTypeOf(data).toEqualTypeOf<`foo_${number}`>();
      }
    });

    test("data-last", () => {
      const [yes, no] = partition([] as `foo_${number}`[], startsWith("hello"));

      expectTypeOf(yes).toEqualTypeOf<(`foo_${number}` & `hello${string}`)[]>();
      // Once the intersection above correctly reduces to `never`, `yes` would
      // become `never[]`; until then this stays green as a canary.
      expectTypeOf(yes).not.toEqualTypeOf<never[]>();
      expectTypeOf(no).toEqualTypeOf<`foo_${number}`[]>();
    });
```

packages/remeda/src/startsWith.test-d.ts
```diff
@@ -388,45 +418,37 @@ describe("reject disjoint prefixes (#1432)", () => {
...
-      // Once the intersection above correctly reduces to `never`, `yes` would
-      // become `never[]`; until then this stays green as a canary.
-      expectTypeOf(yes).not.toEqualTypeOf<never[]>();
-      expectTypeOf(no).toEqualTypeOf<`foo_${number}`[]>();
-    });
+  test("generic data is rejected even when it could match", () => {
+    // @ts-expect-error [ts6133] -- See explanation below...
+    // eslint-disable-next-line unicorn/consistent-function-scoping, @typescript-eslint/explicit-function-return-type, @typescript-eslint/no-unused-vars -- The whole purpose of the test is to see how our types handle an unresolved generic.
+    const startsWithFoo = <T extends string>(data: T) =>
+      // @ts-expect-error [ts2769] -- Intentional! this is what we're testing; the directive reports as unused once the call is accepted.
```
eranhirsch:
```suggestion
      // @ts-expect-error [ts2769] -- Pinned in "known issues!": an unresolved
      // `T` leaves the dead-prefix conditional deferred, so the call is
      // rejected.
```

packages/remeda/src/startsWith.test-d.ts
```diff
@@ -388,45 +418,37 @@ describe("reject disjoint prefixes (#1432)", () => {
...
       );
     });
+
+    test("template data disjoint from the prefix", () => {
+      filter(
+        [] as `foo_${number}`[],
+        // @ts-expect-error [ts2769] -- Intentional! this is what we're testing...
+        startsWith("hello"),
+      );
+    });
```
eranhirsch:
```suggestion

```

packages/remeda/src/startsWith.test-d.ts
```diff
@@ -51,6 +50,15 @@ describe("data-first", () => {
...
 
+  test("template literal data that overlaps the prefix", () => {
+    const data = "/api/v2" as `/api/${string}`;
+    if (startsWith(data, "/api/v2")) {
+      expectTypeOf(data).toEqualTypeOf<`/api/${string}` & `/api/v2${string}`>();
+    } else {
+      expectTypeOf(data).toEqualTypeOf<`/api/${string}`>();
+    }
+  });
+
```
eranhirsch:
```suggestion

```

packages/remeda/src/startsWith.test-d.ts
```diff
@@ -388,45 +418,37 @@ describe("reject disjoint prefixes (#1432)", () => {
...
-
-      expectTypeOf(yes).toEqualTypeOf<(`foo_${number}` & `hello${string}`)[]>();
-      // Once the intersection above correctly reduces to `never`, `yes` would
-      // become `never[]`; until then this stays green as a canary.
-      expectTypeOf(yes).not.toEqualTypeOf<never[]>();
-      expectTypeOf(no).toEqualTypeOf<`foo_${number}`[]>();
-    });
+  test("generic data is rejected even when it could match", () => {
+    // @ts-expect-error [ts6133] -- See explanation below...
+    // eslint-disable-next-line unicorn/consistent-function-scoping, @typescript-eslint/explicit-function-return-type, @typescript-eslint/no-unused-vars -- The whole purpose of the test is to see how our types handle an unresolved generic.
```
eranhirsch:
```suggestion
    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type, unicorn/consistent-function-scoping, @typescript-eslint/no-unused-vars -- Intentional, we need the types to be inferred "through" the type parameter.
```

packages/remeda/src/startsWith.test-d.ts
```diff
@@ -211,6 +213,18 @@ describe("data-last", () => {
...
+      [] as `/api/${string}`[],
+      startsWith("/api/v2"),
+    );
+
+    expectTypeOf(yes).toEqualTypeOf<
+      (`/api/${string}` & `/api/v2${string}`)[]
+    >();
+    expectTypeOf(no).toEqualTypeOf<`/api/${string}`[]>();
+  });
+
```
eranhirsch:
```suggestion

```

packages/remeda/src/endsWith.test-d.ts
```diff
@@ -392,6 +393,16 @@
...
 
 describe("known issues!", () => {
+  test("generic data is rejected even when it could match", () => {
+    // @ts-expect-error [ts6133] -- See explanation below...
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type, unicorn/consistent-function-scoping, @typescript-eslint/no-unused-vars -- Intentional, we can only test how generic type-parameters are inferred through our type parameters via a function.
+    const endsWithBar = <T extends string>(data: T) =>
```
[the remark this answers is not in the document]
eranhirsch: intentional

